PR #95, opened 2019-12-09
Revert duration hint to MAY_DUCK

ml-:
Revert the duration hint passed to `requestAudioFocus` to `AUDIOFOCUS_GAIN_TRANSIENT_MAY_DUCK` and pass `audioFocusChangeListener` to `abandonAudioFocus`.

Starting the following commit: https://github.com/dennisguse/OpenTracks/commit/d1a459efc2417d6c0bf3490459228ec4b19fff1b the duration hint was changed from `AUDIOFOCUS_GAIN_TRANSIENT_MAY_DUCK` to `AUDIOFOCUS_GAIN_TRANSIENT`.
This introduced the behavior that OpenTracks stops audio players when doing the voice announcement, after the announcement was finished the audio player however didn't resume playing and required manual intervention.

IMO the voice announcement shouldn't stop your music since it is annoying and definitely against what the user expects. Ducking the volume, as it was < `v3.2.9`, is the better option.

(In my case - using Amazon Music - I had to press pause and play again to resume the audio. The play/pause button on the BT headset I use didn't work so I had to pull out the device and do it on screen.)

I assume this behavior is due to the `null` that was passed to `abandonAudioFocus`. That's why we should pass the `audioFocusChangeListener` as described in the AudioManager reference: https://developer.android.com/reference/android/media/AudioManager.html#abandonAudioFocus(android.media.AudioManager.OnAudioFocusChangeListener)

Can't test this change before I get home. Feel free give it no priority.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/services/tasks/AnnouncementPeriodicTask.java
+++ b/src/main/java/de/dennisguse/opentracks/services/tasks/AnnouncementPeriodicTask.java
@@ -81,7 +81,7 @@ public class AnnouncementPeriodicTask implements PeriodicTask {
     private final UtteranceProgressListener utteranceListener = new UtteranceProgressListener() {
         @Override
         public void onStart(String utteranceId) {
-            int result = audioManager.requestAudioFocus(audioFocusChangeListener, TextToSpeech.Engine.DEFAULT_STREAM, AudioManager.AUDIOFOCUS_GAIN_TRANSIENT);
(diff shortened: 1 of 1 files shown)


dennisguse:
Looks good.
If you could test this with your device it would be great.

PS: Ignore TravisCI as it sadly broke some days ago - so ignore the failing CI.

ml-:
Excuse the late reply.
I've tested it and it behaves as intended. Thank you or merging.

dennisguse:
Don't worry. Thanks for fixing an issue I introduced. ;)

PR #114, opened 2019-12-29
integrate dashboard plugin

pstorch:
Integrates the dashboard plugin mechanism


dennisguse:
Almost there: the only thing that we need to change is that, at the moment, the whole tables are exposed.
However, filtering should be done in the OpenTracks `CustomContentProvider` rather than filtering on the dashboard plugin side.

dennisguse:
Also, we need to add an app chooser ;)
https://developer.android.com/training/basics/intents/sending

dennisguse:
I changed the interface that only a cursor is provided that exposes the Trackpoints for a specific TrackId.
So, filtering takes place in OpenTracks and the plugin then gets the required data (and nothing more).

NOTE: I rebased to most recent master.

@pstorch: can you squash and merge it?
No need to have separate commits, and this also attaches your user to the changes (as it was your concept).

Issue #17, opened 2019-08-09
Export: option to export with and without pictures


dennisguse:
At the moment, pictures are only included in KMZ exports (not in GPX neither in KML).
1. Make this difference visible in the export dialog
2. Also allow KMZ to export without pictures

PR #119, opened 2020-01-07
Added KMZ without photos, photo details to export menu

nthh:
Solves #17 

Before:
All as KMZ to <>
All as KML to <>
All as GPX to <>

After:
All as KMZ (with photos) to <>
All as KMZ (without photos) to <>
All as KML (without photos)  to <>
All as GPX (without photos) to <>

There's likely a more idiomatic way to do this, I also did not do any testing on the added file format (KMZ_WITH_TRACKDETAIL_AND_SENSORDATA) but can do that later if needed

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/fragments/FileTypeDialogFragment.java
+++ b/src/main/java/de/dennisguse/opentracks/fragments/FileTypeDialogFragment.java
@@ -67,12 +67,14 @@ public class FileTypeDialogFragment extends DialogFragment {
     @Override
     @NonNull
     public Dialog onCreateDialog(Bundle savedInstanceState) {
-        final TrackFileFormat[] trackFileFormats = {TrackFileFormat.KMZ_WITH_TRACKDETAIL_AND_SENSORDATA_AND_PICTURES, TrackFileFormat.KML_WITH_TRACKDETAIL_AND_SENSORDATA, TrackFileFormat.GPX};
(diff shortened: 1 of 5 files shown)


dennisguse:
@nthh Thanks!
Looks good to me - if you anyhow at this stuff do you want to add an option to export only pictures?
Since v3.3.0 pictures are stored in app internal storage and can only be exported within a KMZ.
So, an option to just export the pictures would be really awesome!

In addition, here some more translations:
``
   German:
    <string name="export_with_photos">mit Fotos</string>
    <string name="export_without_photos">ohne Fotos</string>

    Spanish (r419 as well as es):
    <string name="export_with_photos">con fotos</string>
    <string name="export_without_photos">sin fotos</string>
``

(and a really tiny nit-pick: could you make the commit message a sentence; starting upcase + ending with a dot?)

nthh:
Added additional languages/amended commit message.

What do you mean by export just the photos? What format would we use to export them? Or would we just export them as is to the location the user chooses?

dennisguse:
@nthh 
Everything looks good; Let's just wait if @edern-appere can also provide a french translation.
About the picture export, I just create a new issue for this: https://github.com/OpenTracksApp/OpenTracks/issues/123

@edern-appere could you provide a french translation?
That would be great!

dennisguse:
Feature is in v3.3.2

edern-appere:
Hi,
Sorry for the delay. I don't remember how to make a pull request or something like this. 

Here are the translations :
With photos (EN) = avec les photos (FR)
Without photos (EN) = sans les photos (FR)

PR #133, opened 2020-01-27
Forward multiple tracks to dashboard

pstorch:
Forward a list of track IDs to the dashboard. Uri's would look like this `content://de.dennisguse.opentracks.debug.content/trackpoints/62,63,64`

### DIFF
--- a/README_DEVELOPER.md
+++ b/README_DEVELOPER.md
@@ -14,15 +14,15 @@ For the time being, the OpenTracks Data API can be used to implement mainly cust
 
 ## Implementation
 OpenTracks by default does not expose any data to 3rd-party applications.
-On the request of the user (i.e., pressing a button), an Intent is created that exposes the `Track` data (it's summary) as well as the `TrackPoint`s.
+On the request of the user (i.e., pressing a button), an Intent is created that exposes the `Track` data (it's summary) as well as the `TrackPoint`s of one or more selected tracks.
 For each, one URI is provided.
 These URIs notify observers if on content changes (e.g., a new `TrackPoint` was added).
 
 Limitations:
-* only tracks selected by the user is exposed,
+* only tracks selected by the user are exposed,
 * `Waypoints` are not exposed,
 * access is only granted temporarily and automatically revoked (see `FLAG_GRANT_READ_URI_PERMISSION`),
-* no write access is impossible, and
+* no write access is possible, and
 * neither track recordings can be started, paused, or stopped.
 
 All required information can be found in `IntentDashboardUtils.java`. 
--- a/src/androidTest/java/de/dennisguse/opentracks/content/provider/CustomContentProviderUtilsTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/content/provider/CustomContentProviderUtilsTest.java
@@ -766,4 +766,12 @@ public class CustomContentProviderUtilsTest {
         contentProviderUtils.insertTrack(track);
         contentProviderUtils.bulkInsertTrackPoint(track.getLocations().toArray(new Location[0]), track.getLocations().size(), track.getId());
     }
+
+    @Test
+    public void testFormatIdListForUri() {
+        Assert.assertEquals("", ContentProviderUtils.formatIdListForUri(new long[]{}));
+        Assert.assertEquals("12", ContentProviderUtils.formatIdListForUri(new long[]{12}));
+        Assert.assertEquals("42,43,44", ContentProviderUtils.formatIdListForUri(new long[]{42, 43, 44}));
+    }
+
 }
--- a/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
@@ -22,6 +22,7 @@ import android.content.Context;
 import android.database.Cursor;
 import android.location.Location;
 import android.net.Uri;
+import android.text.TextUtils;
 
 import androidx.annotation.VisibleForTesting;
 
@@ -68,6 +69,7 @@ public class ContentProviderUtils {
      * Maximum number of waypoints that will be loaded at one time.
      */
     public static final int MAX_LOADED_WAYPOINTS_POINTS = 10000;
+    public static final String ID_SEPARATOR = ",";
 
     private final IContentResolver contentResolver;
     private int defaultCursorBatchSize = 2000;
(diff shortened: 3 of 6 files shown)


dennisguse:
I will check this on Thursday :)

dennisguse:
The code looks as if it could be prone SQLInjections as the path is just forwarded to the database.
Could this be done differently?

And could you update the README_DEVELOPER.md?

But the general approach is better than providing tons of URIs :)

pstorch:
> The code looks as if it could be prone SQLInjections as the path is just forwarded to the database.

The String concatenation is already in the existing code and with it the possibility of SQLInjections. But the URIs are created and authorized by OpenTracks and can't be manipulated, right? To be sure I can make a small test and try to manipulate the URI in the Dashboard.

pstorch:
1. changing the received Uri in the Dashboard is not possible (not even 1 character), it results in a `java.lang.SecurityException: Permission Denial` error. So I think we are safe
2. updated `README_DEVELOPER.md`
3. I had to comment out the query by `TRACKPOINTS_BY_ID`, otherwise it interfered when selecting only one TrackId. Where is a query by `TrackPointsColumns._ID` used? Did I break anything?

dennisguse:
Permission Denial is expected (but it is better to check it sometimes).
About 3. (`TRACKPOINTS_BY_ID`), I honestly don't know.
By looking at the code I am not even sure would work....
The problem is that the `uriMatcher` should not be able to react to _both_ as URI pattern is actually identical.
```
uriMatcher.addURI(ContentProviderUtils.AUTHORITY_PACKAGE, TrackPointsColumns.TABLE_NAME + "/#", UrlType.TRACKPOINTS_BY_ID.ordinal());
uriMatcher.addURI(ContentProviderUtils.AUTHORITY_PACKAGE, TrackPointsColumns.TABLE_NAME + "/#", UrlType.TRACKPOINTS_BY_TRACKID.ordinal());
```
I have some time on the weekend to check this.
My guess is that it is not used ;)

dennisguse:
Blocked by #141.

pstorch:
@dennisguse I've rebased on master and adjusted to the new Uri templates. Unfortunately the clean solution with `ContentUris.parseId(url)` doesn't work with the comma separated list of IDs.

dennisguse:
`Nitpicking mode: enabled`
@pstorch let's get this stuff on the way.
* Could you move the url creation to `ContentProviderUtils` (from StringUtils)?
* Could you actually implement the parsing in ContentProviderUtils?
  I am just not a big fan on relying that this works: `(" + url.getPathSegments().get(1) + ")");`
  And then use `String.join()`...
* Could you define the ID-separator as constant in `ContentProviderUtils` (and `,` is fine for me ",")
* Could you make SQL keywords uppercase?

`Nitpicking mode: disabled`

dennisguse:
@pstorch Just one more thing: could you make the commit message starting also uppercase?
(That's absolutely the last nit picking today, I promise)

pstorch:
After I implemented the suggestions I wanted to test again, but suddenly all my tracks are gone :sos:

dennisguse:
@pstorch Did you execute the tests on a real device?
The tests are destructive as they delete all data.....

pstorch:
@dennisguse yes, exactly that was my mistake :man_facepalming: 

I hope I understood your comment correctly. Please have a look.

README_DEVELOPER.md
```diff
@@ -14,12 +14,12 @@ For the time being, the OpenTracks Data API can be used to implement mainly cust
...
+On the request of the user (i.e., pressing a button), an Intent is created that exposes the `Track` data (it's summary) as well as the `TrackPoint`s of one or more selected tracks.
 For each, one URI is provided.
 These URIs notify observers if on content changes (e.g., a new `TrackPoint` was added).
 
 Limitations:
-* only tracks selected by the user is exposed,
+* only tracks selected by the user are exposed,
 * `Waypoints` are not exposed,
 * access is only granted temporarily and automatically revoked (see `FLAG_GRANT_READ_URI_PERMISSION`),
 * no write access is impossible, and
```
dennisguse: possible

src/androidTest/java/de/dennisguse/opentracks/content/provider/CustomContentProviderUtilsTest.java
```diff
@@ -766,4 +766,12 @@ private void insertTrackWithLocations(Track track) {
         contentProviderUtils.insertTrack(track);
         contentProviderUtils.bulkInsertTrackPoint(track.getLocations().toArray(new Location[0]), track.getLocations().size(), track.getId());
     }
+
+    @Test
+    public void testFormatIdListForUri() {
+        Assert.assertEquals("", ContentProviderUtils.formatIdListForUri(new long[]{}));
+        Assert.assertEquals("12", ContentProviderUtils.formatIdListForUri(new long[]{12}));
+        Assert.assertEquals("42,43,44", ContentProviderUtils.formatIdListForUri(new long[]{42,43,44}));
```
dennisguse: Can  you applying autoformatting?
(I guess, at the moment the defaults of AndroidStudio are used).

src/androidTest/java/de/dennisguse/opentracks/util/StringUtilsTest.java
```diff
@@ -208,4 +208,5 @@ public void testFormatDecimal() {
 
         Assert.assertEquals("1", StringUtils.formatDecimal(0.99, 1));
     }
+
```
dennisguse: Remove

src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
```diff
@@ -942,4 +944,30 @@ int getDefaultCursorBatchSize() {
...
+            if (idsPathSegment.length() > 0) {
+                idsPathSegment.append(ID_SEPARATOR);
+            }
+            idsPathSegment.append(id);
+        }
+        return idsPathSegment.toString();
+    }
+
+    public static String[] parseTrackIdsFromTrackPointsUri(Uri url) {
+        return TextUtils.split(url.getPathSegments().get(2), ID_SEPARATOR);
```
dennisguse: url.getLastPathSegment()

src/main/java/de/dennisguse/opentracks/util/StringUtils.java
```diff
@@ -415,4 +415,5 @@ public static long parseTime(String xmlDateTime) {
         }
         return new Pair<>(value, unit);
     }
+
```
dennisguse: Remove

pstorch:
@dennisguse  one more try :wink:

dennisguse:
I just don't know why testing on TravisCI fails in some cases.
Hopefully, Github Actions is slightly less picky (then TravisCI can be just used for compiling).

Issue #122, opened 2020-01-08
Marker: allow to attach picture


dennisguse:
Markers can be created taking a picture using the camera.
It would be great to also allow the user to select an image from the internal gallery.
A selected picture should then be copied to the OpenTracks storage.

Details on how to get the picture URL can be found here: https://androidclarified.com/pick-image-gallery-camera-android/

Actually: this is also necessary to create new screenshots for the store listings.
Due to winter creating outdoors useful/nice pictures is rather hard.

dennisguse:
A more modern example can be found here: https://medium.com/@pednekarshashank33/android-10s-scoped-storage-image-picker-gallery-camera-d3dcca427bbf

I guess it would be best to copy the image to the app-internal storage where also the ones taken with camera are stored.

dennisguse:
Fixed by PR #157.

PR #157, opened 2020-02-27
Myissue122

rgmf:

**Describe the pull request**
~~Now, when user click on "Insert Photo" option the app ask to the user what to do: take a photo or select an image from the Gallery~~ Now user can use a new option (Insert gallery image) to create a waypoint with a gallery image. Also, It's not necessary Storage permission because I have used SAF (Storage Access Framework).

I have done the same you can see here: https://medium.com/@pednekarshashank33/android-10s-scoped-storage-image-picker-gallery-camera-d3dcca427bbf

I have tested it on a virtual device and on my Pixel XL 2.

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/122

**Link to the APK**
I have uploaded the APK to my nextcloud:
https://ppp.woelkli.com/s/JMxJ6tHXB9wXN6c
password for download: OpenTracks122

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/MarkerEditActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/MarkerEditActivity.java
@@ -16,18 +16,38 @@
 
 package de.dennisguse.opentracks;
 
+import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.net.Uri;
 import android.os.Bundle;
+import android.os.ParcelFileDescriptor;
+import android.provider.MediaStore;
 import android.util.Log;
+import android.util.Pair;
+import android.view.Menu;
+import android.view.MenuItem;
 import android.view.View;
 import android.view.View.OnClickListener;
 import android.widget.ArrayAdapter;
 import android.widget.AutoCompleteTextView;
 import android.widget.Button;
 import android.widget.EditText;
+import android.widget.ImageView;
+import android.widget.Toast;
+
+import androidx.annotation.NonNull;
+
+import java.io.File;
+import java.io.FileDescriptor;
+import java.io.IOException;
 
 import de.dennisguse.opentracks.content.data.Waypoint;
 import de.dennisguse.opentracks.content.provider.ContentProviderUtils;
 import de.dennisguse.opentracks.services.TrackRecordingServiceConnection;
+import de.dennisguse.opentracks.util.FileUtils;
+import de.dennisguse.opentracks.util.IntentUtils;
 
 /**
  * An activity to add/edit a marker.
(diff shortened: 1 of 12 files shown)


dennisguse:
Thanks. 
From a very quick glance:
* could you squash both commits?
  For this, you just need to squash them locally and then do a force push. Github will update the PR.
* can you use more descriptive commit messages? So what does this change actually do + the issue number...
This makes it easier for future devs to figure out why something was done (also we might migrate to a different got location and then issues are probably not available anymore)
* You can use try for autoclosables and thus  omitting the manual finally block
* Why is FILEPROVIDER now public instead of package limited (I don't have an IDE at hand)

And I cannot install the debug-apk. I actually don't know if it is possible to share this.

And I guess, we should not use a popup to distinguish between camera and selecting an image. The 2nd is probably less likely when a user is actually doing some outdoor stuff and rather happening afterwards.
So, an additional menu item might be better.
What do you think?

Again thanks for your work!

rgmf:
> could you squash both commits?
>       For this, you just need to squash them locally and then do a force push. Github will update the PR.

I think is done (I've never done).

> can you use more descriptive commit messages? So what does this change actually do + the issue number...

Sorry, you're right. Done.


>  You can use try for autoclosables and thus  omitting the manual finally block

Yes, you're right. Done.

> Why is FILEPROVIDER now public instead of package limited (I don't have an IDE at hand)

Before I used FILEPROVIDER in TrackDetailActivity. I have changed it, so now FILEPROVIDER i package limited.

> And I cannot install the debug-apk. I actually don't know if it is possible to share this.

I built the apk from Android Studio and uploaded it to a nextcloud. I tested with other phone and I could install it. How can I pass you the apk?

> And I guess, we should not use a popup to distinguish between camera and selecting an image. The 2nd is probably less likely when a user is actually doing some outdoor stuff and rather happening afterwards.
> So, an additional menu item might be better.
> What do you think?

I think this is the best option but I thought you would prefer phone ask you what to do. I have changed the code, so now there is a new option in the menu.

Thank you ;)

dennisguse:
@rgmf Thanks for the changes!
For some reason, my Android devices do not allow me to install debug builds (I also tried OSMDashboard apks from the release.page). Don't know why and I cannot do a logcat to determine the reason).
Anyhow: your changes look good!

Without trying anything, I yesterday had the idea that this functionality might be actually more related to the MarkerDetailActivity to either set/remove/change a picture of an already created marker. Thus, while tracking the user creates the markers as usual and afterwards can change the pictures if desired.
As a user I would not like to bother setting the image while I tracking but rather afterwards.
I guess both functionalities could actually live in parallel.
What do you think?

rgmf:
> Without trying anything, I yesterday had the idea that this functionality might be actually more related to the MarkerDetailActivity to either set/remove/change a picture of an already created marker. Thus, while tracking the user creates the markers as usual and afterwards can change the pictures if desired.
> As a user I would not like to bother setting the image while I tracking but rather afterwards.
> I guess both functionalities could actually live in parallel.
> What do you think?

Yes, it's needed!!!

For example, I have been using the app this weekend in my outdoor activities. I inserted images (with camera or from gallery) on the way, but later I couldn't change the images or add new ones on the regular waypoints. So this should be fixed.

If you don't mind I'm going to check it out.

I think there will be two scenarios:
- The marker doesn't has image -> when user click on edit button he should be able to add an image (taking a photo or selecting one from the gallery).
- The marker has image -> when user click on edit button he should be able to change the image (taking a photo or selecting one from the gallery) or delete the image.

Now, the only thing you can change from a Marker is: name, marker type and description.

Am I right?

dennisguse:
Interesting idea.
I guess moving everything to MarkerDetail is a good idea. Also it makes the menu less crowded.
And also makes it less weird to have two ways to create markers.

@rgmf if you have time - feel free to work on this :)
About the scenarios: jupp.
Separate icons per menu item (add vs remove picture would be great).

rgmf:
It's done now.

When you add or edit a marker you can add a photo from camera or gallery. All is in MarkerEditActivity like we spoke.

So when you are tracking an activity and you click on add marker button it opens Marker Edit Activity to set a name, type, description and take a photo or select a gallery image.

Later when you open a created marker you can edit all attributes of this marker including the photo (delete, add, change).

dennisguse:
Thanks :)

Issue #151, opened 2020-02-14
ability to set preferred rate by activity type


sporksmith:
It's currently a global setting but e.g. I prefer cycling in miles/hr but running in minutes/mile.

dennisguse:
Is it about the UI or the voice announcements?
#87 is related.

sporksmith:
I'm referring to the UI.

I haven't used the voice announcements, but I imagine it'd be applicable there too.

rgmf:
What do you think something like that?

In **Settings** you have **Preferred rate** with two options: **Speed or Pace**. Well, you can put a third option: **Automatic** (or something like that). If the:

- User choose automatic the app will use pace for RUN and WALK and speed for the others
- User choose speed, the app will use speed for all tracks.
- User choose pace, the app will use pace for all tracks.

Something like this:
```
diff --git a/src/main/java/de/dennisguse/opentracks/fragments/ChartFragment.java b/src/main/java/de/dennisguse/opentracks/fragments/ChartFragment.java
index bed4b8d8..4035df29 100644
--- a/src/main/java/de/dennisguse/opentracks/fragments/ChartFragment.java
+++ b/src/main/java/de/dennisguse/opentracks/fragments/ChartFragment.java
@@ -79,6 +79,7 @@ public class ChartFragment extends Fragment implements TrackDataListener {
     private boolean metricUnits = true;
     private boolean reportSpeed = true;
     private int recordingDistanceInterval;
+    private String category = "";
 
     // Modes of operation
     private boolean chartByDistance;
@@ -109,7 +110,7 @@ public class ChartFragment extends Fragment implements TrackDataListener {
                 }
             }
             if (PreferencesUtils.isKey(getContext(), R.string.stats_rate_key, key)) {
-                boolean speed = PreferencesUtils.isReportSpeed(getContext());
+                boolean speed = PreferencesUtils.isReportSpeed(getContext(), category);
                 if (reportSpeed != speed) {
                     reportSpeed = speed;
                     chartView.setReportSpeed(reportSpeed);
@@ -202,8 +203,10 @@ public class ChartFragment extends Fragment implements TrackDataListener {
         if (isResumed()) {
             if (track == null || track.getTripStatistics() == null) {
                 startTime = -1L;
+                category = "";
                 return;
             }
+            category = track.getCategory();
             startTime = track.getTripStatistics().getStartTime();
         }
     }
[...]

PR #161, opened 2020-03-14
Added new option in Preferred rate setting: By Default.

rgmf:
**Describe the pull request**
I've added a "By Default" option in Preferred rate Settings. I think is a better text than "Auto", but it can be easily changed.

In `TrackIconUtils` I have created a new int array with all categories that use speed by default (there are sports that use speed and the other ones use pace). Every time a developer creates a new Icon has to add to this array the new sport if it uses speed by default.

From outside it can be used `isSpeedIcon` public static method to see if a category (activity type) uses speed (true) or pace (false).

Also `isReportSpeed` static method in `PreferencesUtils.java` now need a second argument: the track's category (activity type).

Those are basically all changes done.

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/151

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/fragments/ChartFragment.java
+++ b/src/main/java/de/dennisguse/opentracks/fragments/ChartFragment.java
@@ -199,6 +199,11 @@ public class ChartFragment extends Fragment implements TrackDataListener {
             }
 
             category = track.getCategory();
+            boolean reportSpeed = PreferencesUtils.isReportSpeed(getContext(), category);
+            if (reportSpeed != chartView.getReportSpeed()) {
+                chartView.setReportSpeed(reportSpeed);
+                chartView.applyReportSpeed();
+            }
             startTime = track.getTrackStatistics().getStartTime_ms();
         }
     }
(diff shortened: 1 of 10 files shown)


dennisguse:
@rgmf Thanks, I will take a look next week - I want to finish some other stuff first :)

dennisguse:
@pstorch @rgmf What do we do about this PR?
I am still not a big fan, but I could imagine it might be useful.
I must admit I have not tried - but what happens if I am recording something and then change the activity type.
Does this result in an immediate change of `StatsFragment`?

PS: @rgmf Still nice work :)

rgmf:
> I must admit I have not tried - but what happens if I am recording something and then change the activity type.
> Does this result in an immediate change of StatsFragment?

If you change from cycling to running, for example, then rate is change from km/h to min/km.

> PS: @rgmf Still nice work :)

Thank you ;)

> I am still not a big fan, but I could imagine it might be useful.

If you have not convinced don't merge it. Don't worry, it's a simple change, think about it and let's see ;)

pstorch:
I don't have an opinion on it. I only do one sport with always the same setting. But if it is useful for others, then yes.

dennisguse:
@rgmf If you like it just merge it ;)

rgmf:
Ok, for me yes :D

PR #162, opened 2020-03-16
Share pictures from marker.

### LINKED ISSUE #89: Share pictures from marker
If a picture marker was created, the picture itself cannot be shared alone from the `MarkerDetailFragment`.

Add a share-button as for tracks.

rgmf:
Fixes #89.

**Describe the pull request**
I added a new menu option in `MarkerDetailFragment` that only is visible in markers with photo/image. This option can be used to share the photo/image of a marker.

Also, I added new strings in english, spanish and catalan languages for images shared (a subject and a text).

I would like to point out a issue that I think you have to think about: 

I would like to point out an issue that I think you have to think about: I've added a new method in IntentUtils that can be used to share an image but nobody avoid that somebody use this method for share any other file because the only thing it needs is a URI (and an URI can be any kind of file). Do you think, this method should be test that URI is a file before to share, something like that:

```
public static Intent newShareImageIntent(Context context, Uri uri) {
        if (uri is not an image) {
                throw new RuntimeException("The file to share has to be an image.");
        }

        ...
}
```

I thought on that but I'd like to know what do you think.

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/89

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/fragments/MarkerDetailFragment.java
+++ b/src/main/java/de/dennisguse/opentracks/fragments/MarkerDetailFragment.java
@@ -60,6 +60,8 @@ public class MarkerDetailFragment extends Fragment {
 
     private static final long HIDE_TEXT_DELAY = 4 * UnitConversions.ONE_SECOND_MS;
 
+    private MenuItem shareMarkerImageMenuItem;
+
     private ContentProviderUtils contentProviderUtils;
     private Handler handler;
     private ImageView photoView;
@@ -145,6 +147,7 @@ public class MarkerDetailFragment extends Fragment {
         // Need to update the waypoint in case returning after an edit
         updateWaypoint(true);
         updateUi();
+        updateMenuItems();
     }
 
     @Override
@@ -170,8 +173,14 @@ public class MarkerDetailFragment extends Fragment {
     @Override
     public void onCreateOptionsMenu(@NonNull Menu menu, MenuInflater inflater) {
         inflater.inflate(R.menu.marker_detail, menu);
-
+        shareMarkerImageMenuItem = menu.findItem(R.id.marker_detail_share);
         updateWaypoint(false);
+        updateMenuItems();
+    }
+
+    private void updateMenuItems() {
+        if (shareMarkerImageMenuItem != null)
+            shareMarkerImageMenuItem.setVisible(waypoint.hasPhoto());
     }
 
     @Override
@@ -187,6 +196,13 @@ public class MarkerDetailFragment extends Fragment {
                 intent = IntentUtils.newIntent(fragmentActivity, MarkerEditActivity.class)
                         .putExtra(MarkerEditActivity.EXTRA_MARKER_ID, markerId);
                 startActivity(intent);
+                return true;
+            case R.id.marker_detail_share:
+                if (waypoint.hasPhoto()) {
+                    intent = IntentUtils.newShareImageIntent(getContext(), waypoint.getPhotoURI());
+                    intent = Intent.createChooser(intent, null);
+                    startActivity(intent);
+                }
                 return true;
             case R.id.marker_detail_delete:
                 DeleteMarkerDialogFragment.showDialog(getChildFragmentManager(), new long[]{markerId});
--- a/src/main/java/de/dennisguse/opentracks/util/IntentUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/util/IntentUtils.java
@@ -107,6 +107,30 @@ public class IntentUtils {
                 .putExtra(Intent.EXTRA_TEXT, context.getString(R.string.share_track_share_file_body, trackDescription));
     }
 
+    /**
+     * Creates an intent to share a waypoint image with an app.
+     *
+     * @param context the context.
+     * @param uri     uri with the image to share.
+     */
+    public static Intent newShareImageIntent(Context context, Uri uri) {
+        String mime = context.getContentResolver().getType(uri);
+
+        /*
+         * Because the #166 bug, when you import KMZ tracks then it creates file:/// from markers with photo.
+         * The photos should be content:/// not file:/// because getType(uri) always returns null for file:///
+         * In .setType, to avoid side effects because the #166 bug described above it checks if mime is null.
+         * If it is then it hardcode "images/*".
+         */
+        return new Intent(Intent.ACTION_SEND)
+                .setType(mime != null ? mime : "image/*")
+                .setAction(Intent.ACTION_SEND)
+                .addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+                .putExtra(Intent.EXTRA_STREAM, uri)
+                .putExtra(Intent.EXTRA_SUBJECT, context.getString(R.string.share_image_subject))
+                .putExtra(Intent.EXTRA_TEXT, context.getString(R.string.share_image_body));
+    }
+
     public static void showCoordinateOnMap(Context context, Waypoint waypoint) {
         showCoordinateOnMap(context, waypoint.getLocation().getLatitude(), waypoint.getLocation().getLongitude(), waypoint.getName());
     }
--- a/src/main/res/menu/marker_detail.xml
+++ b/src/main/res/menu/marker_detail.xml
@@ -21,6 +21,11 @@ limitations under the License.
         android:title="@string/menu_show_on_map"
         app:showAsAction="ifRoom" />
     <item
+        android:id="@+id/marker_detail_share"
+        android:icon="@drawable/ic_share_24dp"
+        android:title="@string/menu_share"
+        app:showAsAction="ifRoom" />
+    <item
         android:id="@+id/marker_detail_edit"
         android:icon="@drawable/ic_edit_24dp"
         android:title="@string/menu_edit"
--- a/src/main/res/values-b+es+419/strings.xml
+++ b/src/main/res/values-b+es+419/strings.xml
@@ -446,4 +446,7 @@
     <string name="waypoint_type_waterfall">cascada</string>
     <string name="waypoint_type_water_fountain">fuente de agua</string>
     <string name="waypoint_type_waypoint">punto de referencia</string>
+    <!-- Share image -->
+    <string name="share_image_subject">Me gustaría compartir una imagen contigo</string>
+    <string name="share_image_body">Creo que esta imagen podría interesarte.</string>
 </resources>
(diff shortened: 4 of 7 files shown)


dennisguse:
At least on my lineageos the sharing fails:
```
2020-03-16 17:58:55.944 12823-12823/de.dennisguse.opentracks.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.dennisguse.opentracks.debug, PID: 12823
    android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.SEND flg=0x1 clip={null T:I think you might be interested in this image.} (has extras) }
```

rgmf:
> At least on my lineageos the sharing fails:
> 
> ```
> 2020-03-16 17:58:55.944 12823-12823/de.dennisguse.opentracks.debug E/AndroidRuntime: FATAL EXCEPTION: main
>     Process: de.dennisguse.opentracks.debug, PID: 12823
>     android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.SEND flg=0x1 clip={null T:I think you might be interested in this image.} (has extras) }
> ```

Do you have any app installed for send files like mail app, whatsapp or telegram?

Anyway, I should check in the code that user has the apps needed to attend the Intent. I've tested it in my two physical phones.

dennisguse:
@rgmf  I have at least K9-Mail.
The issue seems to be that `context.getContentResolver().getType(uri)` returns `null`.
One example uri: `file:///storage/emulated/0/Android/data/de.dennisguse.opentracks.debug/files/Pictures/10/Jan%201%2C%202020%202_25_40%20PM.jpeg`

dennisguse:
And one unrelated (minor) thing: could you move the share-icon to position two?
Then it is at the same for tracks and markers.

rgmf:
> @rgmf I have at least K9-Mail.
> The issue seems to be that `context.getContentResolver().getType(uri)` returns `null`.
> One example uri: `file:///storage/emulated/0/Android/data/de.dennisguse.opentracks.debug/files/Pictures/10/Jan%201%2C%202020%202_25_40%20PM.jpeg`

Ok, its because uri is `file:///`. In all my tests the uri was `content:///` and seem that `getType(uri)` allways returns null for `file:///` uri. I have to find the way to get the mime type ragardless of the uri.

What I don't understand is why in ones phones the uri is a `file:///` and in the others is a `content:///`...

> And one unrelated (minor) thing: could you move the share-icon to position two?
> Then it is at the same for tracks and markers.

Of course ;)

dennisguse:
@rgmf Found it:
```
                    intent = IntentUtils.newShareImageIntent(getContext(), waypoint.getPhotoURI());
                    intent = Intent.createChooser(intent, null);
                    startActivity(intent);
```

`Intent.createChooser()` creates a new intent that needs to be used before passing it to `startActivity()`.
Then also no activity found is handled properly.

And the actual issue with no activity found is really `getType()` as it resolves on my phone to `null`.
This works for me:
```
        return new Intent(Intent.ACTION_SEND)
//               .setType(context.getContentResolver().getType(uri))
                .setType("image/*")
                .setAction(Intent.ACTION_SEND)
                .putExtra(Intent.EXTRA_STREAM, uri)
                .putExtra(Intent.EXTRA_SUBJECT, context.getString(R.string.share_image_subject))
                .putExtra(Intent.EXTRA_TEXT, context.getString(R.string.share_image_body));
```
But I honestly don't know why the url is not identified correctly.

dennisguse:
@rgmf Good catch: it should be a content:// url if the image was captured using the camera.
I just imported the track from a KMZ and then it seems to create a file:// url.
So, that's something we need to fix, but can be done later.
I will confirm this behavior and create a new issue for this.

We can just use `image/*`as fallback if `getType(url) `returned `null`.

rgmf:
> @rgmf Good catch: it should be a content:// url if the image was captured using the camera.
> I just imported the track from a KMZ and then it seems to create a file:// url.
> So, that's something we need to fix, but can be done later.
> I will confirm this behavior and create a new issue for this.
> 
> We can just use `image/*`as fallback if `getType(url) `returned `null`.

Ok.

Do I do this?
`.setType(mime != null ? mime : "image/*")`

dennisguse:
@rgmf And also add a comment, why this was necessary.
Otherwise nobody will figure this one out in some years :)

rgmf:
I've tested that works with file:/// and with content:///

src/main/java/de/dennisguse/opentracks/util/IntentUtils.java
```diff
@@ -99,6 +101,29 @@ public static Intent newShareFileIntent(Context context, long[] trackIds) {
...
+    /**
+     * Creates an intent to share a waypoint image with an app.
+     *
+     * @param context the context.
+     * @param uri     uri with the image to share.
+     */
+    public static Intent newShareImageIntent(Context context, Uri uri) {
+        String mime = context.getContentResolver().getType(uri);
+
+        // Because the #166 bug, when you import KMZ tracks then it creates file:/// from markers
```
dennisguse: Could you use a line break after every sentence?
It is usually easier to read

src/main/java/de/dennisguse/opentracks/fragments/MarkerDetailFragment.java
```diff
@@ -188,6 +197,13 @@ public boolean onOptionsItemSelected(MenuItem item) {
                         .putExtra(MarkerEditActivity.EXTRA_MARKER_ID, markerId);
                 startActivity(intent);
                 return true;
+            case R.id.marker_detail_share:
+                if (waypoint.hasPhoto()) {
+                    intent = IntentUtils.newShareImageIntent(getContext(), waypoint.getPhotoURI());
+                    Intent.createChooser(intent, null);
```
dennisguse: needs to be:
intent = Intent.createChooser(intent, null);

src/main/java/de/dennisguse/opentracks/fragments/MarkerDetailFragment.java
```diff
@@ -170,8 +173,14 @@ public void setMenuVisibility(boolean menuVisible) {
...
     public void onCreateOptionsMenu(@NonNull Menu menu, MenuInflater inflater) {
         inflater.inflate(R.menu.marker_detail, menu);
-
+        shareMarkerImageMenuItem = menu.findItem(R.id.marker_detail_share);
         updateWaypoint(false);
+        updateMenuItems();
+    }
+
+    private void updateMenuItems() {
+        if (shareMarkerImageMenuItem != null)
```
dennisguse: Is this null check required?
And this code could actually be moved into the calling methods (as long as it is only one menuItem).

rgmf: Yes, it's required because I need to check if this option must be hidden every time this MarkerDetailFragment is resumed.

The first time the user enter in the marker detail, MarkerDetailFragment onResume is execute and the menu isn't created yet so shareMarkerImageMenuItem is null.

But now you imagine that you go to the detail of a marker with image. Then you edit the marker and deletes the image. Later you go back to the MarkerDetailFragment, onResume is executed, shareMarkerImageMenuItem isn't null and it must be not visible because the image was deleted.

src/main/java/de/dennisguse/opentracks/util/IntentUtils.java
rgmf: Of course ;)

src/main/java/de/dennisguse/opentracks/fragments/MarkerDetailFragment.java
rgmf: Ops, you are right. Sorry ;)

dennisguse: Happens to me all the time :)

dennisguse:
@rgmf Can you make the changes, so I can merge?

rgmf:
> @rgmf Can you make the changes, so I can merge?

I discovered a problem when you share a bookmark. It works, but you can see in the Logcat this:
`java.lang.SecurityException: Permission Denial: reading androidx.core.content.FileProvider uri content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/47/29%20Feb%202020%2008_36_07.jpeg from pid=26382, uid=1000 requires the provider be exported, or grantUriPermission()`

Is the same thing occur when you share a track. If you try to share a track works but in the logcat I see the same error.

I've fixed it here. I push the changes so you can look at them.

Seem that `createChooser` need permission for use the image to share like you can see in the images below:

**- Without grant permission, createChooser cannot use the image to preview it (with logcat error):**
with_error


**-With grant permission, createChooser can use the image to preview it (without logcat error):**
without_error

dennisguse:
Could we not use `.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)`?
This just gives temporary access to the URI and then we don't need to grant as well as revoke access.

rgmf:
> Could we not use `.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)`?
> This just gives temporary access to the URI and then we don't need to grant as well as revoke access.

It was the first thing I tested but on my phone (Android 10) didn't work :(

When I was in trouble I read about the issue of grant permission. For example: https://medium.com/@benexus/dealing-with-permissions-when-sharing-files-android-m-cee9ecc287bf

`.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)` works for you?

dennisguse:
@rgmf Jupp that works for me.
Actually, the same approach is used for sharing tracks.
i am running on a LineageOS Pixel1

dennisguse:
However, as long as #166 is not fixed, we have the issue that some URLs might actually be file:// and these cannot be shared without manually using a `FileProvider.getUriForFile()`. 
@rgmf do you also have the issue for content://-URIs?

rgmf:
> @rgmf Jupp that works for me.
> Actually, the same approach is used for sharing tracks.
> i am running on a LineageOS Pixel1

So, I change it. The code is more simple with the .addFlag
Anyway, the share works in my phone although it fires the error in the logcat.

;)

rgmf:
> However, as long as #166 is not fixed, we have the issue that some URLs might actually be file:// and these cannot be shared without manually using a `FileProvider.getUriForFile()`.
> @rgmf do you also have the issue for content://-URIs?

@dennisguse I don't have issues sharing file:// neither content://

The only thing is that I see the error in the logcat, but sharing works well.

dennisguse:
@rgmf NOW i got it - took me a little longer.....
Actually, I don't have this issue on my phones as those do not yet support the preview feature.
Anyhow, I am not a big fan of introducing this complication.
I mean in this case it is just to show a preview of an image that was just shown in anyhow.
The error message is not great, but the code gets the actual job done.

src/main/java/de/dennisguse/opentracks/util/IntentUtils.java
dennisguse: This is the last one (I promise): but could you modify the comment layout to use:
1. /* */
Because all 
2. Write one sentence per line followed by a line break
It is usually easier to read and will apply the proper formatting depending on the IDE's window width.

rgmf:
> Last one ;)

Sorry, it's logical. It was a misunderstanding.

If you see anything else let me know ;)

Issue #176, opened 2020-04-06
"Export all" with tracks without points results in a null exception


rgmf:
**Describe the bug**

When you use the "Export all" option in Settings the app cracks and exits if there are tracks without points.

This is because in `KmlTrackWriter` `writeBeginTrack` can be called with a null in the `startTrackPoint` argument. Later, in this method `startTrackPoint` is used without care if it's null or not.

Seem easy to fix but like here is writting out a kml file it's needed testing it deeply and with care so the output file fulfil kml standard. The key question here is: Can exists a <Placemark> without <Point>? If so, this bug is easy to fix.

**To Reproduce**
1.- Record a track with GPS disabled. Only total time is saved.
2.- Stop and save the track.
3.- Go to Settins -> Export all.

Was discovered in #166 because I needed tracks to import so I exported all my tracks.

dennisguse:
Nice that you found this one.

rgmf:
Fix in https://github.com/OpenTracksApp/OpenTracks/pull/178

PR #178, opened 2020-04-07
Importing creates content:// for attached pictures and exporting with…

### LINKED ISSUE #166: Importing a track creates file:// for attached pictures
**Describe the bug**
Importing a track with images, creates file:/// urls instead as content:/// as photoUrl.

**To Reproduce**
1. Create a track with one marker and photo
2. Export track
3. Import exported track
4. A file:// is created for the photoUrl

Was discovered as a side effect of #162.

Does not affect anything in the system.

rgmf:
…out points bug fixed. Fixes #166 and #176.

**Describe the pull request**
"Import all" option now creates content:// for attached pictures instead of file://

Also, "Export all" option doesn't results in a null exception when a file without points is exported.

**Link to the the issue**
#166 
#176 

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackWriter.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackWriter.java
@@ -74,7 +74,7 @@ public class KmlTrackWriter implements TrackWriter {
     /**
      * @param context            the context
      * @param hasMultipleTracks  should encode multiple tracks into one file?
-     * @param exportTrackDetail should detailed information about the track be exported (e.g., title, description, waypoints, timing)?
+     * @param exportTrackDetail  should detailed information about the track be exported (e.g., title, description, waypoints, timing)?
      * @param exportSensorData   should {@link SensorDataSet} be exported?
      * @param exportPhotos       should pictures be exported (if true: exports to KMZ)?
      */
@@ -192,7 +192,8 @@ public class KmlTrackWriter implements TrackWriter {
         this.startTrackPoint = startTrackPoint;
         if (printWriter != null) {
             String name = context.getString(R.string.marker_label_start, track.getName());
-            writePlacemark(name, "", "", START_STYLE, startTrackPoint.getLocation());
+            Location location = startTrackPoint != null ? startTrackPoint.getLocation() : null;
+            writePlacemark(name, "", "", START_STYLE, location);
             printWriter.println("<Placemark>");
 
             if (exportTrackDetail) {
@@ -218,7 +219,8 @@ public class KmlTrackWriter implements TrackWriter {
             if (exportTrackDetail) {
                 String name = context.getString(R.string.marker_label_end, track.getName());
                 String description = descriptionGenerator.generateTrackDescription(track, false);
-                writePlacemark(name, "", description, END_STYLE, endTrackPoint.getLocation());
+                Location location = endTrackPoint != null ? endTrackPoint.getLocation() : null;
+                writePlacemark(name, "", description, END_STYLE, location);
             }
         }
     }
@@ -345,7 +347,7 @@ public class KmlTrackWriter implements TrackWriter {
             writeCategory(category);
 
             if (exportPhotos) {
-                printWriter.println("<Icon><href>" + Uri.decode(photoUrl) + "</href></Icon>");
+                printWriter.println("<Icon><href>" + KmzTrackExporter.buildKmzImageFilePath(Uri.parse(photoUrl)) + "</href></Icon>");
             }
 
             printWriter.print("<ViewVolume>");
--- a/src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
@@ -366,15 +366,32 @@ abstract class AbstractFileTrackImporter extends DefaultHandler implements Track
     /**
      * Gets the photo url for a file.
      *
-     * @param fileName the file name
+     * @param externalPhotoUrl the file name
      */
-    protected String getPhotoUrl(String fileName) {
+    protected String getInternalPhotoUrl(String externalPhotoUrl) {
         if (importTrackId == -1L) {
+            Log.e(TAG, "Track id is invalid.");
+            return null;
+        }
+
+        if (externalPhotoUrl == null) {
+            Log.i(TAG, "External photo url is null.");
+            return null;
+        }
+
+        Uri externalPhotoUri = Uri.parse(externalPhotoUrl);
+        if (externalPhotoUri == null) {
+            Log.w(TAG, "Could not parse external photo url.");
+            return null;
+        }
+        String filename = externalPhotoUri.getLastPathSegment();
+        if (filename == null) {
+            Log.w(TAG, "External photo contains no filename.");
             return null;
         }
 
         File dir = FileUtils.getPhotoDir(context, importTrackId);
-        File file = new File(dir, fileName);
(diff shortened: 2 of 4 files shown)


dennisguse:
I am right that the exported URL is still a content://, which is in the scope of the KMZ file invalid?

rgmf:
> I am right that the exported URL is still a content://, which is in the scope of the KMZ file invalid?

What do you expect in the exported file?

For example, imagine this KMZ file:
```
9.kmz
  |__ images
        |__ Apr 7, 2020 3_22_27_PM.jpeg
  |__ doc.kml
```

Now, in the doc.kml file we have this:
```kml
<href>content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/9/Apr 7, 2020 3_22_27 PM.jpeg</href>
```

What value should it is? This?
```kml
<href>./images/Apr 7, 2020 3_22_27 PM.jpeg</href>
```

I've not touched nothing about this. The only exported part I've touched have been the null exception. But I can fix this too ;)

These changes do not affect the import at all.

dennisguse:
@rgmf I guess yes: https://developers.google.com/kml/documentation/kmzarchives#resolving-relative-references
It looks like the href can be relative link, which then should be evaluated within the current KMZ.

rgmf:
> @rgmf I guess yes: https://developers.google.com/kml/documentation/kmzarchives#resolving-relative-references
> It looks like the href can be relative link, which then should be evaluated within the current KMZ.

Ok @dennisguse, that's makes sense :D

So, I'm going to do so. I think tomorrow in the morning I can get it ;)

rgmf:
@dennisguse look at carefully my solution. I don't know if there is a better option.

The key here is that `KmzTrackExporter` builds the path image file where image will be saved. So, I've used this code and put it in a public static method so whatever class can be use it.

src/main/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackWriter.java
```diff
@@ -345,7 +347,7 @@ private void writePhotoOverlay(String name, String category, String description,
             writeCategory(category);
 
             if (exportPhotos) {
-                printWriter.println("<Icon><href>" + Uri.decode(photoUrl) + "</href></Icon>");
```
dennisguse, line 348: I know it was not handled before, but what happens if the URL is invalid?

src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
```diff
@@ -372,9 +372,13 @@ protected String getPhotoUrl(String fileName) {
...
             return null;
         }
+
         File dir = FileUtils.getPhotoDir(context, importTrackId);
         File file = new File(dir, fileName);
-        return Uri.fromFile(file).toString();
+
+        Uri photoUri = FileUtils.getUriForFile(context, file);
+
+        return photoUri.toString();
```
dennisguse, line 381: Better use: ""  + photoUri to avoid NullPointer.

src/main/java/de/dennisguse/opentracks/io/file/importer/KmlFileTrackImporter.java
```diff
@@ -186,9 +186,9 @@ private void onWaypointEnd() throws SAXException {
             return;
         }
 
+        // If there is photoUrl it has to be changed because that url in kml file isn't a valid content:// to the imported one.
         if (photoUrl != null) {
-            Uri uri = Uri.parse(photoUrl);
-            photoUrl = getPhotoUrl(uri.getLastPathSegment());
+            photoUrl = getPhotoUrl(Uri.parse(photoUrl).getLastPathSegment());
```
dennisguse, line 191: Uri parsing might fail and there might be no path.

And as we are importing here something from somewhere, we should check this carefully.

dennisguse:
@rgmf I have not tested anything and just quickly "read" your code.
What really worries me is that no test are failing, showing that the test coverage is too low.
Especially as it is about reading and writing files...

I guess I can test it tomorrow.

PS: Squashing commits in PRs is nice if the code was not yet read by some other dev.
It is just easier to look into newer commit and check these than checking the one commit that was changed :)
We can also squash while merging.

src/main/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackWriter.java
rgmf: If it's not valid then Marker not show the photo (obviously) because the importer expect that the "last path segment" was valid (we would have a broken link).

I don't know how importer can check it because in the importing process firstly it creates the URI and later copy images to internal memory.

The same for the exporter: firstly creates the URL and then copy the files inside "images".

dennisguse: So, we should first the get all files and the parse the KML, right?
Then we can copy the used files into the app-external directory.

For now, I guess, we should just apply a your fix now (then release v3.6) and then rework this properly afterwards.
About the writer: I would prefer to be careful as we never now what is actually in the database (and what some previous OpenTracks version actually did).
@rgmf What do you think?

rgmf: > For now, I guess, we should just apply a your fix now (then release v3.6) and then rework this > properly afterwards.

Ok @dennisguse 

> So, we should first the get all files and the parse the KML, right?
> Then we can copy the used files into the app-external directory.

Checking that URLs are valid ones need more time.

Doing so in the **importer** seems feasible.

But in the exporter seems more difficult because it writes the kml file first and if it success then write images. If you write images firstly and then you write kml file then you have to delete images if kml file couldn't be written. Anyway I have to deep on that.

> About the writer: I would prefer to be careful as we never now what is actually in the database (and what some previous OpenTracks version actually did).

Writer get track, trackpoint and waypoint from database and write them out to the kml file. The only transformation I do is the photoUrl value. In the database is a file:// or a content:// and in the kml file will be a relative path with the same name.

Export works with file:// and content://

Import now generates only content://

Why do you think we should careful with the writer?

dennisguse: About the exporter: in the future, we might change photoUrl to just contain the actual filename and not storing the whole content-URL.
And existing installations of OpenTracks might contain some weird things in their databases - there weren't to many changes regarding photoUrl. 
I guess the biggest one was migration to Storage Access Framework.
So, basically we should keep it in mind to not rely too much on the current implementation, but also expects some quirks introduced by previous versions.

rgmf: Ok, tomorrow I'll open an Issue to write down all these changes and I'm going to work on it.

dennisguse: And there is another change that affected photoUrl (use of a FileProvider to get images from the camera):
https://github.com/OpenTracksApp/OpenTracks/commit/828a19aea182672374534368a8c09a439c0167f6

It was released with the very first OpenTracks version (v3.0.0).

dennisguse:
@rgmf I just wanted to add a commit to this PR, but I don't have access rights....
I pushed it to `origin` into branch: `rgmf-issue#166`.
It just checks the photoUrl while importing.

rgmf:
> @rgmf I just wanted to add a commit to this PR, but I don't have access rights....
> I pushed it to `origin` into branch: `rgmf-issue#166`.
> It just checks the photoUrl while importing.

It's great :)

Do you want I add the rest in #180?

dennisguse:
@rgmf can you cherry pick it the commit into this PR?

PR #182, opened 2020-04-11
TrackRecordingService updates a recording track's total time every second

dennisguse:
This change moves the timer to update the UI every second from `StatsFragment` to the `TrackRecordingService`.
The timer executes an update of the total time, and this change is then propagated via the `TrackDataHub` to `StatsFragment`.

__Why:__
1. In preparation to implement _elevation gain from barometer_, `TrackRecordingService` needs a timer to update the elevation gain on a regular interval (besides adding new `TrackPoints`).
2. `TrackController` and `StatsFragment` use their own timer and can get out-of-sync due to pause/resume.
3. OSMDashboard was only informed of track updates when a new `TrackPoint` is inserted.
With the change, it gets an update every second.

__Potential drawbacks:__ one SQL update on the track-table every second (even while remaining stationary).
The SQLite database might also behave differently.

@rgmf @pstorch Do you have any thoughts on this?

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/fragments/StatsFragment.java
+++ b/src/main/java/de/dennisguse/opentracks/fragments/StatsFragment.java
@@ -18,7 +18,6 @@ package de.dennisguse.opentracks.fragments;
 
 import android.content.SharedPreferences;
 import android.os.Bundle;
-import android.os.Handler;
 import android.util.Log;
 import android.util.Pair;
 import android.view.KeyEvent;
@@ -49,7 +48,6 @@ import de.dennisguse.opentracks.util.LocationUtils;
 import de.dennisguse.opentracks.util.PreferencesUtils;
 import de.dennisguse.opentracks.util.StringUtils;
 import de.dennisguse.opentracks.util.TrackIconUtils;
-import de.dennisguse.opentracks.util.UnitConversions;
 
 /**
  * A fragment to display track statistics to the user.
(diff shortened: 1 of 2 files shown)


rgmf:
> In preparation to implement elevation gain from barometer, TrackRecordingService needs a timer to update the elevation gain on a regular interval (besides adding new TrackPoints).

> Potential drawbacks: one SQL update on the track-table every second (even while remaining stationary).

@dennisguse is it possible to move the gain from Track-table to TrackPoint-table? At each TrackPoint the gain is saved up to that moment.

In that case the gain is in the last TrackPoint of the Track and you have not to update Track-table every second.

> TrackController and StatsFragment use their own timer and can get out-of-sync due to pause/resume.

I don't know if it is possible to synchronize the two TextView from differents fragments although they use the same timer source because the time in propagation. What do you think? Is it possible?

pstorch:
I would also prefer to only save the  elevation (last one) when a new Trackpoint is saved.

dennisguse:
Thanks for your input.
Then I propose the following: move the timer into `TrackDataHub` (to sync `StatsFragment` and `TrackController`).
And we keep the current behavior to update the `TrackStatistics` on inserting a new `TrackPoint`.
Then I will update the `TrackController` to also use a `TrackDataHub` and we have only one data flow within OpenTracks.

For the elevation gain: I would prefer to save the data (for the time being) into the `TrackStatistics` as everything is already in place.
I will just use a separate `Handler` to trigger computation of the elevation gain with in the `PressureSensorManager` (to be implemented).
Elevation gain-related branch: https://github.com/OpenTracksApp/OpenTracks/tree/barometer%2310

Issue #180, opened 2020-04-09
Improve import/export all tracks


rgmf:
### Is your feature request related to a problem? Please describe.
When user imports kmz files with photos there is a potencial problem: if there are waypoints photos then the URL can be invalid and the code doesn't check anything about this now.

This is likelihood because is an external file (it's outside OpenTracks) so we can't do any assert.

For example, you could have this kmz file:
```
9.kmz
  |__ images
        |__ beach.jpeg
  |__ doc.kml
```
And in the `doc.kml` the URL to the image inside images could be wrong, for example:

```kml
<href>images/mountain.jpeg</href>
```

If you import a file like this, the importer doesn't complain but we'll have a broken link and the image won't be show.

### Describe the solution you'd like
**In the importer (critical):**
It's critical because we are importing an external file. Importer should check that the href in the kml file points to a real image inside kmz file structure.

If importer run into with a broken link then it should go on without image.

**In the exporter**
After https://github.com/OpenTracksApp/OpenTracks/pull/178 the exporter creates an href with a relative path to the images inside kmz file.

In the exporter we should be careful as we never now what is actually in the database (and what some previous OpenTracks version actually did).

Now every waypoint in the database with a photoUrl value results in a `<PhotoOverlay>` in the kml file. But it's necessary to check if this photoUrl point to a valid image inside OpenTracks external storage before creates a `<PhotoOverlay>`.

In the database you can find file:// and content:// because https://github.com/OpenTracksApp/OpenTracks/issues/166 issue.

**Attribute photoUrl**
It uses an attribute called `photoUrl` throughout the importer / exporter which can be misleading because it is used to save the `waypoint.photoUrl` and to save the path relative to the photo in the kmz file.

**Tests**
~~Maybe~~ we should add more test.

### Additional context
- See https://github.com/OpenTracksApp/OpenTracks/pull/178 discussion.
- There is another change that affected photoUrl (use of a FileProvider to get images from the camera): 828a19a

dennisguse:
@rgmf I just wanted to add a commit to this PR, but I don't have access rights....
I pushed it to `origin` into branch: `rgmf-issue#166`.
It just checks the photoUrl while importing.

dennisguse:
One question is actually, if we should consider the KMZ internal file name at all.
We could also generate a one within OpenTracks.

rgmf:
> One question is actually, if we should consider the KMZ internal file name at all.
> We could also generate a one within OpenTracks.

[...]

PR #188, opened 2020-04-16
Improve import/export

rgmf:
### Describe the pull request
**For KMZ importer**
- First: It reads the images inside KMZ file and copy them to OpenTracks external storage.
- Second: It parses doc.kml that is inside KMZ file checking that href in waypoint's tag points to an existing image in OpenTracks storage. If exists it creates a marker with photo. Otherwise only the marker.

It needs to open two times KMZ file. The first one to find images and the second one to parses doc.kml inside KMZ.

**For KMZ exporter**
- It creates the waypoint with image in the doc.kml file only if this image exists inside OpenTracks external storage. Otherwise it creates the waypoint in the doc.kml file without href.
- When it generates description for kml file it adds min/max elevation only if these data are valid.
- Added tests for this last point.

**A little bug**
The method `createTrackPoint` in `AbstractFileTrackImporter` can return null. In that case a `SAXException` must be raised because that means that `org.xml.sax` couldn't parse a point in the file.

I found this bug doing test with GPX files. The App before cracked and now informs to user that there was an error parsing a file.

### Link to the the issue
#180 

### License agreement

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackWriter.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackWriter.java
@@ -32,6 +32,7 @@ import de.dennisguse.opentracks.content.data.TrackPoint;
 import de.dennisguse.opentracks.content.data.Waypoint;
 import de.dennisguse.opentracks.content.provider.ContentProviderUtils;
 import de.dennisguse.opentracks.content.sensor.SensorDataSet;
+import de.dennisguse.opentracks.util.FileUtils;
 import de.dennisguse.opentracks.util.StringUtils;
 
 /**
@@ -162,7 +163,8 @@ public class KmlTrackWriter implements TrackWriter {
     @Override
     public void writeWaypoint(Waypoint waypoint) {
         if (printWriter != null && exportTrackDetail) {
-            if (waypoint.hasPhoto() && exportPhotos) {
+            boolean existsPhoto = FileUtils.getPhotoFileIfExists(context, waypoint.getTrackId(), waypoint.getPhotoURI()) != null;
+            if (waypoint.hasPhoto() && exportPhotos && existsPhoto) {
                 float heading = getHeading(waypoint.getTrackId(), waypoint.getLocation());
                 writePhotoOverlay(waypoint.getName(), waypoint.getCategory(), waypoint.getDescription(), waypoint.getLocation(), waypoint.getPhotoUrl(), heading);
             } else {
--- a/src/main/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackWriter.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackWriter.java
@@ -166,7 +166,7 @@ public class KmlTrackWriter implements TrackWriter {
             boolean existsPhoto = FileUtils.getPhotoFileIfExists(context, waypoint.getTrackId(), waypoint.getPhotoURI()) != null;
             if (waypoint.hasPhoto() && exportPhotos && existsPhoto) {
                 float heading = getHeading(waypoint.getTrackId(), waypoint.getLocation());
-                writePhotoOverlay(waypoint.getName(), waypoint.getCategory(), waypoint.getDescription(), waypoint.getLocation(), waypoint.getPhotoUrl(), heading);
+                writePhotoOverlay(waypoint, heading);
             } else {
                 writePlacemark(waypoint.getName(), waypoint.getCategory(), waypoint.getDescription(), WAYPOINT_STYLE, waypoint.getLocation());
             }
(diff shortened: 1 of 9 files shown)


dennisguse:
Looks good to me.
Some thoughts:
* What happens if the KMZ contained an image that is not referenced from the kml? (are those removed afterwards?)
* What happens if track import fails due to KML issues?
* What happens if the images are not in the images/ folder, but correctly referenced from KML?
  (somebody renamed the folder or added an image manually/using another program)

rgmf:
>     * What happens if the KMZ contained an image that is not referenced from the kml? (are those removed afterwards?)

No, they aren't. They are copied to OpenTracks external storage.

Do you think they should be deleted? If so, KmlImporter should return an array with all images referenced.

>     * What happens if track import fails due to KML issues?

If KML parse fails then `cleanImport` is called in `KmzTrackImporter` (it's a method that was already created). In that method track and images are deleted.

>     * What happens if the images are not in the images/ folder, but correctly referenced from KML?
>       (somebody renamed the folder or added an image manually/using another program)

The code look for images inside "images" folder. If they are in other folder then this Markers will be created without images.

The folder ("images") where importer is looking for is handcoded in `KmzTrackExporter.KMZ_IMAGES_DIR`.

dennisguse:
@rgmf: what is your opinion about this?
Should we merge this (at least it fixes an acute problem) and fix the potential issues afterwards?

rgmf:
> @rgmf: what is your opinion about this?
> Should we merge this (at least it fixes an acute problem) and fix the potential issues afterwards?

Here https://developers.google.com/kml/documentation/kmzarchives#recommended-directory-structure recommends a file called doc.kml and folders with descriptive names. So images should be in a folder called "images". Because of that in the code there are constants that save the name of the kml file inside KMZ (doc.kml) and folder name for images (images).

So I think all is ok. Nobody should change by hand any kmz file, isn't it?

I don't know who coded this parte but I like how it is.

dennisguse:
@rgmf I don't fully agree, because some application might create a "pictures"-folder or might even produce multiple folders (hypothetical).
I guess in 99.99% this won't be a problem - I am just worried about the edge cases.

@rgmf Anyhow, your call :)

rgmf:
> @rgmf I don't fully agree, because some application might create a "pictures"-folder or might even produce multiple folders (hypothetical).
> I guess in 99.99% this won't be a problem - I am just worried about the edge cases.

Ok, so let me to achive 100% jajaja.
I'm going to change the code for doing it flexible to any folder-structure ;)

rgmf:
**After these last changes**
- Importer look for images inside KMZ file (wherever they are).
- Importer clean orphan images after importing process (don't let orphan images in external storage without).
- Importer guarantee backward compatibility. It imports files generated with OpenTracks versions before 3.5.0 (with content:// and file://). Anyway, some day this tricky code must be deleted (it's marked in code with a TODO comment).
- Exporter use waypoint id as a name for marker's images.
- Lastly, a minor bugs are fixed (I've tested with three different phones and with more than 500 GPX/KMZ tracks files and I found some bugs because not null checking).

**The only handicap I find**
All this import/export process depend on file's extension. If user delete the extension of an image inside KMZ file then this image won't be imported. That's why code use filename/extension to find images files but I think is so much expensive to open every file for reading headers and check if every file is an image.

dennisguse:
@rgmf Can you rebase it onto master?

src/main/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackWriter.java
```diff
@@ -322,32 +324,29 @@ private void writePlacemark(String name, String category, String description, St
...
-     * @param heading     the heading
+     * Writes a photo overlay from waypoint.
+     *
+     * @param waypoint Waypoint object.
+     * @param heading  the heading.
      */
-    private void writePhotoOverlay(String name, String category, String description, Location location, String photoUrl, float heading) {
-        if (location != null && exportTrackDetail) {
+    private void writePhotoOverlay(Waypoint waypoint, float heading) {
+        if (waypoint.getLocation() != null && exportTrackDetail) {
```
dennisguse, line 333: Waypoint.getLocation() cannot be null as it is required during creation.
Why should it here be different?

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
```diff
@@ -42,58 +46,197 @@
 
     private static final String TAG = KmzTrackImporter.class.getSimpleName();
 
+    public static final List<String> KMZ_IMAGES_EXT = new ArrayList<>();
```
dennisguse, line 49: Arrays.asList()?

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
```diff
@@ -42,58 +46,197 @@
...
+
+    /**
+     * Copies all images that are inside KMZ to OpenTracks external storage.
+     *
+     * @return false if there are errors or true otherwise.
+     */
+    private boolean copyKmzImages() {
+        try (InputStream inputStream = context.getContentResolver().openInputStream(uriKmzFile);
+            ZipInputStream zipInputStream = new ZipInputStream(inputStream)) {
+            ZipEntry zipEntry;
```
dennisguse, line 103: Move into while?

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
```diff
@@ -42,58 +46,197 @@
...
+        if (fileName == null) {
+            return false;
+        }
+
+        String fileExt = FileUtils.getExtension(fileName.toLowerCase());
+        if (fileExt == null) {
+            return false;
+        }
+
+        for (String ext : KMZ_IMAGES_EXT) {
```
dennisguse, line 161: KMZ_IMAGES_EXT.contains(fileExt)

src/main/java/de/dennisguse/opentracks/util/FileUtils.java
```diff
@@ -95,11 +95,15 @@ public static synchronized String buildUniqueFileName(File directory, String fil
 
     /**
      * Gets the extension from a file name.
-     * Returns null if there is no extension.
+     * Returns null if there is no extension or fileName is null.
```
dennisguse, line 98: @return

dennisguse:
@rgmf Just by quickly reading it: looks good.
I will try out in more detail in afternoon.

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
rgmf: What do you mean to move into?  ZipEntry declaration?

rgmf:
> @rgmf Just by quickly reading it: looks good.
> I will try out in more detail in afternoon.

@dennisguse  when you can, no hurry ;)

dennisguse:
@rgmf Looks good - merge how you want it to be merged :)

Issue #16, opened 2019-08-09
Bluetooth sensors: add multiple sensors


dennisguse:
Add support for recording the speed via a Bluetooth LE Speed Sensor and use this to show the current speed while recording.
Especially for cycling GPS based speed is to imprecise.

Speed data should also be recorded in the database.

dennisguse:
Status:
* connection to multiple sensors works (needs big refactoring still)
* cycling sensor (crank+wheel rotation)
  * connection and also raw-data parsing works
  * cadence computation works
  * settings to select sensors (cadence / speed) implemented

Todo: 
* setting for wheel circumference is necessary for speed computation.
* refactoring SensorDataSet to contain all the measured data
  I guess, we need to implemen SensorData for per Sensor and the SensorDataSet is just collection of it.
* Extend StatsFragment

Time frame: I guess I can provide a prototype by the end of this week (containing Cadence only).

__Note:__ although the _one_ cycling sensor _may_ provides crank and wheel revolutions, often two separate ones are deployed.
Each sensor only provides one measurement - so in this case we need to connect to two different sensors, which both propose to provide exactly the same data (from a GATT perspective), but while reading only one is provided.
See: https://de-eu.wahoofitness.com/devices/bike-sensors

PR #193, opened 2020-04-25
Bluetooth LE: add cycling cadence

dennisguse:

**Describe the pull request**
This PR adds support for Bluetooth LE cycling sensors (for now cadence only; speed is prepared but requires wheel circumference in settings).
OpenTracks now support connecting to multiple sensors at the same time.

Tested with Wahoo CADENCE.
Sadly, Wahoo CADENCE reports wheel revolutions instead of crank; and thus a workaround is in place that checks if sensor that should reported cadence reports speed instead.
Who needs standards, right?

I also did some refactoring.
`SensorDataSet` is now used to pass data from sensors via TrackRecordingService to `StatsFragment` only.
The `TrackRecordingService` extracts the sensor data (measurement only) and adds it directly to a `TrackPoint` (saved to DB).
`CharTview` thus now shows data directly from `TrackPoint`.

**Link to the the issue**
#16 

**License agreement**


dennisguse:
From my side the code is ready.
It is working in itself, but i could not yet test it with real sensors (just indoor and without any comparison).
So, there might still be some computational issues included.

However, as the structural changes are rather severe (touching a lot of files), so I would rather merge the code soonish.

@rgmf @pstorch It  would be awesome if you could do some beta testing that I did not break anything (I hope for the best).

rgmf:
> From my side the code is ready.
> It is working in itself, but i could not yet test it with real sensors (just indoor and without any comparison).
> So, there might still be some computational issues included.
> 
> However, as the structural changes are rather severe (touching a lot of files), so I would rather merge the code soonish.
> 
> @rgmf @pstorch It would be awesome if you could do some beta testing that I did not break anything (I hope for the best).

Ok @dennisguse. Count on me! ;)

dennisguse:
Rebased onto master to include GPS fix.

PR #201, opened 2020-04-29
Fix compatibility issue in styles_custom.xml

john2020star:
Adding indeterminateTintMode="src_in"

**Describe the pull request**
I have found there is a compatibility issue in styles_custom.xml. You forget to add `android:indeterminateTintMode="src_in"`, which can cause the color of ProgressBar not consistent when API Level < 23. This pull request help fix this issue.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/util/DialogUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/util/DialogUtils.java
@@ -85,7 +85,7 @@ public class DialogUtils {
      * @param formatArgs       the format arguments for the message id
      */
     private static ProgressDialog createProgressDialog(boolean spinner, final Context context, int messageId, DialogInterface.OnCancelListener onCancelListener, Object... formatArgs) {
-        final ProgressDialog progressDialog = new ProgressDialog(context);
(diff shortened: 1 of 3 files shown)


dennisguse:
Nice!
Do you have some insights why this is needed?

john2020star:
Actually it is a bug of Android framework. [https://issuetracker.google.com/issues/37035006](url)

In API Level 22, the code in ProgressBar.java looks like this.
`if (a.hasValue(R.styleable.ProgressBar_indeterminateTint)) {
    // ...
    mProgressTintInfo.mIndeterminateTintMode = Drawable.parseTintMode(a.getInt(
                    R.styleable.ProgressBar_indeterminateTintMode, -1), null);
    // ...
}`
If you only set indeterminate, then the indeterminateTintMode will be wrongly applied with the default value even it is not set in the layout xml file, and it was fixed in API Level 23.

john2020star:
I have added a new theme for the dialog to help you actually change the color of ProgressBar in the previous commits, because you need to manually set the style to ProgressDialog before API 22.

To test this, try to delete a track in API Level 22.
Before
image

After
image

It works well in API Level >= 23

dennisguse:
+1

Issue #171, opened 2020-03-28
Split StatsFragment


dennisguse:
StatsFragment is at the moment used while recording as well as for finished tracks.
Depending on the situation some of the TextViews are hidden (e.g., heart rate for finished tracks).

For a nicer UI (and easier maintenance) it would be beneficial to split StatsFragment into two (one for recording; one for finished tracks).

PR #203, opened 2020-04-30
Manually starting GPS and opening a recorded track bug fixed. Fixes #200.

### LINKED ISSUE #200: Manually starting GPS and opening a recorded track
__Actual question:__ do we need the functionality to start the GPS before a recording is started? (the GPS icon in the action menu; disabled from v3.7.+).

IMG_20200429_212216

**To Reproduce**
1. No recording is running
2. TrackListActivity: go to menu and start click start GPS
3. Open an already recorded track
4. TrackDetailActivity opens, but shows UI for recording a track
(should just open and just present the recorded track in non-recording mode).

Bug was introduced while adding the "resume track"-feature.

**Technical information**
 - Version v3.6.0 and upwards

rgmf:
**Describe the pull request**
I've fixed the #200 

Only starts a new record if "someone" request it (resume track option is clicked or record button is clicked).

**Link to the the issue**
#200 

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/TrackListActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/TrackListActivity.java
@@ -205,7 +205,8 @@ public class TrackListActivity extends AbstractListActivity implements ConfirmDe
             if (!PreferencesUtils.isRecording(recordingTrackId)) {
                 // Not recording -> Recording
                 updateMenuItems(false, true);
-                Intent newIntent = IntentUtils.newIntent(TrackListActivity.this, TrackDetailActivity.class);
+                Intent newIntent = IntentUtils.newIntent(TrackListActivity.this, TrackDetailActivity.class)
+                        .putExtra(TrackDetailActivity.EXTRA_RECORDING_REQUEST, true);
                 startActivity(newIntent);
             } else if (recordingTrackPaused) {
                 // Paused -> Resume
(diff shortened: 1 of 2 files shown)


src/main/java/de/dennisguse/opentracks/TrackListActivity.java
```diff
@@ -205,7 +205,7 @@ public void onClick(View v) {
             if (!PreferencesUtils.isRecording(recordingTrackId)) {
                 // Not recording -> Recording
                 updateMenuItems(false, true);
-                Intent newIntent = IntentUtils.newIntent(TrackListActivity.this, TrackDetailActivity.class);
+                Intent newIntent = IntentUtils.newIntent(TrackListActivity.this, TrackDetailActivity.class).putExtra(TrackDetailActivity.EXTRA_RECORDING_REQUEST, true);
```
dennisguse, line 208: That is somehow implicit and hard to see.
Can we somehow make this easier to read?

rgmf: I've reviewed code before resume option and I've followed the same approach. Before, TrackListActivity used a flag called `startNewRecording` to know if user had been clicked on record button to start a recording, so if user clicked on Start GPS button it has not start recording because `startNewRecording == false`.

Now, TrackDetailActivity uses a flag called `isRecordingRequest` to know if it has to start a recording. So, if an Activity wants to start a new recording must starts TrackDetailActivity passing true in the extra argument EXTRA_RECORDING_REQUEST.

Then, if someone want to use TrackDetailActivity should understand what EXTRA_ data is for. Yes, that become code a harder one.

TrackListActivity and TrackDetailActivity need to "speak" each other. The other option is doing so through Service and the logic become harder.

I can't think of anything better, some clue may help. Anyway, I can give this a spin.

---

Offtopic: Maybe https://github.com/OpenTracksApp/OpenTracks/issues/171 is an option to simplify all this. With TrackListActivity, TrackDetailActivity and TrackRecordActivity could be more simple.

dennisguse: I just meant the long line :)
It is so easy to overlook this additional putExtra.

#171: I totally agree.

dennisguse:
@rgmf Nicely done; merge whenever it feels right :)

dennisguse:
Should we create a new release?

rgmf:
> Should we create a new release?

Yes, you can create a new release to include it. It works, I tested it yesterday and all works well.

Issue #159, opened 2020-03-13
Use track name when exporting


dennisguse:
Follow up to #130.

rgmf:
@dennisguse, Why do you think the name of exported files are important? I mean, users only use import/export for take their activities from one place to another or for backup purpose, aren't they?

And now is easy to handle non-duplication tracks when exporting.

dennisguse:
It was mentioned here https://github.com/OpenTracksApp/OpenTracks/issues/130#issuecomment-581579996

@contrapunctus-1 Can you elaborate on your requirement that the "export all" function should also use the track name instead of the trackId?

contrapunctus-1:
@dennisguse it makes it easier when opening them in a file manager or other applications - you have a better idea of which file is what, without needing to open them.

rgmf:
@contrapunctus-1 @dennisguse what do you think about using a name like this?
`<trackId>-<trackName>.<extension>`

For example, a track:
- which id is: 321
- which name is: trail running at Pyrenees

It'll have this name:
`321-trail running at Pyrinees.kmz`

It's easy change the exporting code to not duplicate when exporting looking for the id and the name is descriptive.

contrapunctus-1:
I'd suggest including the starting datetime in there (in the ISO 8601 format; OsmAnd does it too).

rgmf:
@contrapunctus-1 using a ISO 8601 prefix when exporting if there is a file with the same prefix would be overwritten, isn't it? Because if two track names coincide in ISO 8601 data time are the same track, aren't they?

dennisguse:
@rgmf This is fixed, right?

rgmf:
@dennisguse Yes, it is.

Issue #143, opened 2020-02-03
Executing tests deletes all data


dennisguse:
As observed in https://github.com/OpenTracksApp/OpenTracks/pull/133#issuecomment-581637594 it is a problem that the tests delete all data (when executed on a real device).

This happens in: 
* `TrackRecordingServiceTest`
* `TrackRecordingServiceTestLooper`

pstorch:
Is it possible to use a different database name for unit tests?

pstorch:
I executed `ContentProviderUtilsTest`

dennisguse:
Something like this might work: https://stackoverflow.com/questions/9804917/android-testing-start-with-clean-database-for-every-test/38247496

But I am out for today :)

ristello:
Hi,
is this issue still relevant?
I executed the instrumented tests on my phone. The tracks in the DB of the app had not been changed.
It seems that the instrumented tests only change the DB of the debug-app.

dennisguse:
@ristello exactly :)
I usually use the debug build and then it is quite annoying that the data is gone.
But it is not the most important thing todo.

ristello:
@dennisguse I have filed a PR which would solve that issue. The approach is described in the PR.

dennisguse:
This is fixed - thanks to @ristello!

PR #206, opened 2020-05-02
Use track start time (ISO 8601) and track name when exporting

rgmf:
**Describe the pull request**
When export all tracks the name of the files follow this rule:
`<track start time in ISO 8601 compact format>-<trackName>.<extension>`

For example, if you export to KMZ a track called "Passo dello Stelvio by bike" started at "2015-07-29T08:21:33Z" results in a file exported name like this:
`20150729082133-Passo dello Stelvio by bike.kmz`

Every time a export all process is launched exports all tracks without duplication.

The rule for not duplication is that files with the same date time and same extension are the same, so this process will overwrites it.

In this way users can export the same track with several extensions. So you can get:
`20150729082133-Passo dello Stelvio by bike.kmz`
`20150729082133-Passo dello Stelvio by bike.gpx`
`20150729082133-Passo dello Stelvio by bike.kml`

**Important**: this is not compatible with old OpenTracks installations. If a user has his tracks exported in the old way and this user exports all his tracks with the new way then this user will have track files duplicated in the storage. That's why this PR is a Draft PR.

**Tests**
Tests has been developed too.

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/159

**License agreement**

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTaskTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTaskTest.java
@@ -10,6 +10,8 @@ import org.junit.Assert;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
 import org.mockito.Mockito;
 
 import java.io.File;
@@ -32,6 +34,13 @@ import de.dennisguse.opentracks.stats.TrackStatistics;
 import de.dennisguse.opentracks.util.FileUtils;
 import de.dennisguse.opentracks.util.TrackNameUtils;
 
+/**
+ * Tests for the export async task.
+ *
+ * //TODO ATTENTION: This tests deletes all stored tracks in the database.
+ * So, if it is executed on a real device, data might be lost.
+ */
+@RunWith(JUnit4.class)
 public class ExportAsyncTaskTest {
 
     private final Context context = ApplicationProvider.getApplicationContext();
@@ -74,15 +83,11 @@ public class ExportAsyncTaskTest {
 
     @After
     public void tearDown() {
-        for (int i = 0; i < trackIdList.length; i++) {
-            contentProviderUtils.deleteAllTracks(context);
-            contentProviderUtils.deleteTrack(context, trackIdList[i]);
-        }
-
+        contentProviderUtils.deleteAllTracks(context);
         FileUtils.deleteDirectoryRecurse(exportDirectoryFile);
     }
 
-    private String addOneDay(String dateTimeStr) throws ParseException {
+    private static String addOneDay(String dateTimeStr) throws ParseException {
         Calendar c = Calendar.getInstance();
         c.setTime(DATE_FORMAT.parse(dateTimeStr));
         c.add(Calendar.DATE, 1);
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTaskTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTaskTest.java
@@ -1,0 +1,353 @@
+package de.dennisguse.opentracks.io.file.exporter;
+
+import android.content.Context;
+import android.util.Pair;
+
+import androidx.documentfile.provider.DocumentFile;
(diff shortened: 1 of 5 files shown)


src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTaskTest.java
```diff
@@ -0,0 +1,353 @@
...
+import de.dennisguse.opentracks.content.data.TestDataUtil;
+import de.dennisguse.opentracks.content.data.Track;
+import de.dennisguse.opentracks.content.data.TrackPoint;
+import de.dennisguse.opentracks.content.provider.ContentProviderUtils;
+import de.dennisguse.opentracks.io.file.TrackFileFormat;
+import de.dennisguse.opentracks.stats.TrackStatistics;
+import de.dennisguse.opentracks.util.FileUtils;
+import de.dennisguse.opentracks.util.TrackNameUtils;
+
+public class ExportAsyncTaskTest {
```
dennisguse, line 35: RunWith annotation is missing.
Is it not necessary?

Add a comment that this track deletes all data.
See #143 (also add a comment there).

src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTaskTest.java
```diff
@@ -0,0 +1,353 @@
...
+        exportDirectoryFile.mkdir();
+        exportDirectory = DocumentFile.fromFile(exportDirectoryFile);
+
+        Assert.assertTrue(exportDirectoryFile.isDirectory());
+    }
+
+    @After
+    public void tearDown() {
+        for (int i = 0; i < trackIdList.length; i++) {
+            contentProviderUtils.deleteAllTracks(context);
```
dennisguse, line 78: This should not happen in the for loop.
At least it looks weird.

src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTaskTest.java
```diff
@@ -0,0 +1,353 @@
...
+            contentProviderUtils.deleteTrack(context, trackIdList[i]);
+        }
+
+        FileUtils.deleteDirectoryRecurse(exportDirectoryFile);
+    }
+
+    private String addOneDay(String dateTimeStr) throws ParseException {
```
dennisguse, line 85: static

src/main/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTask.java
```diff
@@ -158,6 +152,42 @@ public void onProgressUpdate(int number, int max) {
...
 
+    /**
+     * Build the DocumentFile where track will be exported.
+     * The DocumentFile can be a new one or an existing one (overwriting).
+     * The file's name will be a prefix follow by the track name follow by a suffix.
+     *
+     * @param track the Track object needed to build the name.
+     * @return the DocumentFile where track will be exported.
+     */
+    private DocumentFile buildFile(Track track) {
```
dennisguse, line 163: Move to utils class (static method)?

src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTaskTest.java
rgmf: > RunWith annotation is missing.
> Is it not necessary?

I sincerely don't know, I don't notice any difference but I annotated it. I'm learning Java Testing on the way (I'm mainly Python/Django developer) :sweat_smile:


> Add a comment that this track deletes all data.
> See #143 (also add a comment there).

Why does testing use the Application Context? Shouldn't testing use Mock Context or Instrumentation Context? Maybe is this the issue? I'm really newbie in these issues so sorry if I'm saying a stupid thing :relaxed:

Anyway, commented done.

src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/ExportAsyncTaskTest.java
rgmf: Ops, sorry. Fixed ;)

rgmf:
@dennisguse this PR can be closed. It doesn't sense anymore.

Issue #209, opened 2020-05-02
Edit track while recording.


rgmf:
**Describe the bug**
When you are recording a track you can go to edit menu and change name/type/description. But name is change by date/time when you stop recording.

**To Reproduce**
1. Start new track (click on record button).
2. While is recording go to edit menu.
3. Type a name, change activity type and type a description. Then save and back to recording detail screen.
4. Stop recording.
5. You can see that name has changed by date time. Activity type is ok and description is ok too.

PR #212, opened 2020-05-02
While track is recording user can change the name.

rgmf:
**Describe the pull request**
When a recording starts it sets a name with started date time.

When recording ends It's not needed to set the name, only sets the name if track's name is empty or null.

Maybe all the code that sets a name in endCurrentTrack can be removed. In fact, I think is unnecessary. Anyway, if "something" between start and stop recording would set a null name then this would correct it.

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/209

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/services/TrackRecordingService.java
+++ b/src/main/java/de/dennisguse/opentracks/services/TrackRecordingService.java
@@ -461,12 +461,6 @@ public class TrackRecordingService extends Service {
                 // Update the recording track time
                 updateTrackTotalTime(track);
             }
-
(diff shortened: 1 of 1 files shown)


dennisguse:
Do we need to set the name at end?
I mean it should have been set when the track was started.
And even when the user changed it to "", he decided to do so.
So, what happens if we remove this completely?

dennisguse:
@rgmf: just read your description completely.
I guess we can remove it....

rgmf:
> @rgmf: just read your description completely.
> I guess we can remove it....

Yes, I was completely sure. Removed it ;)

PR #229, opened 2020-05-12
Split detail and record to simplify the code.

rgmf:
### Describe the pull request
- I've split TrackDetailActivity into TrackDetailActivity and TrackRecordActivity.
- I've split StatsFragment into StatsDetailFragment and StatsRecordFragment.
- Both activities share the same ChartFragment because there aren't any difference between detail and record (both use TrackDataHub to feed the charts).

Theses changes simplify the menu logic and record/resume logic. Also, you can create different layouts for recording and detail.

**I've added the datetime of the activity in the Detail layout**. I like to know when I record an activity so i think is useful to put it in the detail. Anyway I can remove it :sweat_smile:

### Link to the the issue
#171 

### License agreement

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -100,7 +100,7 @@ android {
 
 dependencies {
     implementation 'androidx.appcompat:appcompat:1.1.0'
-    implementation 'androidx.preference:preference:1.1.0'
+    implementation 'androidx.preference:preference:1.1.1'
     implementation 'androidx.documentfile:documentfile:1.0.1'
 
     androidTestImplementation 'androidx.test:core:1.2.0'
@@ -114,4 +114,5 @@ dependencies {
     testImplementation 'androidx.test:core:1.2.0'
     testImplementation 'org.mockito:mockito-core:3.0.0'
     implementation 'com.google.android.material:material:1.1.0'
+    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
 }
--- a/doc/activities.txt
+++ b/doc/activities.txt
@@ -9,23 +9,34 @@ from every activity showing a menu.
 endlegend
 
 [*] --> TrackListActivity
-TrackListActivity --> TrackDetailActivity : Start recording
+TrackListActivity --> TrackRecordingActivity : Start recording
 TrackListActivity --> SearchListActivity
 TrackListActivity --> MarkerListActivity : Show all markers
-TrackListActivity --> TrackDetailActivity : Show track
+TrackListActivity --> TrackRecordedActivity : Show track
 TrackListActivity --> TrackDeleteActivity
 
 SearchListActivity --> MarkerDetailActivity : Search result
 SearchListActivity --> TrackDetailActivity : Search result
 
-TrackDetailActivity --> TrackDeleteActivity
-TrackDetailActivity --> TrackEditActivity
-TrackDetailActivity --> MarkerListActivity
-TrackDetailActivity --> MarkerEditActivity : Add marker
-TrackDetailActivity --> CameraIntent : Add marker (with photo)
-TrackDetailActivity --> ShareFileIntent
-TrackDetailActivity --> DashboardIntent : Database cursor
-TrackDetailActivity --> TrackOnMapIntent : KMZ
+TrackRecordedActivity --> TrackDeleteActivity
+TrackRecordedActivity --> TrackEditActivity
+TrackRecordedActivity --> MarkerListActivity
+TrackRecordedActivity --> MarkerEditActivity : Add marker
+TrackRecordedActivity --> CameraIntent : Add marker (with photo)
+TrackRecordedActivity --> ShareFileIntent
+TrackRecordedActivity --> DashboardIntent : Database cursor
+TrackRecordedActivity --> TrackOnMapIntent : KMZ
+TrackRecordedActivity --> TrackRecordingActivity
+TrackRecordedActivity --> SettingsActivity
+TrackRecordedActivity --> TrackListActivity : Back button
+
+TrackRecordingActivity --> TrackOnMapIntent
+TrackRecordingActivity --> MarkerEditActivity : Add marker
+TrackRecordingActivity --> MarkerListActivity
+TrackRecordingActivity --> TrackEditActivity
+TrackRecordingActivity --> SettingsActivity
+TrackRecordingActivity --> TrackRecordedActivity : Stop recording
+TrackRecordingActivity --> TrackListActivity : Back button
 
 MarkerListActivity --> MarkerDetailActivity
 
--- a/doc/activities.txt
+++ b/doc/activities.txt
@@ -16,7 +16,7 @@ TrackListActivity --> TrackRecordedActivity : Show track
 TrackListActivity --> TrackDeleteActivity
 
 SearchListActivity --> MarkerDetailActivity : Search result
-SearchListActivity --> TrackDetailActivity : Search result
+SearchListActivity --> TrackRecordedActivity : Search result
 
 TrackRecordedActivity --> TrackDeleteActivity
 TrackRecordedActivity --> TrackEditActivity
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -79,7 +79,11 @@ limitations under the License.
         </activity>
 
         <activity
-            android:name=".TrackDetailActivity"
+            android:name=".TrackRecordedActivity"
+            android:screenOrientation="userPortrait" />
+
+        <activity
+            android:name=".TrackRecordingActivity"
             android:screenOrientation="userPortrait" />
 
         <activity android:name=".TrackEditActivity" />
--- a/src/main/java/de/dennisguse/opentracks/AbstractListActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/AbstractListActivity.java
@@ -26,7 +26,7 @@ import de.dennisguse.opentracks.util.PreferencesUtils;
 
 /**
  * An abstract class for the following common tasks across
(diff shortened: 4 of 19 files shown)


src/main/java/de/dennisguse/opentracks/TrackRecordActivity.java
```diff
@@ -0,0 +1,456 @@
+/*
+ * Copyright 2008 Google Inc.
```
dennisguse, line 2: Made my day :)

src/main/java/de/dennisguse/opentracks/TrackRecordActivity.java
```diff
@@ -0,0 +1,456 @@
...
+
+            @NonNull
+            @Override
+            public Fragment getItem(int position) {
+                switch (position) {
+                    case 1:
+                        return ChartFragment.newInstance(false);
+                    case 2:
+                        return ChartFragment.newInstance(true);
+                    default: //0
```
dennisguse, line 218: Why not make it 0?
And default to throw RuntimeException?

src/main/java/de/dennisguse/opentracks/util/StringUtils.java
```diff
@@ -63,6 +63,11 @@
         ISO_8601_BASE.setTimeZone(TimeZone.getTimeZone("UTC"));
     }
 
+    private static final long SECOND_IN_MILLIS = 1000;
```
dennisguse, line 66: Should be in UnitConversions.

src/main/java/de/dennisguse/opentracks/util/StringUtils.java
```diff
@@ -91,6 +96,77 @@ public static String formatDateTimeIso8601(long time_ms) {
...
+        return DateUtils.formatDateTime(context, timeInMillis, flags);
+    }
+
+    /**
+     * Returns the number of days since the epoch (January 01, 1970, 12:00 Midnight UTC).
+     *
+     * @param date date in milliseconds.
+     * @return the number of days in UTC time from the epoch.
+     */
+    public static long getDayNumber(long date) {
```
dennisguse, line 121: Why do this computation manually and not using Java's default?
Date expiry = new Date(Long.parseLong(date));

src/main/java/de/dennisguse/opentracks/util/StringUtils.java
```diff
@@ -91,6 +96,77 @@ public static String formatDateTimeIso8601(long time_ms) {
...
+    }
+
+    /**
+     * Converts the datetime into something to display to users.
+     *
+     * @param context      context object.
+     * @param dateInMillis datetime in milliseconds.
+     * @return a string with a friendly user datetime format.
+     */
+    public static String getFriendlyDateTimeString(Context context, long dateInMillis) {
```
dennisguse, line 156: Do we have test coverage for this functions?

dennisguse:
@rgmf Pretty nice!

I will just use different names for the StatsFragment as well as RecordFragement (as well as both xml files).
What about StatisticsRecordedFragment and StatisticsRecordingFragment?

src/main/java/de/dennisguse/opentracks/TrackRecordActivity.java
rgmf: :sweat_smile:

src/main/java/de/dennisguse/opentracks/TrackRecordActivity.java
rgmf: Better.
I've changed TrackDetailActivity too.

src/main/java/de/dennisguse/opentracks/util/StringUtils.java
rgmf: I did all these changes to add datetime in track detail in a human friendly way, but I really don't like so much the way it looks.

Also, layouts will be changed so I'm going to remove all this code ;)

rgmf:
> @rgmf Pretty nice!
> 
> I will just use different names for the StatsFragment as well as RecordFragement (as well as both xml files).
> What about StatisticsRecordedFragment and StatisticsRecordingFragment?

Sounds good and more descriptive :)

rgmf:
Later I'm going to fix a bug I found and push changes ;)

rgmf:
@dennisguse before Resuming feature was added there was a hidden bug (without impact on the final behaviour) but with all these changes I stumple upon it. To reproduce (before):
- Record a new Track.
- While recording go to menu and delete it.
- In the internal code, while service is deleting the track, onResume is called and TrackController is resumed unnecessarily.

Now this bug have a negative impact. This could be fixed with a boolean flag, but I think is a matter of design. I've removed delete option from TrackRecord. If user mistakes and start a track, he can stop an then he'll be in TrackDetail where he can delete the track.

With these changes TrackRecord doesn't need to extend AbstracListActivity.

What do you think? I've created several commits.

dennisguse:
@rgmf Could you rebase the changes on master?
Sadly there are some conflicts (sorry) and I don't know your intentions in detail to be quickly fix it.
Otherwise it looks quite good!

rgmf:
> @rgmf Could you rebase the changes on master?

Of course ;)

> Sadly there are some conflicts (sorry) and I don't know your intentions in detail to be quickly fix it.

No problem, I do it.

dennisguse:
@rgmf My idea was to merge this rather soonish as then I can start working on the elevation gain again and avoid another round of conflicts :)

rgmf:
@dennisguse I've moved your changes on `setHeartRateSensorData`, `setCadenceSensorData` and 
`setSpeedSensorData` from StatsFragment to RecordFragment (the conflict was there).

Rebase done ;)

dennisguse:
Just one more thing: what is the advantage of splitting TrackDetailActivity into two activities?

rgmf:
> Just one more thing: what is the advantage of splitting TrackDetailActivity into two activities?

I think the old Activity focused in couple of things: track stats (A) and track recording (B).

When A -> the activity didn't need TrackRecordingServiceConnection.
When B -> the activity needed TrackRecordingServiceConnection.

When A -> the activity didn't need TrackController and its buttons.
When B -> the activity needed the TrackController and its buttons.

When A -> the activity didn't need flags and status members like recordingTrackPaused, for example.
When B -> the activity needed members to follow the status of recording: recordingTrackId, recordingTrackPaused...

Etc.

Also, every change in the old Activity was more tricky than now. For example, if you wanted to change something with impact only in recording you could to break something on the detail part and the other way around.

Summing up: I think we had an Activity for two things and now we have a more maintainable code.

dennisguse:
I see.

Then we leave it as it is for now.
I was just thinking that we have now animations between these two activities that are not really nice (from the looks, also the state of the tab selector is not transferred or the zoom of the diagrams), but you are right - it makes the code cleaner.
And we can fix on this later ;)

rgmf:
Yes, the animation is weird because it looks a blink. I can work on transition between these two activities. Maybe It can improve.

rgmf:
@dennisguse check my last push out, what do you think something like that?

dennisguse:
Good idea.
Two thoughts about left/right:
1. we would need to apply to this to whole application (especially TrackListActivity)
2. I would go for TrackList -> (left) TrackRecordedActivity -> (left) TrackRecordingActivity and vice versa with right.

What about this: https://developer.android.com/training/transitions/start-activity

dennisguse:
@rgmf Can you also update doc/activities.txt?
It is a PlantUML file describing the flow between activities.

rgmf:
> Good idea.
> Two thoughts about left/right:
> 
>     1. we would need to apply to this to whole application (especially TrackListActivity)
> 
>     2. I would go for TrackList -> (left) TrackRecordedActivity -> (left) TrackRecordingActivity and vice versa with right.

Yes, that's logical.

I've been checking several approaches and I think with a fade animation the transition is more smooth. Tomorrow I'll push the changes.
 
> 
> What about this: https://developer.android.com/training/transitions/start-activity

I did it but those transitions was awful in OpenTracks because every item in the layout moved independently.

Some UI expert will be great here :sweat_smile:

rgmf:
After some tests, I like more a fade in/out for activity transition between recorded and recording. Is a more smooth transition and requires less changes in the code.

Also, I've updated doc/activities.txt.

It's ready for merge.

dennisguse:
Looks pretty awesome!
I love the fade-in/-out animation.

@rgmf For the constraint layout, does the preview work for you in Android Studio?
I also get some warnings (View/TextView not allowed here).

rgmf:
> Looks pretty awesome!
> I love the fade-in/-out animation.

I'm glad you like it :D
 
> @rgmf For the constraint layout, does the preview work for you in Android Studio?
> I also get some warnings (View/TextView not allowed here).

Yes it works well in Android Studio (version I'm using is 3.6.3). No issues, errors or warnings about it. It's strange, do you have any idea what can it is? I'm using Android Studio 3.6.3

I've added tools:text because this help me in design time to see fake data for TextViews where speed, avg. speed, etc, will be shown. Could it be that?

I've tested in Pixel 2 XL and Sony Xperia XZ1 with a screen with only 4.6" and the result its okay.

imagen

imagen

dennisguse:
@rgmf Works now. I needed to run gradle to download constraintlayout dependency.

So, from my side everything looks pretty good!
@rgmf If you believe it is ready, let's merge it.
What would be great, if you could squash the commits by useful chunks of work.
Especially, my two commits :)

rgmf:
> @rgmf If you believe it is ready, let's merge it.
> What would be great, if you could squash the commits by useful chunks of work.
> Especially, my two commits :)

I'll do it ;)

Issue #10, opened 2019-08-09
Present elevation gain (via barometer)


dennisguse:
GPS data is usually to imprecise to record the elevation gain (therefore it was disabled in OpenTracks).
Use the barometer to record the elevation gain.

dennisguse:
I did some research on this topic, especially the sensor fusion part.
Most notably, it is important to notice that both sensors are different:

* GPS (high variance; high accuracy)
  * provides absolute altitude (no drift)
  * absolute accuracy cannot get below 5m (more often worse)

* Barometer (low variance; low accuracy)
  * absolute accuracy can be determined using formula (requires pressure at sea level + temperature): see http://www.hills-database.co.uk/altim.html
  * drift: is affected by weather (slow changes over time; i.e., morning vs evening)
  * 
An approach for sensor fusion is described here: https://users.ece.cmu.edu/~franzf/papers/baroaltitude.pdf

dennisguse:
Another option is store the barometric (relative) ascend/descend separately for each TrackPoint in addition to the absolute GPS value.
Then the elevation gain/decrease can be easily reported and actually no sensor fusion takes place. 
Also varying varying/different intervals of the sensors can then be ignored.
However, exporting/importing the data to KML/GPX might be a problem...

dennisguse:
Current implementation approach: store the elevation gain in the `TrackStatistics`.
This does not solve the import/export problem.

dennisguse:
@rgmf Thanks for your test and suggestion.
I now added two more "enhancements" (more like try and error ><):
1. elevation changes are now limited to 5m (using the defined constant).
If an elevation change is detected, the sensor value is actually computed based (implemented invert of SensorManager.getAltitude()).
2. Exponential smoothing is applied 
(I was to lazy to implement a moving average...)

I will do some really basic (flatland) test on Tuesday.

dennisguse:
I did a 120km test today and got 250m gain while the Etrex 30 reported 450m.
But it was flat land (really no mountains), so I guess it is was okay :)
And the trip took 9hours, so weather changes are included (normal once during the day, bright sunshine without rain).

rgmf:
> I did a 120km test today and got 250m gain while the Etrex 30 reported 450m.
> But it was flat land (really no mountains), so I guess it is was okay :)
> And the trip took 9hours, so weather changes are included (normal once during the day, bright sunshine without rain).

[...]

PR #243, opened 2020-05-31
Elevation gain via barometric (pressure sensor)

dennisguse:
**Describe the pull request**
Use pressure sensor to determine the elevation gain (loss is prepared).
Pressure is polled every 3s and if an elevation change of >0.5m occurs, it is counted (no further processing, e.g., moving average is done).
Elevation gain is stored by `trackpoint` and the total stored in `trackStatistics`.

StatisticsRecordingFragment handling is not done so nicely as the elevation gain is updated every second (getting only the diff of the current trackpoint)  while the total from the trackstatistics is acquired.
This is not nice and needs further rework, but it "seems" to be working.

NOTE: Due to the lack of mountains not well-tested (in terms of: do the elevation gain computations make sense).

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/10

**License agreement**


dennisguse:
Oehm - I just tested my naive implemention (0.3Hz sensor frequency, 1m relative elevation difference threshold).
And it is way too optimistic: I got 7.000m elevation gain but in reality it was something like 200m.
Next step: add some moving average filtering.

rgmf:
@dennisguse, after several real tests, elevation gain works much better. Anyway, it's strange because I get the double of the gain all the routes have.

I fairly sure your approach is really good and I'd say that the problem is that something is duplicating the gain. Anyway it's only intuition.

For example, a couple days ago in a 800m gain route OpenTracks gave me about 1900m on live (StatsRecordingFragment). Later, after finish the activity in StatsRecordedFragment I saw about 950m.

On live, every time I see 15m quickly pass to 30m and so on. I think gain in ElevationSumManager is added to gain in TrackStatistics or the other way around.

Tomorrow I'll bike with two smartphones: one of them to test gain via barometer and the other one to test refactoring service :sweat_smile: I'm going to write down all stats and I'll tell you.

I tell you my perception in the hope of helping ;)

dennisguse:
As the ElevationSumManager only provides the elevation since the last TrackPoint was inserted, I computed the total gain the total gain from TrackStatistics + the current one from ElevationSumManager.
But I added the TrackStatstics elevation gain _twice_ - so depending on the recording interval almost the doubled amount is shown.
And after finishing a recording the TrackStatistics are solely used showing only the actual measured elevation gain.

I did some measurement yesterday (incl. this bug), but got 1600m gain while my Garmin Etrex measured 1100m (and this is about right).
So, we are still way off, but the next step would require sensor fusion between GPS and barometer.
And that is some really difficult magic.

rgmf:
Reviewing the code I saw that sometimes under some circumstances the gain was doubled.

For example, Inside `newTrackPoint` in `TrackRecordingService` we have code like this:

```java
insertTrackPointIfNewer(track, lastTrackPoint);

insertTrackPoint(track, trackPoint);
elevationSumManager.reset();
```
But here it's inserting lastTrackPoint with a gain and then trackPoint with a gain that already was added before in lastTrackPoint.
My solution was added the `setElevationGain` and the `reset` inside `insertTrackPoint`.

dennisguse:
@rgmf That sounds pretty awesome - so the code base was too complicated for me... :)
And the change looks really good!

I will give it a try tomorrow (basically on a flat terrain)!

dennisguse:
@rgmf Damn just took the PiP version with me for testing.
I will do some (limited time / within city) testing tomorrow... :|

dennisguse:
Testing done (flat plains only): I got 9m and 21m on two different phones.
So that looks okay :)

@rgmf if you think it is ready, we can merge it.

rgmf:
I've tested it in really steep mountain passes by car and by bike. Also I've been able to check the OpenTracks results with a friend's Garmin Edge 130 and they were quite acceptable (always about 10 - 30 meters away on routes between 200 and 800 gain meters).

**But all my tests have been in only one phone (Pixel 2 XL).**

@dennisguse I'd merge it and we can wait if someone send us some (bad) feedback.

dennisguse:
@rgmf Works for me - let's do what all tech companies do nowadays.
Check on user devices if the stuff is actually working ><

@pstorch Any trouble with that?

pstorch:
Sorry, haven't had time to test something

rgmf:
@dennisguse  @pstorch this morning I went out for a run and I got more than 2000 gain meters in only 7km :(

For 10 days the gain have been very good by drive, by bike and trekking, so is really strange. What happened? Why barometer is not working well for a run? We need more information about it and it's not ready yet.

dennisguse:
@rgmf did the weather change?
I mean, wie do not account for sudden drop/increase of pressure due to weather events.
I will check it tomorrow again on a relative flat terrain.

rgmf:
@dennisguse not, weather didn't change but for some reason the barometer got crazy.

When I go out for a run I put my phone in a run belt that has a really tight bag in the back and I suspect that here the barometer gets crazy. Next time I go out for a run I'll carry my phone in my hand to compare (I always run the same route).

rgmf:
Today's bike route with a "grupeta":

- A phone of mine with OpenTracks (in the bike stem): 540m.
- Another phone of mine with OpenTracks (in a back maillot bag): 513m.
- Friend 1: 597m (with Bryton GPS).
- Friend 2: 532m (with Strava).
- Friend 3: 526 m (with Garmin Edge 520).
- Friend 4: 539m (with Garmin Edge 520).
- Friend 5: 572m (with Garmin Edge 500).

The results are really good... but running is a disaster :( This week I'll test running but carrying the phone in my hands.

rgmf:
Definitely when I save my phone in my runner belt bag the phone pressure gets crazy. I'm going to figure out how to save pressure values of my phone to analize later and see what is happening and how to fix it.

dennisguse:
Fixed visibility of elevation (gain) according to settings.
Rebased onto `main`.

I guess this is as ready as it ever will be.
@rgmf Any objections?

rgmf:
@dennisguse yes, all is like ever.

In my actual tests it works like a charm by bike, by car, by walking and by hiking. I only have troubles every once in a while when I go for a run and put my smartphone in a tight pocket of my running belt.

Also, we need to do something with export/import because in this case we'll loss the gain. For that, the elevation should be computed using the gain and saving it in every track point in the db (I don't know if this is the best approach).

dennisguse:
Your idea is perfectly fine.
A quick win would be to store the elevationGain separately as the fusion of GPS and barometer is hard (might even be impossible depending on the GPS recording interval).
Then we can handle the gain like a sensor value (put it into an KML array).

dennisguse:
And you what is "just" missing in TrackPoint - persisting the `trackPoint.getElevationGain()`.
I will add this to the DB after work.

rgmf:
> And you what is "just" missing in TrackPoint - persisting the `trackPoint.getElevationGain()`.
> I will add this to the DB after work.

@dennisguse I did it in my local copy for debugging. I push it so you don't have to dedicate time in this. Also, I tested the upgrade and downgrade from 26 to 27 db versions an the other way around.

rgmf:
>     1. the index needs to be recreated: `db.execSQL(TrackPointsColumns.CREATE_TABLE_INDEX);`
>        (this is not tested...)

I don't need to recreate index in upgrading, right? Is it needed to recreate the index when add a new column? Only it's needed in downgrade after deleting and create table, right?

>     2. the data is not read from the db: see https://github.com/OpenTracksApp/OpenTracks/blob/7b454964d5859d38eaa83d15e7c58e350c10f470/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java#L648
>        (this is not tested...)

:+1:

dennisguse:
Code is ready (just needs some squashs).
But let's wait until we figure out the interval thing :)

dennisguse:
I just made the (hopefully final) fine-tuning: if no barometric data is provided, then elevation gain is null intead of 0m.
This is better as 0m is technically wrong if the device just does not have a sensor.
I tested with a Moto XT1039 (no sensor) and a Pixel1 (has sensor).

@rgmf Could you give it a spin?

rgmf:
> @rgmf Could you give it a spin?

This morning I use this branch and it worked perfect.

Yesterday I used it in a phone with no barometer sensor and I think it's a great decision the slash in gain field. Also save NULL instead of 0.

Only one issue: the horizontal line. We needed Barrier to position the horizontal lines below the lowest of two views. I push a commit to fix this.

With this it's okay for me. I think we can merge it.

dennisguse:
Thanks for testing and the fix.

@pstorch @rgmf  Should we include this in the next release or should we first release v3.9.5 containing the bugfixes only?

rgmf:
> @pstorch @rgmf Should we include this in the next release or should we first release v3.9.5 containing the bugfixes only?

IMHO bug fixes first. Then this feature could deserve to v3.10.0. You have done a goregeous work to solve one the oldest issues @dennisguse, right?

dennisguse:
Thanks for the flowers - I started back then when the dinos still roamed the earth...
@rgmf and you did a great job testing and fixing feature yourself!

dennisguse:
Just for the record: work on this PR started end of March.
So, it just took 6 month after writing the first line of code...

PR #246, opened 2020-06-02
GPS status: TrackListActivity inform about GPS status with Icon.

rgmf:
**Describe the pull request**
The icon of TrackListActivity about GPS status change depending on the status:
- **ic_gps_off** when status is GPS_NONE or GPS_DISABLED.
- **ic_gps_not_fixed** when status is GPS_ENABLED (but not yet fixed) or GPS_SIGNAL_LOST.
- **ic_gps_fixed** when status is GPS_FIXED or GPS_OKAY.

I've created a new class called GpsStatus that handle the GPS status depending on locations that it receives from Service.

Also, every class that want to use the GpsStatus needs to pass a callback to the Service Connection. From here it'll receive gps status changes. 

~~I don't know if this is the bes approach and I'd like to think about it a little more. Anyway, I create a Draft PR for any comment about it.~~

~~Also, it needs more testing. So, not ready yet.~~

PS/ This same approach can be used by TrackRecordingActivity, for example, for inform to the user the GPS status while is recording.

**Link to the the issue**

**License agreement**


rgmf:
Finally, if an Activity connects to the service through TrackActivityServiceConnect passing to it a Runnable for gps status, then the Activity can access to GpsStatus results because Service "starts" GpsStatus.

TrackListActivity uses GpsStatus through the Service and change the GPS icon for inform about three status:
1.- Disabled, not connected to GPS.
2.- Connected to GPS and fixed with bad or good signal.
3.- Connecting or fixing.

It's ready to merge.

src/main/java/de/dennisguse/opentracks/services/GpsStatus.java
```diff
@@ -0,0 +1,217 @@
...
+            return SIGNAL_LOST_THRESHOLD_DEFAULT;
+        }
+    }
+
+    /**
+     * This method must be called from service every time a new location is received.
+     * Receive new location and calculate the new status if needed.
+     * Also, it'll run the runnable if signal is bad or stop it if the signal is lost.
+     */
+    public void onLocationChanged(final Location location) {
```
dennisguse, line 111: Way too many ifs here.
Any chance to reduce this somehow?

src/main/java/de/dennisguse/opentracks/services/GpsStatus.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * @param value the preference value for recording interval.
+     * @return the preference value in milliseconds or SIGNAL_LOST_THRESHOLD_DEFAULT if the preference value is a default one.
+     */
+    private int getRecordingIntervalInMillis(int value) {
+        String minDefault = context.getString(R.string.min_recording_interval_default);
+        String minAdaptAccuracy = context.getString(R.string.min_recording_interval_adapt_accuracy);
+        String minAdaptBatteryLife = context.getString(R.string.min_recording_interval_adapt_battery_life);
+        if (value != Integer.valueOf(minDefault) &&
```
dennisguse, line 97: switch statement

src/main/java/de/dennisguse/opentracks/services/GpsStatus.java
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+
+    private Handler gpsStatusHandler = null;
+    private GpsStatusRunner gpsStatusRunner = null;
+
+    public GpsStatus(Context context, GpsStatusListener service) {
+        this.service = service;
+        this.context = context;
+        signalBadThreshold = PreferencesUtils.getRecordingDistanceInterval(context);
```
dennisguse, line 56: Pass as parameter as then GpsStatus doesn't know the preferences at all.

src/main/java/de/dennisguse/opentracks/services/GpsStatus.java
```diff
@@ -0,0 +1,217 @@
...
+    private Location lastLocation = null;
+    private Location lastValidLocation = null;
+
+    private class GpsStatusRunner implements Runnable {
+        private boolean stopped = false;
+
+        @Override
+        public void run() {
+            if (gpsStatus != null && !stopped) {
+                if (!stopped) {
```
dennisguse, line 38: unecessary if

src/main/java/de/dennisguse/opentracks/SearchListActivity.java
```diff
@@ -172,7 +172,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
         contentProviderUtils = new ContentProviderUtils(this);
         sharedPreferences = PreferencesUtils.getSharedPreferences(this);
-        trackRecordingServiceConnection = new TrackRecordingServiceConnection(null);
+        trackRecordingServiceConnection = new TrackRecordingServiceConnection(null, null);
```
dennisguse, line 175: use new TrackRecordingServiceConnection()

src/main/java/de/dennisguse/opentracks/MarkerEditActivity.java
```diff
@@ -88,7 +88,7 @@ protected void onCreate(Bundle bundle) {
 
         trackId = getIntent().getLongExtra(EXTRA_TRACK_ID, -1L);
         markerId = getIntent().getLongExtra(EXTRA_MARKER_ID, -1L);
-        trackRecordingServiceConnection = new TrackRecordingServiceConnection(null);
+        trackRecordingServiceConnection = new TrackRecordingServiceConnection(null, null);
```
dennisguse, line 91: new TrackRecordingServiceConnection()

src/main/java/de/dennisguse/opentracks/services/GpsStatus.java
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * This class handle GPS status according to received locations and some thresholds.
+ */
+class GpsStatus {
+
+    private static final String TAG = GpsStatus.class.getSimpleName();
+
+    // If recording interval not set in preferences then it uses 10 seconds like signal lost threshold.
+    private static final int SIGNAL_LOST_THRESHOLD_DEFAULT = 10000;
```
dennisguse, line 19: is there already such a preference?

src/main/java/de/dennisguse/opentracks/TrackEditActivity.java
```diff
@@ -65,7 +65,7 @@
     protected void onCreate(Bundle bundle) {
         super.onCreate(bundle);
 
-        trackRecordingServiceConnection = new TrackRecordingServiceConnection(null);
+        trackRecordingServiceConnection = new TrackRecordingServiceConnection(null, null);
```
dennisguse, line 68: use new TrackRecordingServiceConnection()

src/main/java/de/dennisguse/opentracks/services/GpsStatus.java
rgmf: In the service uses 30 seconds when user choose "Adapt for battery life" and "Adapt for accuracy":
`30 * UnitConversions.ONE_SECOND_MS`

But I think if GPS didn't receive any point for 10 seconds we can say that signal was lost. Because of that I created a constant in GpsStatus for the case the user don't select any min interval.

src/main/java/de/dennisguse/opentracks/services/GpsStatus.java
rgmf: I rewrite it because values below 1 are special meaning, so it's enough to check if the value is greater than 0.

src/main/java/de/dennisguse/opentracks/services/GpsStatus.java
rgmf: Ok, I simplify the GpsStatus removing one status because GPS_SIGNAL_OK and GPS_FIRST_FIX were the same.

dennisguse:
@rgmf is this ready?
I would have some time tomorrow to do some testing.

rgmf:
> @rgmf is this ready?
> I would have some time tomorrow to do some testing.

For me yes, but this code is related with Service and I wouldn't merged it without more @dennisguse and/or @pstorch testing.

Also, I think we are touching same files in this one and #243 one so maybe this PR could be released together #243.

dennisguse:
I will do some testing today and then we can merge.
#243 can wait another release next weekend - because it needs some practical testing that I cannot do due to the lack of mountains ;)
But let's release them one after the other, so we know which one (might) produce side effects that we did not account for (hope not, but better be safe than sorry).

dennisguse:
@rgmf somehow it does not work for me.
The GPS icon does not change to fix acquired although my phone should have a fix after 5min.

rgmf:
> @rgmf somehow it does not work for me.
> The GPS icon does not change to fix acquired although my phone should have a fix after 5min.

I cannot reproduce it neither in three phones not in emulator. All of them have a fix after the minutes set by recording time interval (samallest, 5s, 1min, 5min...).

The best option though is leave it because if it fails in a phone then it can fail in much more :worried: 

I'll try it again with more time.

dennisguse:
@rgmf I did some further tests and the icon does not change.
I put my phone on a long USB cable, so that I could get logcat while the phone can acquire a GPS fix. And it does:
```
2020-06-14 19:08:02.445 18891-19136/de.dennisguse.opentracks.debug W/TrackRecordingService: Ignore onLocationChangedAsync. Not recording or paused.
```
But the icon does not change (the one without the dot in the center remains).
I am running a LineageOS 17.1 without PlayServices.

I used a minimal recording interval of smallest and 30s.

dennisguse:
@rgmf I just removed the merge commit and rebased onto current master (was just one change).
It simplifies the git history.
But you need to delete you local branch and get it again from origin.

rgmf:
> @rgmf I did some further tests and the icon does not change.
> I put my phone on a long USB cable, so that I could get logcat while the phone can acquire a GPS fix. And it does:
> 
> ```
> 2020-06-14 19:08:02.445 18891-19136/de.dennisguse.opentracks.debug W/TrackRecordingService: Ignore onLocationChangedAsync. Not recording or paused.
> ```
> 
> But the icon does not change (the one without the dot in the center remains).
> I am running a LineageOS 17.1 without PlayServices.
> 
> I used a minimal recording interval of smallest and 30s.

I don't understand. I've done the same test with: smallest, 1s, 5s, 10s, 1min and 5min.

Leave it @dennisguse , I feel bad you are losing your time in this PR. In a few weeks I'm going to root my Pixel 2 XL and install Lineage OS. Then I'll take it back and I'll discover what's happening.

dennisguse:
@rgmf And I found the issue :)
It is **GPS week rollover**: my good old S4 reports year 2000 as GPS date and thus the comparison always fails.
In `TrackRecordingService` this is account for by doing `TrackPointUtils.fixTime(trackPoint);`.

Other feedback: 
* GPSStatus should not access locationmanager directly.
  We can register the necessary listeners from the TrackRecordingService.
* Let TrackRecordingService pass the location to GPSStatus (in onLocationChangedAsync()) rather than the TrackRecordingService.locationListener.
* skip this if not required due to naming conflicts.


And in general, we need to do some restructuring in the TrackRecordingService as we now handle a lot of stuff regarding locations in itself, in GPSStatus and then we have also the NotificationManager as well as BluetoothRemoteSensorManager.
I don't know a good solution at the moment, but I guess we need to move the complexity out of the TrackRecordingService and only handle the lifecycle of objects between start, pause/resume, and stop.
What about a LocationHandler (name to be discussed) that checks if the location needs to be inserted and also exposes the status of the GPS to the application?

PS: Let's not close this PR; let's make it a draft.
We are just not there yet ;)

rgmf:
> I don't know a good solution at the moment, but I guess we need to move the complexity out of the TrackRecordingService and only handle the lifecycle of objects between start, pause/resume, and stop.

Great idea.

> What about a LocationHandler (name to be discussed) that checks if the location needs to be inserted and also exposes the status of the GPS to the application?

That simplifies a lot of things.

PR #248, opened 2020-06-05
forward keep screen on while recording preference to dashboard

pstorch:
Needed for: https://github.com/OpenTracksApp/OSMDashboard/pull/38

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
@@ -21,6 +21,8 @@ public class IntentDashboardUtils {
 
     private static final String ACTION_DASHBOARD_PAYLOAD = ACTION_DASHBOARD + ".Payload";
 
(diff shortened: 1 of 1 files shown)


dennisguse:
@pstorch Love it.
Can you also add show on lock screen?

https://github.com/OpenTracksApp/OpenTracks/blob/1f2e3bcf70f4b94547f5041a90f40bcb06a4ccd2/src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java#L241

pstorch:
@dennisguse just in case for other dashboards, or should the OSMDashboard show something on the lock screen?

dennisguse:
@pstorch the option allows to show an activity basically above the lock screen.
Thus, it is visible without unlocking the phone.
However, this only applies if the activity was focused by the user beforehand.
If it is the background, nothing happens.

pstorch:
Added also the show on lockscreen preference. Good idea.

pstorch:
@dennisguse ok to merge?

Issue #112, opened 2019-12-26
Statistics: analyze all tracks


mwip:
This feature was removed a couple of versions ago. However, I think that it would be very useful to get a quick overview of the training effort. 

Also, it would be nice to be able to group by year, month or week. 

Do you think this would be a valuable feature?

dennisguse:
For sure - I would love to see this implementation properly!
I removed the available implementation as it was just summing up all recorded tracks: there was no differentiation by activity type, filtering or anything else.

mwip:
Awesome. I'd love to help, but unfortunately I am not much of a java developer myself. So if there is anything I could do besides coding, e.g. testing or reviewing, I'd be happy to do this.

dennisguse:
Coding is actually not that complicated in this case.
It is basically about accessing the database (via SQL) and hand the data to the UI (which is actually the complicated part).
What would be great if you could define the goals that you want to achieve (why do want statistics?) as well as a draft of a possible UI design.

Then somebody can just take and implement it.
One technical requirement: try to avoid introducing additional libraries for visualization.

dennisguse:
An API for external plugins is now available: https://github.com/OpenTracksApp/OpenTracks/commit/a3eb1070686fe3a377ab50c4a9bb1c91bd62c279

PR #271, opened 2020-06-09
Aggregated statistics by activity type.

rgmf:
**Describe the pull request**
Added new Activity that shows aggregated statistics by activity type.

I've used ViewModel and LiveData to load aggregated stats from database in a ListView. Also, I've used a custom adapter to that ListView for handle different types of sports.

It shows for now:
- Number of activities per sport type.
- Total distance and time per sport type.
- Avg. speed or Avg. pace per sport type.
- Max. speed or Fastest pace per sport type.

In the future this Activity could be three tabs: one for all times stats, other one for monthly stats and last one for yearly stats. But for the first version I think it's enough for now.

I've added tests for new AggregatedStatistics class.

And, an screenshot:
image

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/112

**License agreement**

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK">
     <output url="file://$PROJECT_DIR$/build/classes" />
   </component>
   <component name="ProjectType">
--- a/src/androidTest/java/de/dennisguse/opentracks/viewmodels/AggregatedStatisticsTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/viewmodels/AggregatedStatisticsTest.java
@@ -1,0 +1,257 @@
+package de.dennisguse.opentracks.viewmodels;
+
+import android.content.Context;
+
+import androidx.test.core.app.ApplicationProvider;
+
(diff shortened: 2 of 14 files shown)


rgmf:
I used AsyncTask. In the last splash git commit I changed AsyncTas with Java's native implementation (https://github.com/OpenTracksApp/OpenTracks/issues/273)

dennisguse:
@rgmf Why aggregate by icon and not by category (`track.getCategory()`).
This value is user changeable independent of the icon.
So, a user might decide to use the bicycle icon for MTB and race biking while settings the category differently.

dennisguse:
@rgmf Can we avoid that `AggregatedStatistics`:
1. needs the context, and
2. that the orders of key in the map is used?

Usually maps are unsorted...

rgmf:
> @rgmf Why aggregate by icon and not by category (`track.getCategory()`).
> This value is user changeable independent of the icon.
> So, a user might decide to use the bicycle icon for MTB and race biking while settings the category differently.

I was thinking about it but I decided to aggregate all in main categories (icon): bike, running, walking... But, you are right, for example if a user decides to split bike in MTB and Race Biking then OpenTracks should aggregate stats in MTB and Race Biking. I'll change it.

rgmf:
> @rgmf Can we avoid that `AggregatedStatistics`:
> 
>     1. needs the context, and

Done.

> 
>     2. that the orders of key in the map is used?

Done.

> 
> 
> Usually maps are unsorted...

Also, aggregated stats uses category instead of icon to aggregate stats (this has more sense).

dennisguse:
@rgmf Squash and merge?
I would include this change in the next release :)

dennisguse:
@rgmf Then you need to push the button :)

dennisguse:
@rgmf Wohoooo!

PR #277, opened 2020-06-12
Upgrade to Android R (API 30)

dennisguse:
As of Android R querying for installed applications is limited (I love this change).
However, we used this functionality for _show on map_.

So far, we check if OSMDashboard is installed (implicit intent) using this feature and if it was not installed, we were falling back to KMZ.

I changed this, so that we OpenTracks provides a "Open as KMZ" activity that takes the provided DashboardAPI URIs to actual KMZ URIs that are hosted by the ShareContentProvider.
When this activity is selected in the IntentChooser for the DashboardAPI, it creates a new IntentChooser for KMZ.

Works like charm in the emulator and on Android 9.

Nitpick: whenever you want to open OSMDashboard now always a IntentChooser is presented.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
 
-    compileSdkVersion 29
+    compileSdkVersion 30
 
     defaultConfig {
         applicationId "de.dennisguse.opentracks"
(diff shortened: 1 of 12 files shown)


dennisguse:
I just added the same for GPX and removed the intentchooser.
Now the target can be chosen and set as permanent selection.

PR #293, opened 2020-06-20
Restructuring TrackRecordingService (location handler, gps status...)

rgmf:
**Describe the pull request**
Restructuring TrackRecordingService.

The idea was original discussed in https://github.com/OpenTracksApp/OpenTracks/pull/246 where there is useful information.

The main idea is that TrackRecordingService only manages the lifetime of different handlers.

**Link to the the issue**

**License agreement**


rgmf:
The code need a lot of clean up and the only thing I've done is separate location listener for now.

I've created a subscribe system where service will subscribe to several handlers through handler server who know all handlers. Every handler can send messages to the service.

dennisguse:
@rgmf Nice! I just glanced over the code and it looks like a good direction!
One thing: it is probably easier to do the TrackRecordingService refactoring first and then merge the GPSStatus changes later.
I know it probably feels like duplicate work to first fix/simplify TrackRecordingService, but it is a critical component - so it would be great if we could track the changes (if we break it) easily.
Re-applying the really cool concept of GPSStatus is then basically re-implementing the TrackRecordingService-part, which we anyhow need to do.
What do you think?

rgmf:
@dennisguse I'm agree with you, in fact all code about GPSStatus is commented. The class exists but it's not used. Anyway, I'm going to delete all about GPSStatus and like you point out I'm going to go step by step.

For now: I've created the location handler that handle location listener through location manager. So this is out from TrackRecordingService.

TrackRecordingService doesn't care about this anymore:
- locationExecutorService.
- locationManager.
- locationListenerPolicy.
- locationListener.

dennisguse:
@rgmf Are you just manually reverting the GPSStatus changes?
Why not just create a new branch-based upon main?

dennisguse:
@rgmf And rebase + git push -f is far easier to read than merge commits ;)

rgmf:
> @rgmf Are you just manually reverting the GPSStatus changes?
> Why not just create a new branch-based upon main?

Yes, I am, because I was doing to things at the same time: location and GPSStatus but after yesterday conversation I rethink it and I'm going to do things step by step.

Is it better to delete this branch and begin again?

dennisguse:
For the TrackRecordingService refactoring I would recommend starting a new branch based upon main. While leaving the GPStatus one in place to continue later on (it still contains some cool changes).

rgmf:
> For the TrackRecordingService refactoring I would recommend starting a new branch based upon main. While leaving the GPStatus one in place to continue later on (it still contains some cool changes).

@dennisguse I have to create another PR with service refactoring branch and close this, right?

PR #299, opened 2020-06-22
Restructuring TrackRecordingService

rgmf:
**Describe the pull request**
Restructuring TrackRecordingService.

The idea was original discussed in #246 where there is useful information.

The main idea is that TrackRecordingService only manages the lifetime of different handlers.

**License agreement**

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/services/AdaptiveLocationListenerPolicyTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/services/AdaptiveLocationListenerPolicyTest.java
@@ -21,6 +21,8 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
+import de.dennisguse.opentracks.services.handlers.AdaptiveLocationListenerPolicy;
+
 /**
  * Tests the {@link AdaptiveLocationListenerPolicy}.
  *
--- a/src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTest.java
@@ -172,7 +172,7 @@ public class TrackRecordingServiceTest {
         // given
         createDummyTrack(trackId, -1L, true);
         TrackRecordingServiceInterface service = ((TrackRecordingServiceInterface) mServiceRule.bindService(createStartIntent(context)));
-        insertLocation(service);
+        newTrackPoint(service);
 
         // when
         service.pauseCurrentTrack();
@@ -182,7 +182,7 @@ public class TrackRecordingServiceTest {
 
         //when
         service.resumeTrack(trackId);
-        insertLocation(service);
+        newTrackPoint(service);
 
         // then
         Assert.assertTrue(service.isRecording());
@@ -201,14 +201,14 @@ public class TrackRecordingServiceTest {
         // given
         createDummyTrack(trackId, -1L, true);
         TrackRecordingServiceInterface service = ((TrackRecordingServiceInterface) mServiceRule.bindService(createStartIntent(context)));
-        insertLocation(service);
+        newTrackPoint(service);
         service.endCurrentTrack();
 
         Assert.assertEquals(1, contentProviderUtils.getTrackPoints(trackId).size());
 
         // when
         service.resumeTrack(trackId);
-        insertLocation(service);
+        newTrackPoint(service);
 
         // then
         Assert.assertTrue(service.isRecording());
@@ -294,7 +294,7 @@ public class TrackRecordingServiceTest {
         TrackRecordingServiceInterface service = ((TrackRecordingServiceInterface) mServiceRule.bindService(createStartIntent(context)));
         service.startNewTrack();
         Assert.assertTrue(service.isRecording());
-        insertLocation(service);
+        newTrackPoint(service);
         long trackId = service.getRecordingTrackId();
 
         // when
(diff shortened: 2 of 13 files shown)


rgmf:
For now: I've created the location handler that handle location listener through location manager. So this is out from TrackRecordingService.

TrackRecordingService doesn't care about this anymore:

- locationExecutorService.
- locationManager.
- locationListenerPolicy.
- locationListener.

dennisguse:
I really like your coding style :)

Three thoughts: 
1. do we still need to process the location processing in an asynchronous manner?
1. why is the workaround that checks if the TrackPoint is newer disabled?
1. who is responsible for creating/persisting the actual TrackPoint (incl the actual decision if to persist at all)?

rgmf:
>     1. do we still need to process the location processing in an asynchronous manner?

Good question. I imagine that `onLocationChangedAsync` was async because all things it did (access to database included), right? From here nothing has been changed. Anyway I'll try without async.

>     2. why is the workaround that checks if the TrackPoint is newer disabled?

I commented the part where location listener policy is updated depending on last idle time until I move all this to LocationHandler.

>     3. who is responsible for creating/persisting the actual TrackPoint (incl the actual decision if to persist at all)?

The service for now. The idea is move all this to LocationHandler that seem more logical, doesn't it?

Who do you think should be the responsible of creating/persisting the actual TrackPoint?

rgmf:
Step 2:

- LocationHandler computes all locations. If location is valid then builds a track point that send together the curren gps accuracy to the service through HandlerServer. Also gets idle time to manage location listener policy.

- Service is responsible of insert or not the track point, add sensor data to the track point, fix time, manage segments and pause.

src/main/java/de/dennisguse/opentracks/services/HandlerServer.java
```diff
@@ -0,0 +1,61 @@
...
+    public void unsubscribe(HandlerSubscriber s) {
+        if (subscriberList.contains(s)) {
+            subscriberList.remove(s);
+            if (subscriberList.isEmpty()) {
+                locationHandler.onStop();
+            }
+        }
+    }
+
+    public void sendTrackPoint(TrackPoint trackPoint, int gpsAccuracy) {
```
dennisguse, line 51: rename to recordingGpsAccuracy :)

dennisguse:
@rgmf One question: why did you create the HandlerServer?
If I got it right, it is a singleton that is (at the moment) only accessed by TrackRecordingService and passes data to LocationHandler.
However, in this process it also leaks the context as it uses it as a property to start and stop the LocationHandler.
It seems to decouple TrackRecordingService and the LocationHandler.
Do you plan to implement other users of the HandlerServer?
LocationHandler looks great.

rgmf:
@dennisguse the idea was that service subscribe to the HandlerServer to start all handlers (at this moment only one handler: LocationHandler). So the Service only has to subscribe to "one thing" instead of:

```
LocationHandler.start();
Whatever1Handler.start();
Whatever2Handler.start();
```

Also, Service will do: `HandlerServer.getInstance(this).stop()/resume()/pause()` and then HandlerServer stops/resumes/pauses all needed handler, and so on.

Story long short: HandlerServer is there for hide all handlers complexity and simplify the Service. Also, Service has to implements an only interface, the one there is in the HandlerServer.

Maybe HandlerServer is not the best name? Maybe my apporoach is wrong?

dennisguse:
Are you planning about letting other parts of OpenTracks (beyond TrackRecordingService) to interact with the handlers?
If yes, I agree, we need a solution like the HandlerServer.
I guess, I am just missing the use case at the moment - what do you have in mind?

In general, we already have a singleton in place (only one TrackRecordingService) and we could just use this one.
What about making the HandlerServer an instance that is owned by the TrackRecordingService while the HandlerServer takes care of the lifecycle of it's handlers (bascially removing the HandlerServer's singleton)?

rgmf:
> What about making the HandlerServer an instance that is owned by the TrackRecordingService while the HandlerServer takes care of the lifecycle of it's handlers (bascially removing the HandlerServer's singleton)?

:ok_hand:

dennisguse:
@rgmf Is this ready to be merged?

rgmf:
> @rgmf Is this ready to be merged?

@dennisguse Yes, it works but, what do you think about Context object management? I'm not so happy with it and I'm thinking about other possibilities about it (other pattern design? other relationships between classes involved?).

Right now: obviously the LocationHandler need the Context (and other future Handlers too) so the ServerHandler contains the Context and offer it to all Handlers through `getContext()` method.

dennisguse:
We can pass the context from TrackRecordingService whenever we actually call handler.start() and .stop(). Then the HandlerServer does not need to figure out how to get the context.

rgmf:
> We can pass the context from TrackRecordingService whenever we actually call handler.start() and .stop(). Then the HandlerServer does not need to figure out how to get the context.

LocationHandler needs the context to access to preferences in shared preferences change listener, so I think the best option is that HandlerServer saved the context.

dennisguse:
One more thing: we can make the TrackRecordingService register as a SharedPreferenceListener and forward the events to the HandlerServer, which then forwards it to it's handlers (incl. the context aka TrackRecordingService).
Then we don't need keep the reference to the context in the handlers. 
I am just trying this.

rgmf:
> One more thing: we can make the TrackRecordingService register as a SharedPreferenceListener and forward the events to the HandlerServer, which then forwards it to it's handlers (incl. the context aka TrackRecordingService).
> Then we don't need keep the reference to the context in the handlers.
> I am just trying this.

Okay, that can be a good solution :thinking:

dennisguse:
Done.
Tests are still failing as insertLocation() and another method were removed from TrackRecordingService.
Don't know yet what the best way is to correctly implement those tests.

rgmf:
> Done.
> Tests are still failing as insertLocation() and another method were removed from TrackRecordingService.
> Don't know yet what the best way is to correctly implement those tests.

`insertLocation()` is now `computeLocation` in `LocationHandler`. I had forgotten this issue in tests :(
I'll dive in it.

rgmf:
Service tests have been re-worked to adapt them to the new LocationHandler and HandlerServer. All tests related with Location have been moved to LocationHandlerTest.

rgmf:
@dennisguse Should I rebase it against main? Or do you prefer doing so at the end just before merge into main?

dennisguse:
Rebasing makes testing easier for me (all most recent changes are included), but whatever fits to your workflow.

dennisguse:
@rgmf Do you need help with this PR?

rgmf:
> @rgmf Do you need help with this PR?

@dennisguse I'm fairly sure this PR works perfectly.

I've began with GPS Status in my local computer copy. Before this weekend I'd like to push these changes, but I don't know how we should do it. Can we merge the work until now and we follow with GPS Status like a Draft? Do you think the next step should be adding GPS Status? Maybe move all the TrackPoint management from Service?

I had thought in the following steps:
- GPS Status.
- Move all TrackPoint management from Service so the service will only handle start/stop/resume/pause... in short, to simplify the logic of the service. This would be the more difficult and critial step.

dennisguse:
Then let's merge this one first - makes the upcoming changes easier to handle. GPS status sounds fine - even if it doesn't work (still hope it is not too complicated), we learned something about the internals ;)

PR #309, opened 2020-07-03
Rework track deletion UI

dennisguse:
Replaced the dialog with an overlay infinite progress bar.
Deletion can be cancelled using the back button.


src/main/java/de/dennisguse/opentracks/TrackDeleteActivity.java
```diff
@@ -44,53 +42,52 @@ public void onCreate(Bundle savedInstanceState) {
...
+        return R.layout.track_delete;
     }
 
-    /**
-     * Invokes when the associated AsyncTask completes.
-     */
     public void onAsyncTaskCompleted() {
-        removeDialog(DIALOG_PROGRESS_ID);
         setResult(RESULT_OK);
         finish();
```
rgmf, line 69: When deleting is finished the user can see the spinner going down. Why don't hide it before `finish()`?

`findViewById(R.id.progress_bar).setVisibility(View.INVISIBLE);`

src/main/java/de/dennisguse/opentracks/AbstractListActivity.java
```diff
@@ -62,17 +62,15 @@ protected void deleteTracks(long[] trackIds) {
...
+        for (long trackId : trackIds) {
+            if (trackId == recordingTrackId) {
+                stopRecording = true;
+                break;
             }
         }
+
         if (stopRecording) {
             getTrackRecordingServiceConnection().stopRecording(this, false);
         }
```
rgmf, line 76: I like the best this transition for `TrackDeleteActivity` start (below `startActivityForResult`):

`overridePendingTransition(android.R.anim.fade_in, android.R.anim.fade_out);`

dennisguse:
Uses a thread now and rebased onto master (with landscape mode).

PR #312, opened 2020-07-06
Export UUID#307

dennisguse:
Export/import the track UUID as `atom:id`.
This prevents importing the same track again as the UUID must be unique in the database.


dennisguse:
Note: code is ready, but will significantly change OpenTracks behavior.
We need to give some different user feedback if the import is not possible, because the UUID is already in the database.
At least, we need  a different error message.

dennisguse:
I assumed that we only export one track per file, but we don't.
So, cannot store the UUID in the file header, but rather in the header of each track: see `writeBeginTracks()`.
The same applies to importing.

dennisguse:
A new insight: KML and GPX support storing multiple tracks within _one_ file.
And OpenTracks supports parsing this.
Export is also possible with the existing code, but we don't use this functionality (except for sending data to other apps like OSMAnd).

This, however, implies that we need to store the UUID per track (and assume multiple tracks per file) and thus the first suggested solution won't be sufficient.
For KML export the UUID as `<folder id="UUID">`
For GPX the UUID as extension to `trk` (see https://www.topografix.com/GPX/1/1/#type_trkType).

Whatever we do here: we need to consider that the parsers are based upon SAX, so the data from a file might not be in the expected order.

dennisguse:
I am done now: the track UUIDs are now exported as a custom element (own namespace, but xml schema is not yet published).
Everything is working.
One change in behavior: if a track is already existing in OpenTracks (by UUID) and the user tries to re-import it, then it will _not_ be imported.
The user will only see the message that X out of Y tracks were successfully imported.
I guess that we should add some kind of dialog asking the user what to do: replace the track or ignore the file.

But anyhow ;)

pstorch:
> I guess that we should add some kind of dialog asking the user what to do: replace the track or ignore the file.

That can be tricky when importing 100s of Tracks. Maybe ask at the first occurrence with a checkbox `apply to all`.

dennisguse:
@pstorch Good point. 
Just implementing the dialog will not be fun as there is no callback at the moment (and the code is .... not easy to adjust).
What about a (temporary) preference setting to either ignore already existing tracks (new behavior) or generate a new UUID while still importing the track (new behavior)?
We can remove this preference, when we finally refactored the importer code.

pstorch:
@dennisguse that'll work for the time being

dennisguse:
@pstorch @rgmf Took some time, but it is ready (I hope).
Could you take a look?

rgmf:
> @pstorch @rgmf Took some time, but it is ready (I hope).
> Could you take a look?

Installed and first test okay (only export and import preventing re-import of tracks). I'll keep testing it :D

pstorch:
I tested the following:
- exported 96 tracks from OpenTracks (release)
- imported 32 tracks (because of #316)
- exported the 32 tracks again to a different directory
- looked into the SQLite DB and saw the UUID for all tracks
- imported the 32 tracks again (should be identifiable now with UUID).
  - when "Prevent re-import of tracks" is off, all Tracks got duplicated (resulted in 64 tracks in the Database)
  - when "Prevent re-import of tracks" is on, no tracks are imported because of `UNIQUE constraint failed: tracks.uuid` Exceptions

I didn't expect the duplication, because the UUID should cause the existing track with the same UUID to overwrite, correct?
Maybe it would be good to query the tracks UUID first to see if it already exists and not throw the Exception. This is necessary anyway if we want to ask the user what to do.

dennisguse:
Actually, the existing tracks are not overwritten. At the moment, unique constraint exception aborts the import and should alsonremove already existing trackpoints. Should we overwrite?

pstorch:
At least these are the two options I would want to have on a conflict:
- skip, because I already have this track (default)
- overwrite, because the one I'm importing is newer (maybe from another device I want to sync)

I can't think of a situation, where I want to have it duplicated with a new UUID. But maybe it's a third option for some users.

rgmf:
Apart to which @pstorch points out, when you import tracks that already exists Dialog says "Error..." but there aren't any error.

> At least these are the two options I would want to have on a conflict:
> 
>     * skip, because I already have this track (default)
> 
>     * overwrite, because the one I'm importing is newer (maybe from another device I want to sync)
> 

Should overwriting (the second case @pstorch points out) be sure in all cases? If so, Dialog wouldn't trigger that error.

> 
> I can't think of a situation, where I want to have it duplicated with a new UUID. But maybe it's a third option for some users.

This is already covered. If you uncheck "Prevent re-import of tracks" in Settings then tracks with same UUID will be duplicated with differente UUID according to the tests I have done.

dennisguse:
@pstorch @rgmf For now, I just wanted to reproduce the old behavior (duplication with new UUID) and 
Usabilty for this sucks, but the main problem is that we need (on unique constraint violoation) a callback from the importer that triggers a dialog with:
* do not import
* overwrite
* import with new UUID (old behavior)

And introducing the callback is not that simply ¯\_(ツ)_/¯

I will continue this in a separate PR as these changes also block (potentially #369).
Does this work?

pstorch:
I'm fine with further improvements in next Issues/PR. I would also vote to have "Prevent re-import of tracks" set to on by default.

pstorch:
Just an idea if the callback is too complicated:
What about a confirmation dialog before starting the import? 
Then the user should choose one of the three options for how to handle conflicts:
- skip (default)
- overwrite
- insert as new

pstorch:
As a bonus for this dialog it would be good to count the found files first. Then we can ask "Do you want to import these 169 Tracks?" And let the user choose the conflict resolution.

rgmf:
> Does this work?

Yes for me.

dennisguse:
@rgmf @pstorch finally done :)
That PR was bothering me for weeks now!

PR #313, opened 2020-07-08
Replacing AsyncTask with Thread

dennisguse:
This is a rough prototype to replace the AsyncTask for ImportActivity.
One real limitation for this is that on activity restart (e.g., due to orientation change), the import will start again from the beginning (and re-import files again).
Android documentation says that we need to go for a ViewModel: https://developer.android.com/guide/topics/resources/runtime-changes#RetainingAnObject
And also the Thread should be static.

Also, I guess, should the Thread be a static class.

New features:
* GPX/KML/KMZ can be imported from a folder in one run.
* prepare to handle import requests from a file application via Intent


dennisguse:
After some reading, I guess that we should go with a JobIntentService rather than a thread.
A thread may not survive the activity restart.

This actually blocks the upcoming release as orientation changes for ExportActivity, TrackDeleteActivity, and also ImportActivity will restart the task completely.
For ImportActivity this is really bad as tracks might be imported multiple times.
For the other two: it is basically waste of time and energy.

dennisguse:
If we need the context within in the background task, we should take `getApplicationContext()` instead of the host activity to prevent (short-term) memory leaks.

dennisguse:
Now the async tasks were replaced with threads.
And I learned a lot about activity versus fragment lifecycles.
In short: the fragments will retain there state while the activities get recreated (and get automatically re-attached to the then existing fragments).
Now orientation changes and other events that restart an activity will work as expected (the running task is started in again).

dennisguse:
Thanks for the feedback :)
I just check the code and also add update the documentation.
And one more thing: import/export does not grab a WakeLock anymore.
I guess, it won't be an issue and we will get one out of the box if we migrate to JobIntentService.

dennisguse:
@rgmf Found the issue: the ExportDirectoryChooser was starting ImportActivity and it should start ExportActivity.
And I also updated the PlantUML diagram.
Fixed now (git push -f).

A

rgmf:
@dennisguse I've pushed a minor changes because export didn't work due to minor bugs (I hope I'm doing it right :D)

### Review after pushing minor changes
#### Things are working for me
- Export all formats with/without screen rotation.
- Import all formats in one run.
- Deleting with/without screen rotation.

#### Issues I've stumpled upon:
- Photos are not importing.
- When OpenTracks is deleting tracks the deletion process stop if screen turns off. This is a problem if "Screen timeout" smartphone setting is short and there are a lot of tracks to be deleted.
- NOT SURE IF IT'S AN ISSUE FOR NOW: if you are deleting and press back button then the spinner vanishes but deleting process go on.

dennisguse:
@rgmf Thanks for testing.
I just re-added the wakelock to track deletion, created  #317 (abort track deletion), and set the default export format to KMZ with pictures (previously it was without pictures, which was misleading I guess).

PR #315, opened 2020-07-11
ignore invalid waypoints when importing tracks

pstorch:
I got a new phone and tried to transfer all tracks from the old phone to the new phone via `Export all` and `Import all` function.
But from 90 tracks I could only import 29.

After some debugging I found out that one track is completely discarded if one of the TrackPoints is invalid. But the Tracks were exported by OpenTracks itself. I searched a little bit more and found that most often the start TrackPoint had coordinates 180.0 and 100.0, which is considered invalid. I changed the code to "just" skip those invalid TrackPoints and was able to import all 90 tracks.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
@@ -289,7 +289,8 @@ abstract class AbstractFileTrackImporter extends DefaultHandler implements Track
         TrackPoint trackPoint = createTrackPoint();
 
         if (!LocationUtils.isValidLocation(trackPoint.getLocation())) {
-            throw new SAXException(createErrorMessage("Invalid location detected: " + trackPoint));
+            Log.w(TAG, "Invalid location skipped: " + trackPoint);
+            return;
         }
         Waypoint waypoint = new Waypoint(trackPoint.getLocation());
 
--- a/src/main/java/de/dennisguse/opentracks/io/file/importer/ImportAsyncTask.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/importer/ImportAsyncTask.java
@@ -168,6 +168,7 @@ class ImportAsyncTask extends AsyncTask<Void, Integer, Boolean> {
             }
         }
 
+        Log.d(TAG, "Importing file: " + file.getName());
         try (InputStream inputStream = importActivity.getContentResolver().openInputStream(file.getUri())) {
             lastSuccessfulTrackId = trackImporter.importFile(inputStream);
             return lastSuccessfulTrackId != -1L;
--- a/src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
@@ -178,7 +178,7 @@ public class KmzTrackImporter implements TrackImporter {
                 if (KmzTrackExporter.KMZ_KML_FILE.equals(fileName)) {
                     trackId = parseKml(zipInputStream);
                     if (trackId == -1L) {
-                        Log.d(TAG, "Unable to parse kml in kmz");
+                        Log.d(TAG, "Unable to parse kml in kmz: " + fileName);
                         return -1L;
                     }
                 }


dennisguse:
Mhh. Ignoring these coordinates might be a problem.

These specific coordinates are used to mark pause/resume as well as signal loss (then only a pause).
From a KML/GPX perspective those should be considered different - we need to check if these are actually marked this way.

First of all, these special should not be exported at all.
But to be able to restore the original data (also for synchronization), we need to recreate these TrackPoints in OpenTracks on import...

If those special coordinates are not re-created on import, then re-computation of TrackStatistics will be off and also the rendering in OSMDashboard.

dennisguse:
@pstorch I still need to check the code - i hope that we don't have such a problem.
i can try to understand (and document) how this works next week.

pstorch:
As far as I could see the `KmlTrackWriter` exports `#start`, `#end` and `#waypoint` types. But the `KmlFileTrackImporter` only reads type `#waypoint`.

dennisguse:
If I recall correctly #start and #end are just to put a special point in a map (recording was started and recording ended). I guess theire main use case is Google Earth. The belonging points are exported in addition.

pstorch:
The question is what waypoint types OpenTracks need to export and import to not loose any information.

rgmf:
In GPX files every special coordinate results in <trkseg> and in KML files special coordinate results in a <gx:track> so in files have shouldn't been invalid coordinates.

So, the issue would be any invalid coordinate was in any exported file, right?

The special coordinates are:
- lat: 100.0 and long: 0.0 for pause.
- lat: 200.0  and long: 0.0 for resume.

I think 180.0 are not an special one.

rgmf:
I forgot in my last comment a thing.

@dennisguse @pstorch  If in my last comment I'm right there wouldn't be any problem if importer ignore invalid coordinates, right?

dennisguse:
@rgmf thanks for the insight!
I am going to modify ExportImportTest to also check a case with pause/resume and signal loss (aka max distance).
But I still don't get how the lat:180 was created and why lat:100 was exported. 
Let's start digging ;)

dennisguse:
I just extended the ExportImportTest and found that for GPX/KML we create segments and we also re-create the PAUSE/RESUME trackpoints.
There is one limitation: the PAUSE/RESUME trackpoints are _not_ re-created with their original timestamp.
They are re-created using the previous TrackPoints timestamp...
I am not really sure what we should do about this.
See https://github.com/OpenTracksApp/OpenTracks/tree/importExport%23315
I created #316 as a follow-up.

Now the bad part - we have really limited test coverage here and knowledge about the data structures (and their is spread about so many parts of OpenTracks that it is really hard to understand).

Anyhow, we need to modify this PR, so that the specific locations with Lat: 100 and Lat 200 are not imported.
OpenTracks _should_ not have exported them and if we import as normal TrackPoints (will be then interpreted as pause/resume), we are in trouble.

And we need to give a lot of love to the import/export code- it is kind of hard to get.
And test coverage (also documentation) is rather sparse.

@rgmf @pstorch Thanks!

pstorch:
> Anyhow, we need to modify this PR, so that the specific locations with Lat: 100 and Lat 200 are not imported.

No problem, you can even close this PR, if you want to do it differently.

> OpenTracks _should_ not have exported them and if we import as normal TrackPoints (will be then interpreted as pause/resume), we are in trouble.

Does it mean that Pause/Resume TrackPoints will be lost after export/import?

dennisguse:
@pstorch I don't know how to fix the problem in quickly :(
You approach is okay - it just's conflicts with the internal data structure (and those might be somehow sub-optimal).
Pause/Resume is restored, but the timestamps are off.

And even one more thing: all GPS-based TrackPoints use GPS time, but pause/resume use device time (also start and stop time of a track).
That might even introduce other weird issues for export/import.

dennisguse:
@pstorch Another question: should we ignore the invalid locations or should we just import them?
I mean within OpenTracks, we are not using them (so we don't care), but then the next export would contain them as well - only in OSMDashboard it will be problematic.

And do you an idea how the invalid locations were actually created?

pstorch:
> I mean within OpenTracks, we are not using them (so we don't care), but then the next export would contain them as well - only in OSMDashboard it will be problematic.

OSMDashboard filters invalid coordinates (it cannot show them on the map). If OpenTracks import them or not doesn't affect OSMDashboard.


> And do you an idea how the invalid locations were actually created?

Have not searched for it yet.

dennisguse:
@pstorch Can we close this in favor of #316?

pstorch:
Yes, let's close this one and do it properly when fixing #316

Issue #115, opened 2020-01-04
Need new icon for markers


dennisguse:
The summary icon for markers is the KML blue push pin (`@drawable/ic_marker_blue_pushpin`).
However, this icon only available as a _low resolution PNG_ and the colors do not really fit the UI.

So, we need a replaced (best case SVG).

This is how it looks like currently:
![Screenshot_20200104-142243_OpenTracks__Debug_](https://user-images.githubusercontent.com/2820825/71766236-02b1d080-2efe-11ea-8716-9c55aa7f7649.png)

dennisguse:
We can replace it with the OpenTracks logo.

DeusFigendi:
Here I vectorized it (manually). Colors could be changed easy.
License is cc0

[(sadly I cannot upload SVG in this issue nor .gz or .zip so I used gist for it).](https://gist.github.com/DeusFigendi/ace7c40e438be9dff262168020ba6a3a)

[![KMLbluepushpin](https://user-images.githubusercontent.com/1711504/87447528-fb7d2500-c5fa-11ea-826f-7820c0fe30a6.png)](https://gist.github.com/DeusFigendi/ace7c40e438be9dff262168020ba6a3a)

dennisguse:
@DeusFigendi Nice!
It would be great if you could create a pull request and put the SVG into this folder:
https://github.com/OpenTracksApp/OpenTracks/tree/main/drawable-svg/

If you any suggestions for a different coloring - please feel free to make suggestions.

DeusFigendi:
> If you any suggestions for a different coloring - please feel free to make suggestions.

Dude, no idea, I just mentioned that recolouring is easy, because you wrote it wouldn't fit the UI.
To be honest I don't even know this project, so I don't know which color scheme would fit the needs XD.

>  It would be great if you could create a pull request and put the SVG into this folder:
sure I can do so.

dennisguse:
@DeusFigendi I have no clue about colors - I always hope that somebody else has ><
The current color is black (or a really dark gray) + orange.
Thx

dennisguse:
We now have a working SVG in #320 that was already merged.
Now, we "just" need:
* to replace the blue push pin
* check that the color fits

PR #320, opened 2020-07-15
Added an orange version of the ic_marker pushpin

DeusFigendi:

**Describe the pull request**
@dennisguse asked for an orange pushpin

**Link to the the issue**
#115 

**License agreement**
As before: cc0 for this graphic.

### DIFF
--- a/drawable-svg/ic_marker_orange_pushpin_with_shadow.svg
+++ b/drawable-svg/ic_marker_orange_pushpin_with_shadow.svg
@@ -19,7 +19,17 @@
      id="defs2">
     <linearGradient
        inkscape:collect="always"
-       id="linearGradient871">
+       id="linearGradient871"
+       
+       x1="91.712517"
+       y1="266.48581"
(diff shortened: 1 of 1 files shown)


dennisguse:
Thx!

Sadly, we cannot directly convert the SVG to the Android XML format.
AndroidStudio complains about the linear gradient (there are two: 1st the actual gradient and 2nd the a rotation of the 1st).
And the 2nd gradient cannot be imported.
I am not yet sure how to solve this - AndroidStudio complains about missing stop-infos for the 2nd gradient (which are really not there).

DeusFigendi:
Try it now, I just moved the "transform"-informations to the first gradient and removed the second. Then I set the path to use the first gradient.
It renders good in my image viewer and in inkscape and in firefox so I hope Android Studio is okay with it :D

dennisguse:
That works like charm!
I will add this properly on the weekend :)

Thanks!

PR #323, opened 2020-07-17
Added GPS Status to the service so activities can access to the GPS Status.

rgmf:
**Describe the pull request**
Added GPS Status feature.

The infrastructure is the following:

LocationHandler o-------- GpsStatus
HandlerServer    o-------- LocationHandler
HandlerServer    o-------- HandlerServerInterface (TrackRecordingService)

Every time GPS status change `HandlerServer` call `TrackRecordingService.newGpsStatus` with the new `GpsStatusValue`.

`TrackRecordingService` informs all registered listeners about changes in GPS status.

`TrackRecordingActivity` and `TrackListActivity` register as listeners through `TrackRecordingServiceBinder` / `TrackRecordingServiceConnection` to receive changes in GpsStatus.

**Tests and changes:**
- 2020-07-17: The code is only tested in AVD.

**Concerns:**
I like this feature because the GPS Status detail has a lot of future possibilities but I'm concerned about performance because `GpsStatus` needs to run a Runnable when it doesn't receive locations through LocationHandler and knowing if GPS signal has been lost.

**License agreement**


dennisguse:
I just tested this PR and works :+1: 

Some small changes:
* Can we rename BoundServiceListener interface?
  Something like TrackRecordingServiceCallback (or something else ><)
* TrackRecordingServiceBinder.`setListener` should also be `addListener`?
  Or can there only be one listener per binder (the code does not say so)?
* We also need to remove listeners (I guess on service disconnect) as otherwise we get a memory leak. Also who keeps track of the listeners? Should this also be done by the TrackRecordingServiceBinder, so it can be used to disconnect completely (incl. listeners)?

Feature request: can we somehow animate the icon while GPS is not yet fixed?

dennisguse:
And just I forgot: ServiceUtils should be removed (as it's functionality is replaced by the listener).
So far, it is still used to determine the status of the GPS icon.

rgmf:
>     * Can we rename BoundServiceListener interface?
>       Something like TrackRecordingServiceCallback (or something else ><)

Okay, it seems a better name for that ;)

> 
>     * TrackRecordingServiceBinder.`setListener` should also be `addListener`?
>       Or can there only be one listener per binder (the code does not say so)?

Yes, you are right. At first only one listener was used but I realized that more than one listener can be used (for example TrackListActivity and TrackRecordingActivity).

> 
>     * We also need to remove listeners (I guess on service disconnect) as otherwise we get a memory leak. Also who keeps track of the listeners? Should this also be done by the TrackRecordingServiceBinder, so it can be used to disconnect completely (incl. listeners)?

Okay.

> Feature request: can we somehow animate the icon while GPS is not yet fixed?

Yes, I thought about it, but I didn't know how to do with this. I'll check animation icon possibility out.

> And just I forgot: ServiceUtils should be removed (as it's functionality is replaced by the listener).
> So far, it is still used to determine the status of the GPS icon.

Okay, I'll do it.

rgmf:
Changes done for now:
- Rename BoundServiceListener interface to TrackRecordingServiceCallback.
- Move setListener to addListener and clean listeners array in TrackRecordingService.
- ServiceUtils have been removed so it doesn't need anymore. TrackListActivity uses now gpsStatusValue whose value comes from the service to know the gps status.
- Added an animated icon when GPS is ready but not fixed according to user preferences.

dennisguse:
And I love the animation!

Why not merging this one already and move the TrackRecordingActivity TODO to another PR?
This PR already contains a lot of useful work.

I will do some testing tomorrow1

rgmf:
> Why not merging this one already and move the TrackRecordingActivity TODO to another PR?

Yes, I think it's the better option. Tomorrow I'll remove TrackRecordingActivity's GPS Status code and then this PR will be ready to be tested and merged.

dennisguse:
@rgmf I would like to release a new version tomorrow, should we add this one as well?

rgmf:
@dennisguse this works for me, I've been using it all the week without problems. I think we can merge it :+1:

dennisguse:
@rgmf then push to button ;)

PR #344, opened 2020-07-31
Charts: removed interactivity, so the viewpager gets the TouchEvents.

### LINKED ISSUE #105: TrackDetailActivity: ViewPager not working on charts
**Describe the bug**
Swiping only works from stats to charts but not the other way around.

**To Reproduce**
1. Go to 'TrackDetailView' and go to one chart.
2. Try to swipe back to stats (on the chart)
3. Nothing happens

The issue lies on https://github.com/dennisguse/OpenTracks/blob/20f1524701a3ae26b74ec84e3eb822b2b6ca701b/src/main/java/de/dennisguse/opentracks/chart/ChartView.java#L477

As the touch event is consumed in `ChartView` and thus does not reach the `ViewPager` on `TrackDetailActivity`.

dennisguse:
Fixes #105.

We have a ViewPager to move on the TrackRecordActivity and TrackRecordingActivity from one page to another.
However, this only worked so far on from the statistics view to the first chart, as the charts consume touch events to provide zoom and scroll support.

This PR remove the processing of touchevents for zooming and scrolling, so these reach the viewpager.
And then one can flip between the charts and also go to the statistics via gesture.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/chart/ChartView.java
+++ b/src/main/java/de/dennisguse/opentracks/chart/ChartView.java
@@ -29,7 +29,6 @@ import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.view.GestureDetector;
 import android.view.MotionEvent;
-import android.view.ScaleGestureDetector;
 import android.view.View;
 import android.widget.Scroller;
 
@@ -51,7 +50,8 @@ import de.dennisguse.opentracks.util.UnitConversions;
 
 /**
  * Visualization of the chart.
- * Provides support for zooming (via pinch), scrolling, flinging, and selecting shown markers (single touch).
+ * Provides support for selecting shown markers (single touch).
+ * Support for zooming (via pinch), scrolling, and flinging was removed as it conflicts with the ViewPager this view is shown on.
  *
  * @author Sandor Dornbush
  * @author Leif Hendrik Wilden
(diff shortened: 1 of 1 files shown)


dennisguse:
@pstorch @mondlicht-und-sterne @rgmf Do you have any opinions on that?
Is the zooming functionality of the charts and important feature or can we kick it?

dennisguse:
Note: I stumbled upon this while figuring out how to improve landscape mode while recording.
Here, we show the TabLayout (Stats, By time, By Distance), so the user can switch between the three tabs.
But if we make the viewpager work, then we could also remove TabLayout.
I guess it does not add too much user value showing it.
And the TabLayout cannot be rotated: it is always from left to right and does not support from top to bottom.
I saw this when trying to move the TabLayout (in landscape mode) onto the right (opposite side of the pause/stop button).

rgmf:
> @pstorch @mondlicht-und-sterne @rgmf Do you have any opinions on that?
> Is the zooming functionality of the charts and important feature or can we kick it?

I've never used the zoom feature. What I love is the orientation to see the graphs in a landscape mode. Anyway, I think zoom in and out can be useful for some people.

pstorch:
For me same as @rgmf. I haven't used it, but it might be useful for someone.

dennisguse:
Thanks for the input!
Then let's not remove this functionality and will figure out an alternative way :)

Issue #186, opened 2020-04-14
How can I make a permanent automatic backup?


MrS1R1US:
Good afternoon. Tell me, I want make a permanent automatic backup. Where are training files stored to automatically duplicate them on my computer? Clicking "export" in the settings is a big time consuming.

dennisguse:
The data is stored in a SQLite database.
So, there are no files until "Export all" is triggered.
@MrS1R1US does this answer your question?

MrS1R1US:
> 
> 
> The data is stored in a SQLite database.
> So, there are no files until "Export all" is triggered.
> @MrS1R1US does this answer your question?

Can I synchronize my workouts using SQLite database synchronization?

dennisguse:
For backup purpose: yes, for synchronization: no.
I also would like to have a synchronization feature, but it is rather hard to implement and so far there have always other more important things (e.g., #10).

pstorch:
If the database is up to 25 MB and the app was installed via Google PlayStore, we could enable the autobackup feature of Android:
https://developer.android.com/guide/topics/data/autobackup

https://github.com/OpenTracksApp/OpenTracks/blob/b09eeaede342f8524e56de91019a6928996ebbf3/src/main/AndroidManifest.xml#L22

But I guess a backup is mostly requested by power users which have lots of tracks (and photo markers), then the 25 MB limit is easily crossed.

MrS1R1US:
> 
> 
> For backup purpose: yes, for synchronization: no.
> I also would like to have a synchronization feature, but it is rather hard to implement and so far there have always other more important things (e.g., #10).

In case of reinstalling the application, can I recover the data?

How do I get to this base, where does it lie? I will connect it through syncthing(app) to the PC

MrS1R1US:
> If the database is up to 25 MB and the app was installed via Google PlayStore, we could enable the autobackup feature of Android

No. F-droid.

pstorch:
Via the Storage Access Framework we could let the user choose a backup folder and copy the SQL-Lite DB there after each new Track was recorded for example. This backup directory could be one for syncthing or nextcloud, or whatever is available via SAF.

MrS1R1US:
> 
> 
> Via the Storage Access Framework we could let the user choose a backup folder and copy the SQL-Lite DB there after each new Track was recorded for example. This backup directory could be one for syncthing or nextcloud, or whatever is available via SAF.

Great) How do I use the Storage Access Framework?

[...]

Issue #90, opened 2019-12-04
Sync with another phone / webserver via NextCloud/OwnCloud


maxolasersquad:
**Describe the solution you'd like**
I would like to be able to choose a location where the tracks are saved. Ultimately I'd like to use the file chooser to save to my Nextcloud server. I do this with KeePass to keep my password file backed up and synced. The intention here is that my tracks would be backed up and synced across devices. This solution should allow automatic cloud backups without the app itself requiring Internet permissions or cloud support directly. 

**Describe alternatives you've considered**
Direct integration with Next/Own cloud would be easier, but would violate the privacy intentions of this app as I understand it.

dennisguse:
Nice idea. 
Is your goal to sync multiple devices or just backup
 the data?
What do you think about the following solution:
1.  After finishing a track (i.e., pressing the stop button), the track is exported automatically to the external storage (I.e. /OpenTracks/FORMAT/trackData)
2. Next/OwnCloud can monitor this folder and then upload the files

Then you could also automatically download to another device and import it automatically to OpenTracks. 

However, this is not implemented right now. 
One assumption in OoenTracks is that after finishing a track it won't change. 
For now this is (almost) true, but might change in the future. 
The following issues would need to be solved:
1. export must be triggered manually (in settings)
2. import must be triggered manually
3. configuration option for the user (mainly if and which file format)
4. OpenTracks does at the moment not overwrite files that were already exported (it adds a number to the filename and exports it again)

All of this is doable and I don't see any issues with integrating it this way. 

Direct integration with any Next/OwnCloud would be out of scope (you are right).
Or if you really want it, it could be implemented as a plugin.

maxolasersquad:
Both sync and backup are my goal. I'm currently using Run Keeper and have years of workout data stored with them. I'd like to self-host my workout data with a FOSS app like this one and keep it archived.

Ideally I'd like to also create a Nextcloud app that could display the workout data.

dennisguse:
Finished migration to Storage Access Framework: import/export functionality.
Does not allow

MrS1R1US:
>     1. After finishing a track (i.e., pressing the stop button), the track is exported automatically to the external storage (I.e. /OpenTracks/FORMAT/trackData)

[...]

PR #361, opened 2020-08-07
Export track after recording

thrillfall:

**Describe the pull request**
Instant post-workout gpx export to external storage

**Link to the the issue**
#186 #90 

**License agreement**

### DIFF
--- a/.idea/jarRepositories.xml
+++ b/.idea/jarRepositories.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="RemoteRepositoriesConfiguration">
+    <remote-repository>
+      <option name="id" value="central" />
+      <option name="name" value="Maven Central repository" />
+      <option name="url" value="https://repo1.maven.org/maven2" />
+    </remote-repository>
+    <remote-repository>
+      <option name="id" value="jboss.community" />
+      <option name="name" value="JBoss Community repository" />
+      <option name="url" value="https://repository.jboss.org/nexus/content/repositories/public/" />
+    </remote-repository>
+    <remote-repository>
+      <option name="id" value="BintrayJCenter" />
+      <option name="name" value="BintrayJCenter" />
+      <option name="url" value="https://jcenter.bintray.com/" />
+    </remote-repository>
+    <remote-repository>
+      <option name="id" value="Google" />
+      <option name="name" value="Google" />
+      <option name="url" value="https://dl.google.com/dl/android/maven2/" />
+    </remote-repository>
+  </component>
+</project>
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -169,6 +169,7 @@ limitations under the License.
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
         </activity>
+        <activity android:name=".io.file.importer.SingleTrackExportDirectoryChooserActivity" />
 
         <provider
             android:name=".content.provider.ShareContentProvider"
--- a/src/main/java/de/dennisguse/opentracks/SingleTrackExportThread.java
+++ b/src/main/java/de/dennisguse/opentracks/SingleTrackExportThread.java
@@ -1,0 +1,37 @@
+package de.dennisguse.opentracks;
+
+import android.util.Log;
+
+import java.io.OutputStream;
+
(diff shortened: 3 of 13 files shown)


thrillfall:
i copy&pasted a lot of the exporter code and reduced it to only work with a single track and get rid of the unused (at least in my case) dependency to ApplicationContext. This is as far as i got to make a simple gpx export work. 
Any suggestions as to how to improve this changeset are highly appreciated.

dennisguse:
Did not read the code (phone only).
Would you mind modifying your code to support the preference (default file format), so the user can set the format.

And what is your use case?

thrillfall:
My usecase is that i'd like to have my tracks (after finishing the workout) uploaded automatically to my nextcloud where i can then enjoy the tracks in the maps app.

I did not want to restrict the exportable format to only gpx in the beginning. But after reading the exporter related code i found that there are issues that i did not want to build on top off but rather duplicate parts of the code to make the smallest useful functionality (GPX export) working.
The biggest issues that i found was that FileTrackExporter and KmzTrackExporter share the same interface but FiletrackExporter   does not need the Context in writeTrack(). It seems both exporters have evolved in different ways but are still stuck to the old interface. Context should maybe be injected by constructor (in KmzTrackExporter) thus keeping the interface clean of only partialy used dependencies.
Secondly i did not understand at all how all the instantiations of the different exporter strategies where done in TrackFileFormat. 
An enums that yields static factory methods is something i have never seen before. But i am no Java developer so i assume its common.
Third and last i did not want to use the interface of FileTrackExporter that expects an array of tracks where i only have s single track to export. I'd assume that there are 2 classes (or at least methods) where one deals with exporting of a single track and the other of an array of tracks. All but one usage of the exporter pass an array of the single track that needs exporting. The only usage of actually more then one track to export was some "weird" import from file by splitting the filename by underscore.

TL;DR; it was hard for me to understand the code and use it as is in satisfying manner (mostly because i really could not grasp the complexity of the implementation as i am no java developer). So started with the smallest increment that i needed.

I subsequent increments i'd like to evolve this into a multiformat automatic export but i'd start out refactoring the current implementation (at least i'd try :smile:)

dennisguse:
Yeah, this part of the code needs some love - but you got something working :+1: 
I am at the moment working on exporting/importing the UUID of a track - I will extend the original exporter/importer code.

From my perspective, it would be better to slowly refactor the old code (I know it is really ............).

One thing: (although it is not used at the moment) OpenTracks supports reading/writing multiple tracks into one GPX/KML/KMZ. I don't know what the original use case was, but I would actually like to keep this functionality. Other applications might create files using this approach. I also was not aware that this is handled as it is implemented.

Some days ago, I created a class PlantUML diagram: https://github.com/OpenTracksApp/OpenTracks/pull/312/commits/65349d4b1e2dd011c332e64c96132ff8e1102f47
This might be helpful.

One thing about functionality: I guess it would be better to make the folder to be exported configurable (via settings) instead of asking the user every time (just make sure to persist the permissions of the URL you get from the directory chooser as otherwise a device reboot will revoke these permissions). Then you can also skip the newly implemented activity.

dennisguse:
PS: TrackFileFormat is also weird - we should do this differently.
Suggestions welcome.

thrillfall:
> One thing about functionality: I guess it would be better to make the folder to be exported configurable (via settings) instead of asking the user every time (just make sure to persist the permissions of the URL you get from the directory chooser as otherwise a device reboot will revoke these permissions). Then you can also skip the newly implemented activity.

i thought thats what i did with src/main/java/de/dennisguse/opentracks/io/file/importer/SingleTrackExportDirectoryChooserActivity.java?
I set the target directory in the settings and once the export happens it writes the file to that directory. Did i miss anything?

thrillfall:
I currently trying to refactor the "multi-track" import/export and see what can be simplified. Until then i'd close this MR and open a new one once i have a satisfying solution that can do all the formats.
Thanks for your esponses and interest in the contribution

dennisguse:
@thrillfall You are right, I misread SingleTrackExportDirectoryChooserActivity.
You are storing it as preference, but you also need to store the permission that you can write to this directory later on (after device restart).
See https://developer.android.com/training/data-storage/shared/documents-files
and search for persist permissions.
Also I think this preference should be reachable via the SettingsActivity as well.

dennisguse:
@thrillfall If you have any questions, let us know.
You can also open an issue for this.

PR #362, opened 2020-08-08
fix #241 stop button lock

### LINKED ISSUE #241: Feature Request: Stop button lock
Accidental clicking of stop. I've seen this on other running apps that require you to press an "unlock" button before stopping a recording to prevent accidental stopping of a track.  An alternative method might be requiring a long press to stop the recording.

pstorch:
Adds a delay and animation to the stop button

First time I worked with android animations. Was kind of challenge for me. Probably I did a lot wrong here. But maybe you get the idea of my proposal.


dennisguse:
I like it although it is hard to see, when I have my thumb on the button :+1: 
Should we do the same for start and pause?

+ vibration feedback would be cool, when the action is actually triggered (actually a short one when the press starts and a more impressive one when the action is triggered).
I guess, OpenTracks crashed once after pressing the button, but have no logcat on my traveling machine.

pstorch:
Found the reason of the crash, thanks for pointing it out.
I've added a short vibration at start and a long vibration at stop. Not sure if this gives the right user feedback. Please give it a try.
I've also added a Toast to inform the user to hold the button to stop the recording.

Let's discsuss if the other button also needs this delay at the ticket #241

pstorch:
Didn't know we can move the Toast around :wink: 
Now it is on top.
You can merge it, when you think it's ok.

dennisguse:
There is a bug:
1. start a recording
2. press pause (until paused)
3. press stop (until crash of activity)

```
    Process: de.dennisguse.opentracks.debug, PID: 20391
    java.lang.RuntimeException: Can't create handler inside thread Thread[Thread-4,5,main] that has not called Looper.prepare()
        at android.os.Handler.<init>(Handler.java:207)
        at android.os.Handler.<init>(Handler.java:119)
        at android.graphics.drawable.AnimatedVectorDrawable$VectorDrawableAnimatorRT.startAnimation(AnimatedVectorDrawable.java:1676)
        at android.graphics.drawable.AnimatedVectorDrawable$VectorDrawableAnimatorRT.start(AnimatedVectorDrawable.java:1617)
        at android.graphics.drawable.AnimatedVectorDrawable.start(AnimatedVectorDrawable.java:930)
        at de.dennisguse.opentracks.TrackController$ButtonDelay.run(TrackController.java:152)
        at java.lang.Thread.run(Thread.java:919)
```

pstorch:
Hmm, that I've tested :thinking: 
I'll try again. Thanks for catching it.

dennisguse:
Actually, it is not the pausing that causing this.
Seems to be some kind of race condition - I could reproduce now just by pausing.
But pause and then stop does not always reproduce this.

pstorch:
I'm going wild at the moment with starting, pausing and stopping in various situation and can't reproduce it :disappointed:

pstorch:
I wrapped the call to `AnimatedVectorDrawable.start` also in `activity.runOnUiThread`.
@dennisguse can you give it a try, if it still happens for you?

dennisguse:
@pstorch tomorrow ;)

src/main/java/de/dennisguse/opentracks/util/ActivityUtils.java
```diff
@@ -129,4 +136,20 @@ public boolean onSuggestionClick(int position) {
             }
         });
     }
+
+    public static void vibrate(Activity activity, int milliseconds) {
+        final Vibrator vibrator = (Vibrator) activity.getSystemService(VIBRATOR_SERVICE);
+        if (Build.VERSION.SDK_INT >= 26) {
```
dennisguse, line 142: Better: 
```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```

dennisguse:
+1 
@pstorch can you "squash and merge" (otherwise github changes the author and I don't want to do it locally).

PR #365, opened 2020-08-11
zoomable marker photos

pstorch:
Make the marker photos zoomable.

I just implemented it for another app and I thought it might be useful for OpenTracks as well.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,6 +14,7 @@ allprojects {
     repositories {
         google()
         jcenter()
+        maven { url "https://jitpack.io" }
     }
 }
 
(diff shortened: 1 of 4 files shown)


dennisguse:
Is this is that important that we need to add a library for this?
At the moment, we are pretty library free (not counting androidx)...

pstorch:
I somehow guessed, that this extra library will be challenged by you :wink: 
No problem, it was not much work. If you object, let's just close this PR. It's just a comfort feature, not essential.

dennisguse:
@pstorch Well played :)

Issue #134, opened 2020-01-27
While recording: report last lap time.


slackline:
It would be useful to have lap-times reported both whilst recording and in viewing recorded tracks.

The notification on the defined distance (e.g. 1km) states total time, total distance and pace, but its not clear to me if the pace is for the last 1km or overall and whilst I can do some quick maths to work out my lap time after a few km I tend to slip.  Having the lap-time reported on the distance (i.e. 5m26sec) would be useful to know if pace is being maintained.

Post activity having a breakdown of statistics (time and pace) in a table for each unit distance covered would be informative to, allowing fastest and slowest km (or other user defined distance) to be displayed.

dennisguse:
Just to clarify: do you figured what kind of pace (overall vs. last 1km)?
I am not sure what is reported at the moment, I just hope it is the most recent one.
Any insights on this (I don't have the time to check the code :|)?

About the statistics: if don't use [OSMDashboard](https://github.com/OpenTracksApp/OSMDashboard) and OSMAnd~ is installed, you can actually see detailed statistics there.
Actually, https://github.com/OpenTracksApp/OpenTracks/issues/112 is a similar request.

slackline:
I have a feeling, but no hard evidence, that currently you get the pace reported for the total distance covered.  This is based on doing a regular run and paying attention to the times on the 1km laps because my run starts with a short uphill then relatively flat for 5km, after that its gradual uphill and then it steepens and I'm sure I slow down on those sections but the pace reported only drops gradually (if it were reported as the pace per 1km lap then 6km and 7km would be really low).  At the end of the run its spot on for the total distance (~11km in 1hr2mins).

Hadn't come across OSMDashboard before but will check it out, thank you.

#112 is interesting too (I had actually read it before submitting this issue), but its really the reporting of the pace whilst recording out on a run/cycle that I find appealing.  I keep note mentally of the last total distance thats reported (i.e. "you have done 5km in 26m and 37sec at a pace of 10.9km/h") and when I hit the 6km distance work out how long its taken me to do the 6th km, but I would find it more useful having "your lap time is 5m 14sec" reported (for every 1km or other unit as specified) rather than the overall pace.

dennisguse:
Understood :)

Just one tiny but: if you time to implement this please do so (I won't be able to invest much time in OpenTracks until mid-March).
I guess it is not too difficult.
@pstorch would be able accept pull requests (I hope).

slackline:
[...]

Issue #87, opened 2019-11-30
Statistics: Show pace


sedrubal:
**Is your feature request related to a problem? Please describe.**
A common size for runners is the pace (min / km).

**Describe the solution you'd like**
OpenTrack should display it additional to or instead of the speed. You could maybe toggle between speed an pace by tapping on the speed section.

dennisguse:
Would you like to see only the pace or also the avg/min/max?

About toggling: this is usually a functionality that is hard to discover for new users.

dennisguse:
Would you like to see only the pace or also the avg/min/max?

About toggling: this is usually a functionality that is hard to discover for new users.

dennisguse:
And this is blocked by #15.

sedrubal:
I think it would be interesting to see the pace in avg/min/max or even a list of paces for each kilometer?

And you are right that toggling might not be intuitive.

dennisguse:
What do you think about not using either pace or speed but rather pace (yes/no) and speed (yes/no)?

dennisguse:
@rgmf Did we not do something about this?

rgmf:
@dennisguse no, but for runners it's an important feature (I miss it when going out for a run). In TrackRecordedActivity we should add the list of km with its pace. I'm going to look into it...

rgmf:
For run pace is showed several versions ago.

> I think it would be interesting to see the pace in avg/min/max or even a **list of paces for each kilometer**?

I'm going to check it out the possibility of show a list of pace for each km/mi because it's a feature I miss when I'm going out for a run.

dennisguse:
@rgmf I removed this functionality due to technical reasons: statistics were stored as special waypoints. It was quite a while ago (can't search the commit easily on a phone sorry), but the implementation in TrackRecordingService was cool. I just did not like the complexity of storing data. The issue to track this is #134.

rgmf:
@dennisguse my idea is not store anything but show a ListView in TrackRecordedActivity with pace by km, 2km, 3km... or whatever the user select. The pace stats by km can be extracted from TrackPoints.

Anyway I'll read #134 at home.

PR #368, opened 2020-08-12
Interval stats: distance, speed and pace by selected interval.

rgmf:
**Describe the pull request**
Added intervals to StatisticsRecordedFragment so user can analize his/her partials or intervals. User can select through spinner the length of the interval: 1km (or mi), 2km (or mi), 3km (or mi), 4km (or mi), 5km (or mi) and 10km (or mi).

Also, StatisticsRecordedFragment does't use the hub to load the stats anymore. It wasn't needed and there were a lot of methods from interface that it didn't implement.

Finally, StatisticsRecordedFragment save the track object instead of trackId because it called several times to database to load track.

2020-08-12: At this moment I create a draft PR because I'd like to create tests to test IntervalStatistics class.

intervals

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/134
https://github.com/OpenTracksApp/OpenTracks/issues/87
https://github.com/OpenTracksApp/OpenTracks/issues/87#issuecomment-561024811

**License agreement**


dennisguse:
Without testing: why not move the interval into the preferences? I guess, we will also need it for voice announcements, right?

rgmf:
> Without testing: why not move the interval into the preferences?

Because is typical analizing several partials or intervals. For example, if you run a marathon then you want to know how was every km and every 10k, for instance. So user can see all partials s-he want in a very fast way in StatsRecordedFragment screen.

> I guess, we will also need it for voice announcements, right?

The partials are only available after recording, only in StatsRecordedFragment but not in StatsRecordingFragment. Anyway, all it's ready to add this funcionality to other parts of OpenTracks (all other parts can use the view and the model created).

I was thinking on adding this feature to recorded tracks and later to recording tracks. @dennisguse do you think the best option is doing all so in the same PR?

For recording a preference with interval length would be a better option.

dennisguse:
About preferences: understood.
About space: why not move into another tab?
Then  you have all the space you need :)

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,84 @@
...
+
+import de.dennisguse.opentracks.content.data.TrackPoint;
+import de.dennisguse.opentracks.util.LocationUtils;
+import de.dennisguse.opentracks.util.UnitConversions;
+
+public class IntervalStatistics {
+    List<Interval> intervalList;
+
+    public IntervalStatistics() {
+        intervalList = new ArrayList<>();
```
dennisguse, line 14: Constructor can be removed; just initialize it in line 11.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,84 @@
...
+            intervalList.add(interval);
+        }
+    }
+
+    public List<Interval> getIntervalList() {
+        return intervalList;
+    }
+
+    public static class Interval {
+        float distance = 0f;
```
dennisguse, line 60: make private

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,84 @@
...
+        float time = 0f;
```
dennisguse, line 61: make private

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
```diff
@@ -0,0 +1,103 @@
...
+            IntervalStatistics intervalStatistics = new IntervalStatistics();
+            float distanceInterval = PreferencesUtils.isMetricUnits(context) ? (float) (interval.getValue() * UnitConversions.KM_TO_M) : (float) (interval.getValue() * UnitConversions.MI_TO_M);
+            intervalStatistics.build(trackPointList, distanceInterval);
+
+            intervalStats.postValue(intervalStatistics);
+        }).start();
+    }
+
+    public enum IntervalOption {
+        OPTION_1(1),
```
dennisguse, line 60: i guess this will be either 1mile / 1km, right?
So, we are independent of the actual unit

rgmf: Yes, it is.

rgmf:
> About preferences: understood.
> About space: why not move into another tab?
> Then you have all the space you need :)

Yes, good idea :ok_hand:

dennisguse:
Sorry I forgot to answer to all you questions..
We definitely need voice announcements for the last interval back (I killed this feature as I misunderstood it and didn't like the idea that it is stored in db). And about while recording vs. only afterwards. From my perspective, I would not take a look while recording as I have other things to do then. So, I would say afterwards only (and check how it feels).
And splitting up PRs is a good idea - we had some really big ones recently and it took (for my feeling) too long until they were merged.

rgmf:
> Sorry I forgot to answer to all you questions..
> We definitely need voice announcements for the last interval back (I killed this feature as I misunderstood it and didn't like the idea that it is stored in db). And about while recording vs. only afterwards. From my perspective, I would not take a look while recording as I have other things to do then. So, I would say afterwards only (and check how it feels).
> And splitting up PRs is a good idea - we had some really big ones recently and it took (for my feeling) too long until they were merged.

Too much late @dennisguse :sweat_smile: 

I'm just finishing intervals while user is recording. These intervals are shown in a new Tab and in the reverse order (the last one in the first place of the list).

The code used for recorded intervals is reused for recording intervals so it's very easy.

The announcement for last km/mi pace will come in another PR I'll open just when I finish this one.

rgmf:
It's ready to test and (I hope) perfect to use. The only thing is that I'd like to add tests to IntervalStatistics class before move this Draft into a PR.

dennisguse:
Ready for merging?

rgmf:
> Ready for merging?

No, I would like to improve the performance in recording intervals (in the new tab).

src/main/java/de/dennisguse/opentracks/fragments/IntervalsRecordingFragment.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private IntervalStatisticsModel viewModel;
+    private IntervalReverseListView intervalListView;
+
+    private class IntervalRunner implements Runnable {
+        private boolean stopped = false;
+
+        @Override
+        public void run() {
+            if (!stopped) {
```
dennisguse, line 42: Why not check isResume() of parent class?
Then we don't need stop.
See StatisticsRecordingFragment:107.

src/main/java/de/dennisguse/opentracks/fragments/IntervalsRecordingFragment.java
```diff
@@ -0,0 +1,136 @@
...
+        viewModel = null;
+    }
+
+    /**
+     * Update intervals through {@link IntervalStatisticsModel} view model.
+     *
+     * @param interval intervals will split in this interval if not null. If it's null then view model will use the default one.
+     */
+    private void updateIntervals(@Nullable IntervalStatisticsModel.IntervalOption interval) {
+        if (viewModel == null | intervalListView == null) {
```
dennisguse, line 109: ||

src/main/java/de/dennisguse/opentracks/fragments/StatisticsRecordedFragment.java
```diff
@@ -24,38 +24,48 @@
...
-public class StatisticsRecordedFragment extends Fragment implements TrackDataListener {
+public class StatisticsRecordedFragment extends Fragment implements IntervalListView.IntervalListListener {
+
+    private static final String TRACK_ID_KEY = "trackId";
 
     private TrackStatistics trackStatistics = null;
     private String category = "";
+    private Track track;
+
+    private IntervalStatisticsModel viewModel = null;
```
dennisguse, line 61: Why initialize some vars with `= null` directly?
Isn't implicit?

src/main/java/de/dennisguse/opentracks/fragments/StatisticsRecordedFragment.java
```diff
@@ -135,28 +174,28 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
-        TrackDataHub trackDataHub = ((TrackRecordedActivity) getActivity()).getTrackDataHub();
-        trackDataHub.unregisterTrackDataListener(this);
+        loadStatistics();
     }
 
     @Override
     public void onDestroyView() {
         super.onDestroyView();
 
+        PreferencesUtils.unregister(getContext(), sharedPreferenceChangeListener);
```
dennisguse, line 197: +1

src/main/java/de/dennisguse/opentracks/fragments/StatisticsRecordedFragment.java
```diff
@@ -291,5 +298,19 @@ private void updateUI() {
             speedMovingValue.setText(parts.first);
             speedMovingUnit.setText(parts.second);
         }
+
+        // Set intervals.
```
dennisguse, line 302: Commt is helpful?

src/main/java/de/dennisguse/opentracks/util/UnitConversions.java
```diff
@@ -51,6 +51,8 @@
     public static final double MM_TO_M = 0.001;
 
     // Distance
+    // multiplication factor to convert miles to meters
+    public static final double MI_TO_M = 1609.344;
```
dennisguse, line 55: Isn't this: 
Can you introduce 
`
public static final double MI_TO_KM = 1 / KM_TO_MI;
public static final double MI_TO_M = MI_TO_KM * KM_TO_M;
`

PS: Did not test this.....

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,80 @@
...
+
+import de.dennisguse.opentracks.content.data.TrackPoint;
+import de.dennisguse.opentracks.util.LocationUtils;
+import de.dennisguse.opentracks.util.UnitConversions;
+
+public class IntervalStatistics {
+    List<Interval> intervalList;
+
+    public void build(List<TrackPoint> trackPointList, float distanceInterval) {
+        intervalList = new ArrayList<>();
```
dennisguse, line 14: Default initialize in l:11 and then use intervalList.clear();

This may save some memory allocation (or the runtime is really good).

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,80 @@
...
+            intervalList.add(interval);
+        }
+    }
+
+    public List<Interval> getIntervalList() {
+        return intervalList;
+    }
+
+    public static class Interval {
+        private float distance = 0f;
```
dennisguse, line 56: distance_m?

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,80 @@
...
+        private float time = 0f;
```
dennisguse, line 57: time_ms?

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,80 @@
...
+
+                if (interval.distance >= distanceInterval) {
+                    float adjustFactor = distanceInterval / interval.distance;
+                    Interval adjustedInterval = new Interval(interval);
+                    adjustedInterval.distance *= adjustFactor;
+                    adjustedInterval.time *= adjustFactor;
+
+                    intervalList.add(adjustedInterval);
+
+                    Interval newInterval = new Interval();
```
dennisguse, line 37: What about:

```
interval = new Interval(interval.distance - adjustedInterval.distance, interval.time - adjustedInterval.time);
```

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,80 @@
...
+            TrackPoint trackPoint = trackPointList.get(i);
+
+            if (LocationUtils.isValidLocation(trackPoint.getLocation()) && LocationUtils.isValidLocation(prevTrackPoint.getLocation())) {
+                interval.distance += prevTrackPoint.distanceTo(trackPoint);
+                interval.time += trackPoint.getTime() - prevTrackPoint.getTime();
+
+                if (interval.distance >= distanceInterval) {
+                    float adjustFactor = distanceInterval / interval.distance;
+                    Interval adjustedInterval = new Interval(interval);
+                    adjustedInterval.distance *= adjustFactor;
```
dennisguse, line 32: For Interval:
public void adjust(float adjustFactor) {
  distance *= adjustFactor;
  time *= adjustFactor;
}

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public List<Interval> getIntervalList() {
+        return intervalList;
+    }
+
+    public static class Interval {
+        private float distance = 0f;
+        private float time = 0f;
+
+        public float getDistance() {
```
dennisguse, line 59: Put constructor(s) first

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -0,0 +1,80 @@
...
+
+        public Interval(Interval i) {
+            distance = i.distance;
+            time = i.time;
+        }
+
+        /**
+         * @return speed of the interval in m/s.
+         */
+        public float getSpeed() {
```
dennisguse, line 73: getSpeed_ms()

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * This view model is used to load intervals for a track.
+ * It uses a default interval but it can be set from outside to manage the interval length.
+ */
+public class IntervalStatisticsModel extends AndroidViewModel {
+
+    private MutableLiveData<IntervalStatistics> intervalStats;
+    private IntervalOption interval;
+
+    public IntervalStatisticsModel(@NonNull Application application) {
```
dennisguse, line 28: Why application?
I guess, we pass it nowwhere else, right?

Is this just a context?

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
```diff
@@ -0,0 +1,116 @@
...
+
+        IntervalOption(int value) {
+            this.value = value;
+        }
+
+        /**
+         * @param pos position of the interval option.
+         * @return    the interval option that is in the position pos.
+         */
+        public static IntervalOption getIntervalOption(int pos) {
```
dennisguse, line 90: What is the benefit of using a method?
I would also argue that we should fail immediately rather than returning a default option.

```
   IntervalOption.values();
````

File: src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java [hunk 20]
```diff
@@ -0,0 +1,116 @@
...
+            if (values().length > pos) {
+                return values()[pos];
+            } else {
+                return OPTION_1;
+            }
+        }
+
+        public int getValue() {
```
dennisguse, line 98: either static methods first or at the end.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
```diff
@@ -0,0 +1,116 @@
...
+            return value;
+        }
+
+        /**
+         * @return a string array with all options.
+         */
+        public static String[] getAllValues() {
```
dennisguse, line 105: This is just for to string conversion, right?
If yes, then I would move this functionality to the actual user of this function.

```
        int[] a = {1};
        Arrays.stream(a).mapToObj(String::valueOf).toArray(String[]::new);
```

src/main/java/de/dennisguse/opentracks/views/IntervalListView.java
```diff
@@ -0,0 +1,96 @@
...
+        if (PreferencesUtils.isKey(getContext(), R.string.stats_units_key, key) || PreferencesUtils.isKey(getContext(), R.string.stats_rate_key, key)) {
+            if (spinnerIntervalsUnit != null) {
+                spinnerIntervalsUnit.setText(PreferencesUtils.isMetricUnits(context) ? context.getString(R.string.unit_kilometer) : context.getString(R.string.unit_mile));
+            }
+        }
+    };
+
+    public IntervalListView(Context context, IntervalListListener listener) {
+        super(context);
+        this.context = context;
```
dennisguse, line 44: Can we not use the content of the parent?
(PS: don't know)

src/main/java/de/dennisguse/opentracks/views/IntervalListView.java
```diff
@@ -0,0 +1,96 @@
...
+        this.listener = listener;
+        PreferencesUtils.register(getContext(), sharedPreferenceChangeListener);
+        init();
+    }
+
+    protected void init() {
```
dennisguse, line 50: Why not put this into the constructor?

src/main/java/de/dennisguse/opentracks/views/IntervalReverseListView.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.List;
+
+import de.dennisguse.opentracks.adapters.IntervalStatisticsAdapter;
+import de.dennisguse.opentracks.viewmodels.IntervalStatistics;
+
+/**
+ * LinearLayout view used to build a list of intervals in a reverse mode, the last one will appear in the first position on the LinearLayout.
+ * This class is an specialization of {@link IntervalListView} that display the views contained in the LinearLayout in a reverse mode.
+ */
+public class IntervalReverseListView extends IntervalListView {
```
dennisguse, line 15: Can move this into IntervalListView as a public static class?

src/main/java/de/dennisguse/opentracks/views/IntervalReverseListView.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public IntervalReverseListView(Context context, IntervalListListener listener) {
+        super(context, listener);
+    }
+
+    public void display(List<IntervalStatistics.Interval> intervalList) {
+        if (intervalList != null) {
```
dennisguse, line 22: if (intervalList == null) {
  return;
}

dennisguse (changes requested):
Smaller stuff as individual comments.
* For a recorded track, the fragment is shown not as a tab.
* the split interval selector has an orange "thing" below, but it seems to be add an odd position
* why not center the table (left/right margin the same)

And (I did not see it in the code): does the update timer also run for a already recorded track?

dennisguse:
And we can remove the "Intervals" title from the fragment as it is shown by the tablayout.

src/androidTest/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsTest.java
```diff
@@ -0,0 +1,144 @@
...
+		return trackStatisticsUpdater.getTrackStatistics();
+	}
+
+	/**
+	 * Tests that build method compute the distance correctly comparing the result with TrackStatisticsUpdater result.
+	 */
+	@Test
+    public void testBuild_1() {
+		// With 50 points and interval distance of 1000m.
+
```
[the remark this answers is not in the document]
rgmf: Strange... I see it well formatted. Anyway I've reformatted.

src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
```diff
@@ -382,8 +383,10 @@ public Fragment getItem(int position) {
                 case 0:
                     return new StatisticsRecordingFragment();
                 case 1:
-                    return ChartFragment.newInstance(false);
+                    return new IntervalsRecordingFragment();
```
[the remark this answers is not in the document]
rgmf: I thought that this was the best option because it not receive any arguments. Changed ;)

src/main/java/de/dennisguse/opentracks/adapters/IntervalStatisticsAdapter.java
```diff
@@ -0,0 +1,65 @@
...
+import de.dennisguse.opentracks.viewmodels.IntervalStatistics;
+
+public class IntervalStatisticsAdapter extends ArrayAdapter<IntervalStatistics.Interval> {
+
+    private boolean metricUnits;
+    private float accDistance = 0f;
+
+    public IntervalStatisticsAdapter(Context context, List<IntervalStatistics.Interval> intervalList) {
+        super(context, R.layout.interval_stats_list_item, intervalList);
+        metricUnits = PreferencesUtils.isMetricUnits(context);
```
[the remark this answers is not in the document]
rgmf: Yes, every time view's `display` method is called it recreates the adapter. The fragments that use this view are responsible of re-`display`ing the view and they're who listen preferences changes to update UI (including the view).

src/main/java/de/dennisguse/opentracks/adapters/IntervalStatisticsAdapter.java
```diff
@@ -0,0 +1,65 @@
...
+
+import de.dennisguse.opentracks.R;
+import de.dennisguse.opentracks.util.PreferencesUtils;
+import de.dennisguse.opentracks.util.StringUtils;
+import de.dennisguse.opentracks.viewmodels.IntervalStatistics;
+
+public class IntervalStatisticsAdapter extends ArrayAdapter<IntervalStatistics.Interval> {
+
+    private boolean metricUnits;
+    private float accDistance = 0f;
```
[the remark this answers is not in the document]
rgmf: Ooohhhh yes, I've seen a lot of times the unit added to the name of variables... sorry :+1:

src/main/java/de/dennisguse/opentracks/adapters/IntervalStatisticsAdapter.java
```diff
@@ -0,0 +1,65 @@
...
+            intervalView.setTag(viewHolder);
+        } else {
+            viewHolder = (ViewHolder) intervalView.getTag();
+        }
+
+        accDistance += interval.getDistance();
+        viewHolder.distance.setText(StringUtils.formatDistance(getContext(), accDistance, metricUnits));
+        Pair<String, String> speedParts = StringUtils.getSpeedParts(getContext(), interval.getSpeed(), metricUnits, true);
+        viewHolder.speed.setText(speedParts.first + " " + speedParts.second);
+        viewHolder.pace.setText(StringUtils.getSpeedParts(getContext(), interval.getSpeed(), metricUnits, false).first);
```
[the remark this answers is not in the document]
rgmf: Because pace is time. When you see stats in a table, the column with pace doesn't show unit. I've never seen the unit but It's only my experience and I don't know what standard says about this (if any standard about it).

src/main/java/de/dennisguse/opentracks/fragments/IntervalsRecordingFragment.java
```diff
@@ -0,0 +1,136 @@
...
+import de.dennisguse.opentracks.views.IntervalReverseListView;
+
+/**
+ * A fragment to display the intervals from recording track.
+ */
+public class IntervalsRecordingFragment extends Fragment implements IntervalListView.IntervalListListener {
+
+    private static final String TAG = IntervalsRecordingFragment.class.getSimpleName();
+
+    private static final long UI_UPDATE_INTERVAL = 5 * UnitConversions.ONE_SECOND_MS;
```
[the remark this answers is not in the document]
rgmf: Of course ;)

src/main/java/de/dennisguse/opentracks/adapters/IntervalStatisticsAdapter.java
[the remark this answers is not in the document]
dennisguse: One the statistics fragment pace is shown with unit.
i don't know if it is meaningful to the user.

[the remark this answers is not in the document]
rgmf: You are right... So I have to add here the unit too for maintain the coherence.

Anyway, I'm going to look into about this matter... For the moment I'll add the unit

;)

src/main/java/de/dennisguse/opentracks/fragments/IntervalsRecordingFragment.java
rgmf: Oopps :(

src/main/java/de/dennisguse/opentracks/fragments/StatisticsRecordedFragment.java
rgmf: Lack of Java experience... I won't forget anymore :smile:

src/main/java/de/dennisguse/opentracks/util/UnitConversions.java
rgmf: > Can you introduce
> public static final double MI_TO_KM = 1 / KM_TO_MI; public static final double MI_TO_M = MI_TO_KM * KM_TO_M;

:+1: 

> PS: Did not test this.....

I check it out and I'll add tests if not any.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
rgmf: `AndroidViewModel` needs `Application` (when I call `super(application)`).

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
rgmf: The benefit of using the method was not to fail immediately. I've removed it then ;)

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
rgmf: Is this okay?

Review this please:

```java
int[] intValues = Arrays.stream(IntervalStatisticsModel.IntervalOption.values()).mapToInt(i -> i.getValue()).toArray();
spinnerIntervals.setAdapter(new ArrayAdapter<>(getContext(), android.R.layout.simple_spinner_dropdown_item, Arrays.stream(intValues).mapToObj(String::valueOf).toArray(String[]::new)));
```

src/main/java/de/dennisguse/opentracks/views/IntervalListView.java
rgmf: Yes, we can use `getContext()`. Done.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
dennisguse: Why not?
```
String[] values = Arrays.stream(IntervalStatisticsModel.IntervalOption.values()).mapToObj(String::valueOf).toArray(String[]::new))
spinnerIntervals.setAdapter(new ArrayAdapter<>(getContext(), android.R.layout.simple_spinner_dropdown_item, values);
```

rgmf: I didn't sure what you want was it.
Okay, then :+1:

dennisguse: PS: Not tested ;)

rgmf:
> Smaller stuff as individual comments.
> 
>     * For a recorded track, the fragment is shown not as a tab.

For recorded I most like is in the same STATS fragment. In fact it's part of the stats of the track recorded. Less code, less clasess...

For recording the more simple is the stats you see the better it is. So the idea of moving intervals to other fragment (tab) was great.

Anyway, for me is really acceptable to move intervals to another tab/fragment. I you think so let me know and I'll move intervals to another fragment/tab ;)

>     * the split interval selector has an orange "thing" below, but it seems to be add an odd position

It's the spinner style. Look at this:

>     * why not center the table (left/right margin the same)
> 
> 
> And (I did not see it in the code): does the update timer also run for a already recorded track?

rgmf:
> Smaller stuff as individual comments.
> 
>     * For a recorded track, the fragment is shown not as a tab.

For recorded I most like is in the same STATS fragment. In fact it's part of the stats of the track recorded. Less code, less clasess...

For recording the more simple is the stats you see the better it is. So the idea of moving intervals to other fragment (tab) was great.

Anyway, for me is really acceptable to move intervals to another tab/fragment. I you think so let me know and I'll move intervals to another fragment/tab ;)

>     * the split interval selector has an orange "thing" below, but it seems to be add an odd position

It's the spinner style. Look at this:

spinner

Here you can see limits of spinner. The spinner has a big padding because it's make easy to the user to touch on spinner easily.


>     * why not center the table (left/right margin the same)

An imagen will help me to explain why :D

imagen

Also, adding the unit to pace the space is filler.

> And (I did not see it in the code): does the update timer also run for a already recorded track?

Recorded track doesn't need the timer because it already finished. The timer is for refreshing intervals with the news TrackPoints. For recorded you already have all TrackPoints.

rgmf:
OpenTracks announces avg. speed or pace and now announces current speed or pace (from current kilometer or mile) too.

This last commit fixes #134 

The announcement is now like this, for example:

"OpenTracks total distance 20.00 kilometers in 5 minutes 0 seconds at 240.0 kilometers per hour with a current kilometer speed of 30 kilometers per hour"

I don't know if the message it's okay or it's too long because it's not a feature I use.

pstorch:
I don't use announcements

dennisguse:
@rgmf i also don't use this feature that often.
Let's leave the message for now as it is, because it is completely translated.
But we should probably tweak it afterwards.

dennisguse:
PS: My main problem with voice announcements is that LineageOS stopped shipping a TTS system...
Sucks.

dennisguse:
@rgmf Just one more thing:
Why is the implementation different for recording vs. recorded?
For recording the intervals are in a separate tab and for recorded tracks the interval is below the statictics.
Is this intentionally and why?

About centering: I meant not the content of the table/layout (that looks fine like it is), but rather the margin to the left and right should the same (left and right to the end of the screen).

rgmf:
> Why is the implementation different for recording vs. recorded?
> For recording the intervals are in a separate tab and for recorded tracks the interval is below the statictics.
> Is this intentionally and why?

In **recorded** screen you have all stats in the same tab. It's the option I like the best: looking into the stats without the needed of tab changing.

In **recording** the situation is different: while you are recording you don't want to have to scroll to see the live stats. I want all stats in sight so here the approach is organize all information in tabs. If you want to see main stats (speed, distance, time...) then you move to stats tab. If you want to see intervals then you move to interval stats. If you want to see map then you move to map. If you want to see graphs then you move to graph tab. So you have several tab with different kind of data.

> About centering: I meant not the content of the table/layout (that looks fine like it is), but rather the margin to the left and right should the same (left and right to the end of the screen).

Ahhh, okay, I'll look into this.

dennisguse:
I just moved the Interval for recording into tab (it felt too weird to have different tabs count)
Additional change: trackid is always passed from parent (getting it from preference will be removed anyhow)

And I am slightly confused:
* why do you get the trackPoints via contentProvider instead of TrackDataHub (here it is already filtered; especially important for long tracks)? (and processing should use a cursor rather than the method that is meant for testing ><)
* why did you decide to extend a LinearLayout?
For me it feels like as if a ListView would be more appropriate (similar to TrackListActivity)
* why do you need interval_recordings? It just gives a ScrollView.
And then on root_view the intervalListView is added manually (that part confuses me)
Why not instead:
```
    <de.dennisguse.opentracks.views.IntervalListView
        android:id="@+id/interval_list"
        android:layout_width="match_parent"
        android:layout_height="wrap_content"
        android:orientation="vertical"/>
```

Anyhow - the PR is too big already.
My suggestion: merge it and improve afterwards.
@rgmf What do you think?

dennisguse:
Actually it feels even more like a tableLayout.

rgmf:
@dennisguse I like the change to a tab in recorded.

> why do you get the trackPoints via contentProvider instead of TrackDataHub (here it is already filtered; especially important for long tracks)? (and processing should use a cursor rather than the method that is meant for testing ><)

Ooopss, I see the "VisibleForTesting" annotation. Okay, let me to the move from content provider to TrackDataHub.

> why did you decide to extend a LinearLayout?

I don't remember why I decided to use a LinearLayout  because it was a long time ago ><. I think it was easy and simple. I can move it to a ListView.

>  My suggestion: merge it and improve afterwards.
@rgmf What do you think?

I'm in no hurry, I'd wait to have everything as you say.

dennisguse:
@rgmf "as you say" sounds like being at work ;)
Don't worry - the functionality works, so we can merge it, release it, and improve afterwards.
The good thing is: we are not building rockets here - we can shot as often as we want ><

PS: My Android UI skills are really limited (haven't done it since university) - so not everything that feels weird for me is not necessarily bad.
It is just my gut feeling...

dennisguse:
@rgmf Do you know why this test fails on CI?
```
 de.dennisguse.opentracks.services.handlers.HandlerServerTest > sendTrackPoint[test(AVD) - 10] FAILED 
	Wanted but not invoked:
	server.newTrackPoint(
```
Was there a change there?

rgmf:
> Was there a change there?

I launch tests on AVD with API 21 and on AVD with API 28 a I don't 

> @rgmf Do you know why this test fails on CI?

No idea.

> 
> ```
>  de.dennisguse.opentracks.services.handlers.HandlerServerTest > sendTrackPoint[test(AVD) - 10] FAILED 
> 	Wanted but not invoked:
> 	server.newTrackPoint(
> ```
> 
> Was there a change there?

As far as I know there wasn't any changes there.

rgmf:
> @rgmf "as you say" sounds like being at work ;)

Thi is much better that being at work :rofl: :rofl: 

In my latter work the important thing was releasing not doing the better :roll_eyes: 

> Don't worry - the functionality works, so we can merge it, release it, and improve afterwards.
> The good thing is: we are not building rockets here - we can shot as often as we want ><

I've been testing it for a log time ago and it works really well on my Pixel 2 XL... but reading all your reviews I understand the code can be better and I now see things that before I don't.

Anyway, let's get shot (merge) :smile:

dennisguse:
@rgmf I just added a proper commit message to my last commit.
Feel free to push the button ><

rgmf:
@dennisguse I've changed the order of intervals for Recorded stats: natural order is the best here. For Recording the focus is on the last interval (km or mi) but not in Recorded one ;)

dennisguse:
I fucked up yesterday (was too late probably): recording was failing due to an NPE (trackId for a new track was not yet known).
Fixed now - sometimes it just one of those days...
And I(!) need UI tests... for real.

slackline:
Hi,

I've now had a chance to give this a whirl and really like the fact that I get the time taken on the last kilometer/mile.

One thing though the announcement for this is..

"With a current pace of X Minutes  and y seconds" (or words to that effect, close to above).

I was wondering if to avoid confusion/ambiguity it would be simpler to refer to it as the "Lap time of X".  This is more logical because a pace is an average over a period of time, which is the traditional announcement and what precedes the new addition.  Whilst it could be argued that even the time for the current km/mile is also an average as speed will have fluctuated over the distance, it is more traditional to refer to this as a "Lap time" rather than a pace.

Thanks for implementing/working on this @rgmf and @dennisguse.

rgmf:
@slackline thank you for your feedback. I'll change the announcement to "Lap time of..."

slackline:
@rgmf thats great, thanks.

PR #369, opened 2020-08-12
Multi format instant export

thrillfall:
**Describe the pull request**
Instant post-workout track export to external storage

**Link to the the issue**
#186 #90


thrillfall:
@dennisguse i read about persisting permission but i am more or less lost as the where to put those two lines?
```
final int takeFlags = intent.getFlags()
            & (Intent.FLAG_GRANT_READ_URI_PERMISSION
            | Intent.FLAG_GRANT_WRITE_URI_PERMISSION);
// Check for the freshest data.
getContentResolver().takePersistableUriPermission(uri, takeFlags);
```
i would assume in SingleTrackExportDirectoryChooserActivity.java?

dennisguse:
@thrillfall I guess, you are right!
Never used this feature, but just saw it in the doc :)

And I won't have time before the weekend to give you any more feedback - I want to finish some other stuff first...

thrillfall:
The post-workout export now uses the configured format.

The ony thing that is still off is that i get several exportet files during the workout.
I get the first exported file with filename [startdate_and_time]**.GPX** with around 2kb of data and then i get 4-7 subsequent files over the course of the recording with **.GPX (n)** where n is incremented every time and each file contains more data. The last file contains the whole track.
Can anyone give me a hint why that is?

dennisguse:
Because you export everytime, the TrackRecordingActivity is stopped by Android.
And that might happen actually quite often - take a look at the lifecycle: https://developer.android.com/guide/components/activities/activity-lifecycle

Actually, the TrackRecordingService should trigger the export as it is the component responsible for the recording (and also the only one that knows when the recording is done).

thrillfall:
> Because you export everytime, the TrackRecordingActivity is stopped by Android.
> And that might happen actually quite often - take a look at the lifecycle: https://developer.android.com/guide/components/activities/activity-lifecycle
> 
> Actually, the TrackRecordingService should trigger the export as it is the component responsible for the recording (and also the only one that knows when the recording is done).

Thanks for the explanation. I moved the code accordingly. 

From my side that would be all i need for this feature. Does anyone have any further corrections/improvements to make?

dennisguse:
For now blocked until #312 is merged.

@thrillfall i need to get #312 merged first (took already too long).
In the mean time could you do two things with this PR (too make the review easier):
* Can you check if you can rebase it onto #312?
* Can you squash and reorder the commits (so that they are meaningful together)?

PS: I did not read your PR, so no feedback so far.
Probably on the weekend.

thrillfall:
i reordered the commits. Refactorings come first followed by the export functionality in one commit. I'll rebase onto your proposed branch and then create a new mr

dennisguse:
You can 'git push -f" to the original branch, github updates the PR automatically.

thrillfall:
i changed the target branch and rebased to that.

Issue #185, opened 2020-04-14
Preserving the heart rate history of the entire workout and pulse zones


MrS1R1US:
** Is your feature request related to a problem? Please describe. **
Heart rate analysis by time or distance of training is lacking as it is implemented in Polar Beat or Runtastic applications.

** Describe the solution you'd like **
Add custom heart rate zones and a historical heart rate graph by time or distance training. For an athlete this is very important.

** Describe alternatives you've considered **
Add custom heart rate zones and a historical heart rate graph by time or distance training. For an athlete this is very important.

** Additional context **
Screenshots:
[https://i.hizliresim.com/ZXnmp3.jpg](url)
[https://www.wareable.com/media/imager/201606/14064-original.jpg](url)

MrS1R1US:
What can you say about this? Does it make sense to wait for such an introduction? Very critical is not enough.

dennisguse:
@MrS1R1US It definitively makes sense to wait :)
If you have some time, you could create a concept how to implement this (e.g., screen shots / scribbles).

One (technical) issue might be that only _one_ heart rate measurement is stored per GPS location.
So, if a GPS is recorded every 10 seconds the only _last_ measurement in this period is stored.
For longer intervals this might even be more problematic...

MrS1R1US:
@dennisguse 
If you really do this, I will certainly draw a layout in Photoshop about how it will look. And I will be very grateful to you for such functionality!) An interval of 10 seconds for large distances is normal. For small workouts, 5-10 minutes is not very. In any case, you need to try, because this will attract the attention of athletes who train with sensors like me. For the sake of OpenTracks, I demolished the official firmware and installed Lineage OS, after which I got GPS.

DevRonin:
What phone model do you have?
I also have LineageOS 15 (Android 8.1.0) on LeEco LE2 snapdragon version.
I ordered a Polar H10 and Garmin HRM dual.

As a starter, a new tab, with a  simple graph like the one with elevation vs distance will be good.
Y - heart rate
X - this could be time, distance, speed, etc.

Beside current HR, some stats for HR training will be good, like total time spent on certain interval, max. continuous time in that interval, etc.
Also possibility to set a target HR interval & low(under)/high(over) limit alarm will be nice.

DevRonin:
> @MrS1R1US It definitively makes sense to wait :)
> If you have some time, you could create a concept how to implement this (e.g., screen shots / scribbles).
> 
[...]

PR #377, opened 2020-08-15
add aggregated statistics diagram

thrillfall:
**Describe the pull request**
Instant post-workout track export to external storage

**Link to the the issue**
#186 #90


dennisguse:
Nice work! 
Just one (not so tiny thing) - could you reuse the ChartView that we already have?
So far, we are (almost) dependency free.

thrillfall:
i tried using ChartView but that is just not what i can handle. 
Correct me if i am wrong but ChartView works with a List of ChartPoint. These ChartPoints are a representation of tracked locations which is not something i can provide.
My chart works with a set of 2d vectors computated from several tracks. Trying to force this computated data into a structure that does something totally different (render a set of locations) seems totally wrong and a lot of effort.

TL;DR;  i don't think its a good idea to use ChartView for the graph i want to provide.

rgmf:
Great idea!!! And huge job!!! :D

> Currently this is an additional menu entry in settings. I could imagine converting the existing aggregated stats activity and this to a tabbed view. what do you think?

IMHO the best option would be added "fastest average speed" and "greatest distance" in the `ListView` that is in `AggregatedStatisticsActivity` just below of "Avg. moving pace" and "Fastest pace" for every type of activity.

Also, I'd add the `onClick` listener to each item of that `ListView` so user clicks on "TRAIL RUNNING" section then `ShowWorkoutsAggregatedDiagramActivity` is opened with diagrams. In this way user doesn't need select type of activity (in the spinner) because s-he selects it clicking on the item s-he want to see.

What do you think @thrillfall?

Finally, I've checked `ShowWorkoutsAggregatedDiagramActivity` out with my actual tracks and I see this:

Captura de pantalla_2020-08-16_16-28-13

That it's very difficult to analize.

I think, this diagrams should showed by week because we have a space limit for show all activities.

The long story short (IMHO):
- Diagrams by week: focused on activities.
- Diagrams by month: focused on aggregated stats by week (4 or 5 weeks).
- Diagrams by year: focused on aggregated stats by month (12 months).

I know it's a huge job but I think the result will be great.

I comment all this because I was thinking about all that for several weeks ago, but it's only an opinion and it can be focused from a lot of perspectives.

What do you think? @thrillfall @dennisguse

dennisguse:
About the general data presentation: I don't really have an opinion about that as I rarely need it.
But for our users: it would be awesome!
And we even have some requests to improve our data representation (like #185).

And I have another proposal (for discussion): #379.

thrillfall:
> That it's very difficult to analize.
@rgmf You are right. But this weird curving speed graph is a bug. Somehow the values for the date do not align for speed and distance point on the graph.

Thank you also foir all the nice ideas.
I'd also favor moving the maximum distance and speed (with date) to the existing view.

For now users would (given the bug is fixed) have the option to zoom in/out and go back and forth in the graph to see the results for the time range in view. 
As for your idea to have different aggregations for different time ranges i'd see that as a topic for another increment and not as a show stopper for this MR. Lets give the other users the graph as is (without that bug) and extend it from release to release.

thrillfall:
> i tried using ChartView but that is just not what i can handle.
> Correct me if i am wrong but ChartView works with a List of ChartPoint. These ChartPoints are a representation of tracked locations which is not something i can provide.
> My chart works with a set of 2d vectors computated from several tracks. Trying to force this computated data into a structure that does something totally different (render a set of locations) seems totally wrong and a lot of effort.
> 
> TL;DR; i don't think its a good idea to use ChartView for the graph i want to provide.

so can i assume from your response that i'd be an option to use the given approach for the graph? I would neither have the time nor the will to force this into ChartView.

dennisguse:
I would love this - only thing is the technical implementation.
And if we find a way to kick/replace the current charts: I am all in.
My general concern is maintainability (@thrillfall that's my issue with external libs).

@thrillfall just go ahead - we will figure something out.
And by the way: thanks for your time that you already spent!

thrillfall:
@rgmf the weird curve was not so much a bug but a special mode i choose for displaying the speed graph. I changed that back to the default linear mode which should make the graph more readable
Screenshot_20200816-230127_OpenTracks_(Debug)

rgmf:
Great @thrillfall ;)

I'll test it tomorrow to see the results with all my tracks but seem more much better :D

PR #385, opened 2020-08-19
Id#276

### LINKED ISSUE #276: Refactor trackId to a `class` instead of a `long`
We might need to change trackId from `long` to something more useful like `UUID` to enable synchronization between multiple installtions of OpenTracks.

A first step would be to change trackId (which is used everywhere) to a class containing one long.

dennisguse:
Migrate trackid and waypointId to class.
Track.Id and Waypoint.Id instead of "long".

These changes prepare the change to not use sharedPreferences to communicate the recordingState within OpenTracks, but rather use the TrackRecordingService.

Tests are working and I tested manually.

Fixes #276.


dennisguse:
@rgmf @pstorch Can you test the changes manually?
I hope I did not overlook something.

PS: Track.Id.getValid() will be removed in the future.

src/main/java/de/dennisguse/opentracks/content/data/Track.java
```diff
@@ -144,6 +149,12 @@ public int describeContents() {
             return 0;
         }
 
+        @NonNull
+        @Override
+        public String toString() {
+            throw new RuntimeException("Not supported");
```
pstorch, line 156: Isn't that dangerous? Will the debugger crash, when you look at an Id variable?

rgmf:
@dennisguse when I try to delete a track then OpenTracks cracks with this error:

`java.lang.ClassCastException: android.os.Parcelable[] cannot be cast to de.dennisguse.opentracks.content.data.Track$Id[]`

in `TrackDeleteActivity`:

`trackIds = (Track.Id[]) intent.getParcelableArrayExtra(EXTRA_TRACK_IDS);`

src/main/java/de/dennisguse/opentracks/content/data/Track.java
dennisguse: The debugger survives (at leas tin AndroidStudio) and will just to a toString on the exception.
We can remove this afterwards; I just wanted to make sure that we never use toString in old context (aka `"Something" + trackId`) and get unexpected/weird results.

rgmf:
Feeling and issues after an actual test:
- Activity type cannot be changed while you are recording.
- Total time is not flowing like before.
- Speed doesn't move, it keeps in 0km/h all the time. But average speed works.

dennisguse:
Fixes implemented.

@rgmf Thanks for these findings ;)
I did not see it myself :)

PR #391, opened 2020-08-24
Settings: animate the Bluetooth icon.

dennisguse:
... even more animations ;)


dennisguse:
And I forgot: the dialog icon in any of the Bluetooth sensor selection dialogs is animated.

dennisguse:
@pstorch Due to #393.
Do we need to change something here?

src/main/res/drawable/ic_bluetooth_searching_animated_24dp.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<animated-vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:drawable="@drawable/ic_bluetooth_searching_24dp">
```
pstorch, line 3: @dennisguse this was the problematic line in the stop-button animation. But I just tested this in an API 21 Emulator and the bluetooth icon was animated and didn't crash. Hmmm.

Issue #390, opened 2020-08-24
force close when stopping track


nutpantz:
**Describe the bug**
I start a track let it run for a few minutes and when i push to stop (or pause)the track it force closes and automatically starts recording the same track again on restart... And is suck in the force close auto record loop
Tried force close and clear data and re install.  Same issue..
Works fine on my s7 with Google services
**To Reproduce**

If applicable:
* add screenshots
* add logcat output

**Technical information**
 - Device: Galaxy s4
 - OS:  LineageOS  androud 7.12
 - Version  v3.9.1

From fdroid no Google play or services installed
Had no issues on the last version. Been using this app for months very happy until latest update
Also crashes on my levono phab running Android 6 with Google services disabled
 also crashes on Nexus 7 lineage Android 7.1.2 no Google

rgmf:
@nutpantz do you have the same issue when you pause the track? Your issue remember me an weird issue I had when new buttons feature were added. In my case the resource `ic_button_stop_anim.xml` was the problem, Android didn't found it but this was in its place. I fixed the problem uninstalling OpenTracks and installing it again.

But seem you cannot fix the problema with reinstallation :(

nutpantz:
Yes it appears it has the same issue when pausing the track..
(Which i have never done before)

pstorch:
That's strange. Are the Google services needed for the animation?
I don't have a device without, so unfortunately I can't reproduce.

nutpantz:
I can confirm after testing with my Nexus 7 that 3.8.1 was the last working version for this bug

pstorch:
3.9.0 introduced the long press + animation on the stop and pause button to prevent accidental stop of the recording.
Have you tried reinstalling the App as @rgmf mentioned? Be careful and backup your tracks first.

rgmf:
> That's strange. Are the Google services needed for the animation?
> I don't have a device without, so unfortunately I can't reproduce.

It works for me in phones with Google Services and without them.

nutpantz:
I have tried fresh install from fdroid to my Nexus 7 and reistalll to my s4 and Lenovo.. still same issue

dennisguse:
@rgmf Works for me (LineageOS 17.1; one S4/GT9505)

pstorch:
@nutpantz does it vibrate, when you push the stop/pause button?

Any chance to get access to the logcat?

nutpantz:
[...]

PR #393, opened 2020-08-25
Fix force close on stop button for older devices

pstorch:
Discovered in #390 for devices with API Level 25. Not fixing the initial issue, but should be corrected as well.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/TrackController.java
+++ b/src/main/java/de/dennisguse/opentracks/TrackController.java
@@ -148,18 +148,22 @@ public class TrackController {
 
         @Override
         public void run() {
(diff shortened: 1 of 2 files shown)


dennisguse:
Should we trigger a release for this?

pstorch:
@dennisguse I don't know how stable it is for a release. But for me it's a go.

rgmf:
> Should we trigger a release for this?

There are two important fixes:
- Animation issue.
- An this one: https://github.com/OpenTracksApp/OpenTracks/commit/60d5e7e51510a2bf177375eb19c7ce24635e2569

So it's important a new release.

Issue #406, opened 2020-09-03
Intervals: issue in recorded and in recording.


rgmf:
**Describe the bug**
There are troubles with intervals. In recorded you only can see intervals if you change the interval (spinner). In recording you cannot see intervals.

**To Reproduce in recorded**
1. Go to any track you recorded.
2. Move to intervals tab.
3. You won't see intervals.
4. Change spinner interval value and then you'll see intervals.

**To Reproduce in recording**
1. Record a new track.
2. Move to intervals tab.
3. Intervals don't show.

dennisguse:
That is my fault. It is the initialization of the interval.

PR #407, opened 2020-09-03
Fixed some intervals issues.

rgmf:
**Describe the pull request**
Recorded activity didn't initialize the intervals. Also, invalidate didn't work in view model.

Recording activity didn't show intervals because the activity passed a null Track.Id because the Track.Id is created in the Runnable for new tracks. Now interval fragment listen for a new Track.Id. I've kept the argument in newInstance because we don't know if Runnable finished before fragment is created, right? *I don't know if this is the best approach so I create a Draft PR.*

Finally, unit changed must result in a interval refreshing.

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/406

**License agreement**


rgmf:
@dennisguse if you need to modify something quickly do it although commit's author change. I don't know if in recording intervals fragment I did the best.

PR #408, opened 2020-09-03
Refactoring.

### LINKED ISSUE #406: Intervals: issue in recorded and in recording.

dennisguse:
Fixes #406 and simplifies code.

@rgmf I have not read #407 (and should not as it is too late for reading new code ><), but it sounds like the proper fix to the trackId thing.
I will take a look tomorrow.


rgmf (approved):
I like the simplifications. Recorded intervals initialization is fixed.

You can merge it onto main and I merge your changes with #407 where I fixed the unit settings change and recording intervals if you consider it appropiate.

PR #414, opened 2020-09-05
Present speed only with one fractional digit.

dennisguse:
10.1km/h instead of 10.19km/h.

Having two decimal digits suggests a precision that GPS will anyhow not provide.
I guess cycling speed sensors won't as well.

Collected data is still stored with "full" precision.


rgmf:
Aside the GPS precision I prefer one decimal for speed so I love this change :D

PR #417, opened 2020-09-06
Day / night mode

### LINKED ISSUE #231: Theming support day/night
**Describe the bug**
Dark theme is only partially supported

**To Reproduce**
1. Switch to "night mode" on android system (9.0)
2. Open OpenTracks
3. Go to "By time" or "by distance"
4. Graphs use bright background

If applicable:
Screenshot_20200516-155740_OpenTracks
Screenshot_20200516-155743_OpenTracks

**Technical information**
 - Device: Samsung Galaxy S10
 - OS: default stock android
 - Version 9.0

dennisguse:
Slowly implementing day/night mode.

This will include:
* [x] migration to app compat themes
* [x] app-wide configuration settings (default: night)
* [x] tinting all icons properly
* [x] cleaning up themes (this will also fix that the color of the search icon is different)
* [x] move to material design components: https://material.io/develop/android/components/text-fields
* [x] make activity spinner look nice
* [x] charts
* [ ] ~~set day background for trackController~~ (later)

This PR changes the UI just by theming for day/night mode.
From my perspective some things do not yet look as polished as they were others are better.
What I don't like yet is the spinner styling (e.g., activity selector), but it saves us a lot of headaches not styling it :)

Fixes #231.

PS: I just found out that I toggle day/night mode easily on my LineageOS phone ;)


dennisguse:
@thrillfall What do you think?

rgmf (commented):
This works for me. I like it.

pstorch:
Only a few minor remarks:
The font is bigger now in the track list activity, so 1 track item less fits on the screen. Too big I opinion.
And the start/stop buttons have black background in day mode (I see you want to do this later).
The exclamation mark icon when deleting a track is too light in day mode, I think.

dennisguse:
Thanks for testing and the feedback.
Almost done.

About the start/stop background: also yes.
The icons have a transparent background and this looks awkward on a bright background.
And my color selection skills are basically zero, but I will try something less bright.
Suggestions welcome.

PS: The delete dialog background also does not fit to the dark style - it is too dark compared to the preference dialogs.
Don't yet why.

dennisguse:
The AlertDialog needs to be Androidx - that fixes the dark background color styling.

dennisguse:
@kylecorry31 This PR implements day/night mode for OpenTracks.
Technically Trail-Sense and OpenTracks will look very much alike as both apps use almost the default color codes :)
Major difference is that our orange is slightly different (FF8800 vs FF6D00), but Trail-Sense's orange (from my perspective) fits better to the day theme.
Long story short: do you have any trouble, if we use the same for day/night in OpenTracks?

kylecorry31:
No problem at all!

dennisguse:
@kylecorry31 I tested this some days ago and then the colorAccent does not fit anymore with the logo. So, I guess we need something in the middle.

PR #419, opened 2020-09-08
unify export filename logic

thrillfall:
unify creating export filename


dennisguse:
Can you the function to get the Track.Id instead of the long. The goal of the (buggy) refactoring in the prior versions was to be precise about when the id is used. So, all calls to getId() should happen as late as possible. Then it is easy to see later on were it is actually used. PS just force push ;)

PR #424, opened 2020-09-09
Theming: add a SplashScreen for TrackListActivity.

dennisguse:


src/main/AndroidManifest.xml
```diff
@@ -138,7 +137,9 @@ limitations under the License.
 
         <activity android:name=".TrackEditActivity" />
 
-        <activity android:name=".TrackListActivity">
+        <activity
+            android:name=".TrackListActivity"
+            android:theme="@style/SplashTheme">
```
rgmf, line 142: I get the following error when I try to build:

AAPT: error: resource style/SplashTheme (aka de.dennisguse.opentracks.debug:style/SplashTheme) not found.

rgmf (commented):
I get errors when build on an AVD API 28.

rgmf:
@dennisguse I think is great, I love it. Also this will avoid the message "Record your next adventure" for very slow devices that takes a while to show the list. Awesome :D

rgmf (approved):

dennisguse:
I am not sure about the new adventure message, as I guess it is determined by the resource adapter.
So, we basically show it until the database provided data to the activity.

dennisguse:
@rgmf I disabled the flacky test as this is really annoying.
I often get build failed from CI, which are false/positive and I thus tend to ignore the positive/positive ones.
Not a good practice - so if CI fails now: something is not compiling/working.
Thx for testing.

PR #431, opened 2020-09-13
Interval list refactoring

rgmf:
Following this: https://github.com/OpenTracksApp/OpenTracks/pull/368#issuecomment-685105520

I've moved LinearLayout to ListView for intervals.

Also I've refactored intervals fragment to delete duplicated code. Previous IntervalsRecordingFragment was a version of IntervalsRecordedFragment. So now we have only one fragment (IntervalsFragment) with a version like an inner class that refresh the intervals using a runnable.

Finally, intervals adapter now accept two modes:
- Stack from top: the first item of the list is on the top.
- Stack from bottom: the first item of the list is on the bottom (reverse list).


src/main/java/de/dennisguse/opentracks/adapters/IntervalStatisticsAdapter.java
```diff
@@ -48,7 +51,7 @@ public View getView(int position, @Nullable View intervalView, @NonNull ViewGrou
             viewHolder = (ViewHolder) intervalView.getTag();
         }
 
-        sumDistance_m += interval.getDistance_m();
+        sumDistance_m = actualPosition + 1 == getCount() && actualPosition > 0 ? (actualPosition * getItem(actualPosition - 1).getDistance_m()) + interval.getDistance_m() : (actualPosition + 1) * interval.getDistance_m();
```
dennisguse, line 54: Split over multiple statements.

dennisguse (approved):
Looks nice.
I like the new table-like view.

PR #434, opened 2020-09-19
Interval statistics: show unit in dropdown.

dennisguse:


src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
```diff
@@ -75,5 +75,10 @@ private void loadIntervalStats(final Track.Id trackId, IntervalOption interval)
         public int getValue() {
             return value;
         }
+
+        @Override
+        public String toString() {
+            return "" + value;
```
rgmf, line 81: Only for curiosity: why do you use ```"" + value``` instead of ```String.valueOf(value)```? Is it just a preference of syntax or a question of performance?

rgmf (commented):
Good idea :+1: 

The only thing is that I cannot see unit when I click on dropdown to select the option.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
dennisguse: It just look simpler (to me). I hope that the generated bytecode is the same.

rgmf: Thank you @dennisguse for the clarification. I understand, yes it's more simple :+1:

dennisguse:
@rgmf Unit is now visible in dropdown as well.
But we get a warning that we should not concatenate strings (as it likely to not be the same in all languages), but rather use a proper R.string (probably like `value_integer_kilometer`). 
I will take a look at this during the week.

rgmf:
@dennisguse I'm working on interval too but in other parts so any conflict shouldn't be triggered.

dennisguse:
I am not doing anything else - we can also merge this now and then we don't get in trouble at all :)

rgmf:
> I am not doing anything else - we can also merge this now and then we don't get in trouble at all :)

Good for me :+1:

rgmf:
@dennisguse only a little change to fix a little issue: https://github.com/OpenTracksApp/OpenTracks/commit/5c1f669b2bd12589904a5895c0c5f47be8e0159b

PR #466, opened 2020-10-19
ImportActivity intent filters.

rgmf:
These intent filters work for me. These are like the ones you can see in OsmAnd.

Also, read this: https://stackoverflow.com/a/8599921/8727038

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -80,6 +80,7 @@ limitations under the License.
                 <category android:name="android.intent.category.BROWSABLE" />
                 <data android:scheme="file"/>
                 <data android:scheme="content"/>
+                <data android:scheme="data"/>
                 <data android:host="*"/>
                 <data android:mimeType="*/*"/>
                 <data android:pathPattern=".*\\.gpx" />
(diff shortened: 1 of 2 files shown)


pstorch:
With this change at least open GPX files from FileManager is possible with OpenTracks, but still not sharing of GPX Tracks via Locus Maps nor OsmAnd.

pstorch:
@rgmf @dennisguse I've got Locus and OsmAnd sharing of GPX tracks working with OpenTracks. Added my changes to this branch.

Finally I found out to listen to the SEND Action with the mime types of GPX and also read the ClipData of the Intent.
Maybe we should add similar Intent filters with mime types of kmz and kml as well.

pstorch:
And we need the Intent filters with and without the mimetype. A null mimetype matches differently to a "" mimetype.

http://stackoverflow.com/questions/1733195/android-intent-filter-for-a-particular-file-extension

rgmf:
Great @pstorch 👏 

> Maybe we should add similar Intent filters with mime types of kmz and kml as well.

I think we should.

src/main/AndroidManifest.xml
```diff
@@ -72,34 +72,94 @@ limitations under the License.
...
+                <data android:pathPattern=".*\\..*\\..*\\..*\\..*\\..*\\..*\\..*\\..*\\..*\\.gpx" />
             </intent-filter>
 
             <intent-filter>
+                <action android:name="android.intent.action.SEND"/>
+                <category android:name="android.intent.category.DEFAULT"/>
+                <data android:mimeType="application/gpx+xml"/>
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.SEND"/>
```
dennisguse, line 126: Why not include BROWSABLE and VIEW here?

dennisguse (approved):
Ugly ugly ugly, but if it works :)

src/main/AndroidManifest.xml
pstorch: No idea, I don't know what other apps are using. But if it doesn't harm, we could/should add it.

And yes, the whole intent filter thingy from Android is very ugly.

rgmf: If you include VIEW together SEND a "Missing URL" error is gotten. So if we need VIEW and SEND then we have to create two intent-filter (the other solution is even more ugly: add ```tools:ignore="AppLinkUrlError"``` attribute to intent-filter).

rgmf:
For me is okay (really ugly but it works).
@pstorch when you think is ready, splash the two commits and merge it.

pstorch:
I added some more filters and also the mime types for kml and kmz, but haven't found another app which could share kml/kmz files. If you have some recommendations I would be happy to get a hint.
I'll merge the current status as it is an improvement anyway. If we find more cases, we can still further improve it.

Issue #471, opened 2020-10-22
Track import result not shown if ImportActivity not visible (anymore)


pstorch:
If the track import takes longer and the phone turns of the display, the result dialog crashes:

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.dennisguse.opentracks.debug, PID: 20559
    java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
        at androidx.fragment.app.FragmentManager.checkStateLoss(FragmentManager.java:1691)
        at androidx.fragment.app.FragmentManager.enqueueAction(FragmentManager.java:1731)
        at androidx.fragment.app.BackStackRecord.commitInternal(BackStackRecord.java:321)
        at androidx.fragment.app.BackStackRecord.commit(BackStackRecord.java:286)
        at androidx.fragment.app.DialogFragment.show(DialogFragment.java:181)
        at de.dennisguse.opentracks.fragments.ImportFinishedDialogFragment.showDialog(ImportFinishedDialogFragment.java:50)
        at de.dennisguse.opentracks.io.file.importer.ImportActivity.lambda$onImportFinished$0$ImportActivity(ImportActivity.java:70)
        at de.dennisguse.opentracks.io.file.importer.-$$Lambda$ImportActivity$plu8T-FPTkG1j8B2BGV237dSfYM.run(Unknown Source:8)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:237)
        at android.app.ActivityThread.main(ActivityThread.java:8016)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1076)
```

Would it be better to replace this dialog by a regular screen as implemented in the ExportActivity?

rgmf:
@pstorch this would be a logical step. In fact, I'd like to do it this weekend (we'll see).

My intention is doing the same we have now in ExportActivity with a summary:
- Imported tracks.
- Tracks that already exists.
- Errors... with the error message triggered.

If you don't mind I take this issue because I think it will be very similar to ExportActivity.

PR #472, opened 2020-10-24
Import all activity with summary

rgmf:
**Describe the pull request**
Import all activity with summary.

I catch errors from exceptions so the activity know the error (parser error, already exists error, etc) and user can see what happened if some file didn't was imported (see screenshots).

It's similar to ExportActivity and its funcionality is the same including the back button functionality for cancel process.

I've created a IoUtils file for common utilities (they are used by Export and Import).

This is a draft PR because it needs to be tested and I'd like to review the code and see if some part can be optimized but I publish it so you have the oportunity of testing it too - if you want and can :D
So, it's not ready yet but if you have some comment then the feedback is welcome.

Some screenshots:

imagen

imagen

imagen

**Link to the the issue**
#471 

**License agreement**


pstorch:
I just gave it a quick shot. The new screen works as far as I tested. And double back to cancel the import as well.

While testing I recognized a very slow import. Mainly because I reimported my exported tracks which I still have in OpenTracks and I have this "Prevent re-import of tracks" setting enabled. This causes each time the Unique constraint exception and then the cleanup. The cleanup deletes the imported track incl. a vacuum of the database which takes about 19 seconds each time.
Wouldn't it be better to first check for the UUID and if it already exist don't import the track so it doesn't need to be removed afterwards? I can open a separate issue for it, because it is not related to the UI change.

```
2020-10-24 23:36:07.316 4276-4786/de.dennisguse.opentracks.debug E/AbstractFileTrackImporter: Unable to import file
    android.database.sqlite.SQLiteConstraintException: UNIQUE constraint failed: tracks.uuid (code 2067 SQLITE_CONSTRAINT_UNIQUE[2067])
        at android.database.sqlite.SQLiteConnection.nativeExecuteForChangedRowCount(Native Method)
        at android.database.sqlite.SQLiteConnection.executeForChangedRowCount(SQLiteConnection.java:1074)
        at android.database.sqlite.SQLiteSession.executeForChangedRowCount(SQLiteSession.java:756)
        at android.database.sqlite.SQLiteStatement.executeUpdateDelete(SQLiteStatement.java:66)
        at android.database.sqlite.SQLiteDatabase.updateWithOnConflict(SQLiteDatabase.java:2220)
        at android.database.sqlite.SQLiteDatabase.update(SQLiteDatabase.java:2166)
        at de.dennisguse.opentracks.content.provider.CustomContentProvider.update(CustomContentProvider.java:281)
        at android.content.ContentProvider$Transport.update(ContentProvider.java:420)
        at android.content.ContentResolver.update(ContentResolver.java:2000)
        at de.dennisguse.opentracks.android.ContentResolverWrapper.update(ContentResolverWrapper.java:37)
        at de.dennisguse.opentracks.content.provider.ContentProviderUtils.updateTrack(ContentProviderUtils.java:305)
        at de.dennisguse.opentracks.io.file.importer.AbstractFileTrackImporter.onTrackEnd(AbstractFileTrackImporter.java:279)
        at de.dennisguse.opentracks.io.file.importer.KmlFileTrackImporter.endElement(KmlFileTrackImporter.java:121)
        at org.apache.harmony.xml.ExpatParser.endElement(ExpatParser.java:164)
        at org.apache.harmony.xml.ExpatParser.appendBytes(Native Method)
        at org.apache.harmony.xml.ExpatParser.parseFragment(ExpatParser.java:523)
        at org.apache.harmony.xml.ExpatParser.parseDocument(ExpatParser.java:484)
        at org.apache.harmony.xml.ExpatReader.parse(ExpatReader.java:322)
        at org.apache.harmony.xml.ExpatReader.parse(ExpatReader.java:285)
        at javax.xml.parsers.SAXParser.parse(SAXParser.java:390)
        at javax.xml.parsers.SAXParser.parse(SAXParser.java:187)
        at de.dennisguse.opentracks.io.file.importer.AbstractFileTrackImporter.importFile(AbstractFileTrackImporter.java:135)
        at de.dennisguse.opentracks.io.file.importer.KmlFileTrackImporter.importFile(KmlFileTrackImporter.java:38)
        at de.dennisguse.opentracks.io.file.importer.KmzTrackImporter.parseKml(KmzTrackImporter.java:260)
        at de.dennisguse.opentracks.io.file.importer.KmzTrackImporter.findAndParseKmlFile(KmzTrackImporter.java:188)
        at de.dennisguse.opentracks.io.file.importer.KmzTrackImporter.importFile(KmzTrackImporter.java:83)
        at de.dennisguse.opentracks.io.file.importer.ImportService.importFile(ImportService.java:70)
        at de.dennisguse.opentracks.io.file.importer.ImportService.onHandleWork(ImportService.java:50)
        at androidx.core.app.JobIntentService$CommandProcessor.doInBackground(JobIntentService.java:392)
        at androidx.core.app.JobIntentService$CommandProcessor.doInBackground(JobIntentService.java:383)
        at android.os.AsyncTask$3.call(AsyncTask.java:378)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:919)
2020-10-24 23:36:07.316 4276-4786/de.dennisguse.opentracks.debug W/CustomContentProvider: Deleting table trackpoints
2020-10-24 23:36:07.325 4276-4786/de.dennisguse.opentracks.debug W/CustomContentProvider: Deleting table waypoints
2020-10-24 23:36:07.327 4276-4786/de.dennisguse.opentracks.debug W/CustomContentProvider: Deleting table tracks
2020-10-24 23:36:07.332 4276-4786/de.dennisguse.opentracks.debug I/CustomContentProvider: Vacuuming the database.
```

BTW: I also recognized the slow vacuum of the database in debug mode when deleting a few tracks from the TracksList.

dennisguse:
That was my doing - my problem was that the SAXParser adds data in the order it is written in the file and to not buffer in memory, the data is written to db immediately. To implement the UUID quickly, I added the delete when duplication is detected. But yiu are right, it is not cool. Let's move it to a new issue.

@rgmf nice work. Will check today :)

src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
```diff
@@ -126,7 +127,7 @@ public void characters(char[] ch, int start, int length) {
     }
 
     @Override
-    public Track.Id importFile(InputStream inputStream) {
+    public Track.Id importFile(InputStream inputStream)  throws SAXException, SQLiteConstraintException, IOException, ParserConfigurationException {
```
pstorch, line 130: Offtopic: have I already expressed my dislike of Javas checked Exceptions? 

This long list could be omitted by wrapping it in a RuntimeException at line 144, but it is a matter of taste.

rgmf: I don't like either.

Do you mean something like this (in the line 144)?
```java
throw new RuntimeException(e);
```

Then, in ImportService, Can I know what exception was triggered? I need it so I know if the error was an already exists (SQLiteConstraintException) or a parser error (SAXException or ParserConfigurationException), so I can inform to the final user in ImportActivity.

rgmf:
@pstorch @dennisguse, I agree with you, the issue @pstorch point out should be fix in another issue because it's another problem.

src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
pstorch: That I didn't check, so you need to know the type of exception where you catch it?
You can call getCause() to get the wrapped exception, but maybe this is not the ideal thing to do here.

rgmf: Yes, I need the type of exception to show to the user the cause of the errors in the ImportActivity (look the third screenshot). I do that because I think is really important the user knows the cause of the error (silent errors are unuseful ones). This way the user knows if it is OpenTracks's fault or its files are badly parsed, for example.

rgmf:
> Will check today :)

There is a known bug. To reproduce: import a lot of files, rotate the screen, turn off the screen, turn on the screen before finish process -> the counts fails.

If you find some other bug let me know.

src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
pstorch: A solution in the middle could be to only distinguish between the SQLiteConstraintException and the others. The first is because of duplicate UUID and the others represent errors reading the file. 
But then you would need two catch clauses, 1 for SQLiteConstraintException which is just rethrown an one for the others to get wrapped into a RuntimeException. 
I suggest not spending too much time here now. When we refactor the UUID check, this might change as well.

Or dirty: just declare Exception here and don't care. 😉

rgmf: Thank you @pstorch 
I'll look at this.

dennisguse: My two cents: don't pass a RuntimeException or even Exception.
I have seen larger projects where in higher layers of the code have hard time to figure out what broke (just by reading the code).
Nowadays, I am big fan of just declaring a custom (so to say "business") exception for each case the implementation needs to distinguish and add the originally catched exception as cause.
Then it is easy to understand which exceptions were actually expected and the internal implementation logic is not exposed.

PS: don't declare exception as it is really hard to understand what potential issues might arise.
PPS: The last months I spent a lot of time refactoring and saw things like this: 
```
try {
 Object a = // call to some internal object
 return a;
} catch (Exception e) {
  return null;
}
```
And we are still working on such things, as it takes quite some time to figure out why it is working at all (and if the catch is actually needed etc.). And many cases the exception originates from somewhere in some library and it was passed to some higher layer and then somebody added a try-catch there.
Not too much fun...

pstorch: I feel your pain from larger codebases. Saw all kinds of strange exception handling.

I would suggest to derive the BusinessException from RuntimeException, so you don't have to declare it everywhere only catch where you know what to do.

rgmf: In my last commit I've added custom exceptions to distinguish the origin of the error's cause.

I don't know if I've understanded you, @dennisguse and @pstorch is that what you talked about to do? It's something I've never done...

src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
```diff
@@ -127,21 +126,26 @@ public void characters(char[] ch, int start, int length) {
...
+                // TODO Multi track is not supported yet.
+                throw new ImportParserException("Multi track not supported");
             }
             return trackIds.get(0);
-        } catch (MultiTracksImportException | IOException | SAXException | ParserConfigurationException | SQLiteConstraintException e) {
+        } catch (IOException | SAXException | ParserConfigurationException e) {
             Log.e(TAG, "Unable to import file", e);
             cleanImport();
-            throw e;
+            throw new ImportParserException(e.getMessage());
```
pstorch, line 144: Put the whole Exception into the constructor not only the message. Then later you can call getCause() to get the original exception including Stack trace

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
```diff
@@ -71,7 +66,7 @@ public KmzTrackImporter(Context context, Uri uriFile) {
     }
 
     @Override
-    public Track.Id importFile(InputStream inputStream) throws SAXException, SQLiteConstraintException, IOException, ParserConfigurationException {
+    public Track.Id importFile(InputStream inputStream) throws ImportParserException, ImportAlreadyExistsException {
```
pstorch, line 69: Throws declaration of the two new Exceptions is not necessary, because they are unchecked Exceptions.

src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
rgmf: Okay :+1:

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
rgmf: Ops!!! Yes, of course.

src/main/java/de/dennisguse/opentracks/io/file/ErrorListDialog.java
```diff
@@ -0,0 +1,35 @@
...
+import de.dennisguse.opentracks.R;
+
+public class ErrorListDialog extends DialogFragment {
+
+    public static final String TAG = ErrorListDialog.class.getSimpleName();
+
+    private List<String> trackErrors;
+    private String title;
+
+    public ErrorListDialog(String title, List<String> trackErrors) {
```
dennisguse, line 20: is it sufficient to pass parameters via the constructor?
If the activity will be restored all fragments should be created from a bundle.
Like this: 
```

    public static void showDialog(FragmentManager fragmentManager, Uri documentFileUri, boolean isDirectory) {
        Bundle bundle = new Bundle();
        bundle.putParcelable(EXTRA_DATA_URI_KEY, documentFileUri);
        bundle.putBoolean(EXTRA_IS_TREE_KEY, isDirectory);

        ImportProgressDialogFragment dialogFragment = new ImportProgressDialogFragment();
        dialogFragment.setArguments(bundle);
        dialogFragment.setRetainInstance(true);
        dialogFragment.show(fragmentManager, IMPORT_DIALOG_TAG);
    }
```

src/main/java/de/dennisguse/opentracks/io/file/importer/ImportActivity.java
```diff
@@ -16,62 +16,255 @@
...
+    private ImageView viewAlertIcon;
+    private TextView viewAlertMsg;
+    private Button viewLeftButton;
+    private Button viewRightButton;
+
+    boolean doubleBackToCancel = false;
+
+    private Uri documentUri;
+    private boolean isDirectory;
+    private List<DocumentFile> fileList;
```
dennisguse, line 83: Initialize here.

src/main/java/de/dennisguse/opentracks/io/file/importer/ImportActivity.java
```diff
@@ -16,62 +16,255 @@
...
+    private void initImport() {
+        if (getTotalDone() == fileImportTotalCount) {
+            onImportCompleted();
+        } else {
+            importNextFile();
+        }
+    }
+
+    private void importNextFile() {
+        int i = getTotalDone();
```
dennisguse, line 220: rename to count?

src/main/java/de/dennisguse/opentracks/io/file/importer/ImportService.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final String TAG = ImportService.class.getSimpleName();
+
+    private static final int JOB_ID = 2;
+
+    private static final String EXTRA_RECEIVER = "extra_receiver";
+    private static final String EXTRA_URI = "extra_uri";
+
+    private ResultReceiver resultReceiver;
+
+    public static void enqueue(Context context, ImportServiceResultReceiver receiver, Uri uri) {
```
dennisguse, line 34: Can we change to pass a DocumentFile instead of Uri?

src/main/java/de/dennisguse/opentracks/io/file/importer/ImportServiceResultReceiver.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    protected void onReceiveResult(int resultCode, Bundle resultData) {
+        if (receiver != null) {
+            receiver.onReceiveResult(resultCode, resultData);
+        }
+    }
+
+    public void setReceiver(Receiver receiver) {
```
dennisguse, line 34: Why not pass in the constructor?
It should never be null right?

src/main/res/values/strings.xml
```diff
@@ -285,14 +285,26 @@ limitations under the License.
...
+    <string name="import_parser_error">Parser error: %1$s</string>
+    <string name="import_unable_to_import_file">Unable to import file: %1$s</string>
+    <string name="import_error_info">%1$s: %2$s</string>
+    <string name="import_error_list_dialog_title">Files not imported</string>
+    <string name="import_file_imported">File %1$s imported</string>
     <string name="import_error">Imported %1$d of %2$s from %3$s</string>
     <string name="import_no_directory">%1$s does not exist</string>
     <string name="import_no_file">There were no files to import. Connect your device to your computer and place the files to import in %1$s.</string>
     <string name="import_no_file_title">Import files</string>
-    <string name="import_progress_message">Importing from %1$s&#8230;</string>
```
dennisguse, line 292: IntelliJ's code analyze complains about this change.
The &#8230; is actually the correct thing for "..."

dennisguse (commented):
I really like the use of a JobIntentService.
Great idea.
Then we will properly behave without CPU locks or similar!

src/main/java/de/dennisguse/opentracks/io/file/ErrorListDialog.java
rgmf: You're right. What a screw up! I didn't test it :see_no_evil:

src/main/java/de/dennisguse/opentracks/io/file/importer/ImportService.java
rgmf: No, we can't. DocumentFile doesn't implement Parcelable so we cannot pass it through Extra.

rgmf:
@dennisguse @pstorch We have a handicap when we import and there is a screen rotation.

Imagine ImportActivity has imported tracks 1, 2 and 3. The track 4 is importing by the JobIntentService when screen is rotated.

ImportActivity's savedInstance has tracks 1, 2 and 3 as imported but no the 4 one. So, ImportActivity import the 4 one once again.

I don't know how to deal with it.

dennisguse:
Can we put this info into the bundle?
Something like "not yet done, but started".

rgmf:
> Can we put this info into the bundle?
> Something like "not yet done, but started".

Yes, it's an option. I'll look into this possibility. Thanks ;)

rgmf:
I've fixed with ViewModel.

Now, we have:
- an Activity (ImportActivity) that mantain the view data and handle screen rotation,
- a Model (ImportViewModel) that gets the data from the Service,
- a JobIntentService (ImportService) that does the task of importing files.

In that way, we don't have an Activity that handle a lot of data (before it has view and model). Now we have three little pieces doing a specially thing.

ImportActivity observes the Model and change its views.

pstorch:
So many new things to learn for me. Great.

rgmf:
Ready for me... Sunday I'll squash commits and merge onto main if any complain ;)

pstorch:
Works great. :+1:

dennisguse:
Works great! 
@rgmf I just added a small cleanup commit, please also squash :)

rgmf:
@dennisguse thank you... always learning with your changes/cleanups/adjustments

dennisguse:
@rgmf Me too :partying_face:

PR #480, opened 2020-10-29
Database enforces foreign key constraints incl. cascading deletes.

dennisguse:
So far the three tables were independent of each other (from a database perspective).
I added now foreign key constraints incl. cascading deletes for trackId.
Thus, if a track get's deleted the database automatically deletes all referencing trackpoints and marker automatically.

However, this has a drawback: There is only _one_ notification for consent observer (i.e., the track got deleted). So far, we sent one for each trackpoint and marker as well.

Also the trackId column is now NOT NULL.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/content/provider/CustomSQLiteOpenHelper.java
+++ b/src/main/java/de/dennisguse/opentracks/content/provider/CustomSQLiteOpenHelper.java
@@ -25,7 +25,7 @@ public class CustomSQLiteOpenHelper extends SQLiteOpenHelper {
 
     private static final String TAG = CustomSQLiteOpenHelper.class.getSimpleName();
 
-    private static final int DATABASE_VERSION = 27;
+    private static final int DATABASE_VERSION = 28;
 
     @VisibleForTesting
     public static final String DATABASE_NAME = "database.db";
@@ -73,7 +73,9 @@ public class CustomSQLiteOpenHelper extends SQLiteOpenHelper {
                 case 27:
                     upgradeFrom26to27(db);
                     break;
-
+                case 28:
+                    upgradeFrom27to28(db);
+                    break;
 
                 default:
                     throw new RuntimeException("Not implemented: upgrade to " + toVersion);
@@ -97,6 +99,9 @@ public class CustomSQLiteOpenHelper extends SQLiteOpenHelper {
                     break;
                 case 26:
                     downgradeFrom27to26(db);
+                    break;
+                case 27:
+                    downgradeFrom28to27(db);
                     break;
 
                 default:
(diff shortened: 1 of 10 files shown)


src/main/java/de/dennisguse/opentracks/content/provider/CustomSQLiteOpenHelper.java
```diff
@@ -220,4 +234,54 @@ private void downgradeFrom27to26(SQLiteDatabase db) {
...
+        db.setTransactionSuccessful();
+        db.endTransaction();
+    }
+
+    private void downgradeFrom28to27(SQLiteDatabase db) {
+        db.beginTransaction();
+
+        // TrackPoints
+        db.execSQL("ALTER TABLE trackpoints RENAME TO trackpoints_old");
+        db.execSQL("CREATE TABLE trackpoints (_id INTEGER PRIMARY KEY AUTOINCREMENT, trackid INTEGER NOT NULL, longitude INTEGER, latitude INTEGER, time INTEGER, elevation FLOAT, accuracy FLOAT, speed FLOAT, bearing FLOAT, sensor_heartrate FLOAT, sensor_cadence FLOAT, sensor_power FLOAT, elevation_gain FLOAT, FOREIGN KEY (trackid) REFERENCES tracks(_id) ON UPDATE CASCADE ON DELETE CASCADE)");
```
pstorch, line 270: downgrade contains the foreign key?

pstorch (approved):
So far it looks good and works for me. Didn't know one has to enable foreign key processing in SQLite to work.

src/main/java/de/dennisguse/opentracks/content/provider/CustomSQLiteOpenHelper.java
dennisguse: Thanks. Fixed.

pstorch:
Interesting: while testing this PR I first saw that OpenTracks implemeted DB downgrade methods. I've never used that for other Apps so far and I was wondering if this really works.
Now I switched to the import_activity branch and saw that it is not working:

```
2020-10-30 20:00:54.644 24381-24381/de.dennisguse.opentracks.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.dennisguse.opentracks.debug, PID: 24381
    java.lang.RuntimeException: Unable to get provider de.dennisguse.opentracks.content.provider.ShareContentProvider: java.lang.RuntimeException: Not implemented: downgrade to 27

```

Actually it can't work: how should an older App know how to downgrade from a newer version? Is there any situation where the downgrade is usefull? No offense: I'm always learning. There might be a situation I'm not aware of.

rgmf (approved):

rgmf:
The only thing I've done is using this branch and check that it works with my recorded tracks.

Reading the code it seems good for me :+1:

PR #487, opened 2020-10-31
optimize multiple track deletion

pstorch:
After #480 got implemented, we can now delete multiple tracks in one go and only cause DB vacuum once.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
@@ -33,6 +33,7 @@ import androidx.loader.content.Loader;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
```diff
@@ -179,6 +180,17 @@ public void deleteAllTracks(Context context) {
...
     }
 
+    public void deleteTracks(Context context, @NonNull List<Track.Id> trackIds) {
+        // Delete track folder resources.
+        for (Track.Id trackId : trackIds) {
+            FileUtils.deleteDirectoryRecurse(FileUtils.getPhotoDir(context, trackId));
+        }
+
+        // Delete track last since it triggers a database vacuum call
+        String whereClause = String.format(TracksColumns._ID + " IN (%s)", TextUtils.join(",", Collections.nCopies(trackIds.size(), "?")));
```
dennisguse, line 190: Love this!
Will update ShareContentprovider :+1:

dennisguse (approved):
+1
I did not see that one :rofl:

PR #489, opened 2020-11-01
Shrinking all APK builds without obfuscation.

### LINKED ISSUE #433: Shrink APK?
Not that we need to this, but it would be cool (from an engineering perspective) to reduce the APK size.
So, far OpenTracks is less than 10Mb - which is already great :+1: 

https://developer.android.com/topic/performance/reduce-apk-size

dennisguse:
We can save quite some space by using shrinking the APK (removing unused things; I guess mainly support library stuff) without obfuscation.

Screenshot from 2020-11-01 09-12-42

Merge after next release.
Needs some serious testing...

Fixes #433.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -57,7 +57,6 @@ android {
         buildConfigField "String", "VERSION_NAME_FULL", "\"${getVersionName()}\""
 
 
-
         minSdkVersion 21
         targetSdkVersion 30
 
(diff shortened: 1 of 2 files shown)


dennisguse:
But tests are failing as some methods used by tests only are removed.

dennisguse:
Got it: for testing, we need to build without shrinking as otherwise methods are removed used by tests only.
Therefore, the _debug_ build now builds without shrinking.

That means, we need to test the release build.
But we don't use reflection anywhere, so we should be good to go.

dennisguse:
Seems to be working.

PR #494, opened 2020-11-01
fix #491 crash export on android 11

### LINKED ISSUE #491: Crash when trying to export all
**Describe the bug**
When I try to export all of my tracks the application crashes.

**To Reproduce**
1. Tap the ellipses in the top-right
2. Tap settings
3. Tap "Export all"
4. Browse to the desired destination and then tap the "Use this folder" button
5. The application crashes.

**Technical information**
 - Pixel 3 XL
 - Android 11
 - OpenTracks v3.11.1-a42bd432e installed from F-Droid.

pstorch:
I tested the bug on emulater with android 11. The problem didn't happen with OSMDashboard, so I compared the code and found some differences.

@rgmf @dennisguse Please have a look.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/io/file/importer/DirectoryChooserActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/importer/DirectoryChooserActivity.java
@@ -43,6 +43,10 @@ public abstract class DirectoryChooserActivity extends AppCompatActivity {
             if (resultCode == Activity.RESULT_OK) {
                 Uri directoryUri = resultData.getData();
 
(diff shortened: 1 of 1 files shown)


rgmf:
@pstorch this works in a Pixel 3a with Android 11 :smiley: 

But, every time you access to folder selection you cannot see the bottom button to select the folder where you are. You have to back and enter again.

Anyway, the more important thing is fixed.

pstorch:
@rgmf thanks for testing. Yes, I saw the missing button as well. Probably because the uri permission to that directory is still granted.

dennisguse:
Then let's release :)

Issue #492, opened 2020-11-01
`SearchListActivity` and `SearchEngine` need love


dennisguse:
I guess, I read `SearchListActivity` and `SearchEngine` for the first time...

Issues:
* everything is loaded from database
* data is kept in memory
* data is unmarshalled, marshalled, and then unmarshalled again (to different data structures)

What would need to be done:
1. `SearchEngine.search()` returns Tracks/Markers (all data is loaded although not everything is used)
    Suggestion: Only return trackId or markerId
2. SearchListActivity gets data directly from `SearchEngine.search()` (everything is in memory)
   Suggestions: user a CursorLoader and ResourceAdapter (like in TrackListActivity)
   Try to use only one query to the database (UNION of Track and Marker) and only load required data.
3. (if possible) do `SearchEngine.search()` computation using SQL on database level

https://developer.android.com/reference/android/database/sqlite/SQLiteQueryBuilder.html#buildUnionQuery(java.lang.String%5B%5D

Follow-up to https://github.com/OpenTracksApp/OpenTracks/issues/328

dennisguse:
Note: performance of the current implementation must be really bad.
So, for low-end devices (with a lot of tracks/markers stored), we can really improve here.

dennisguse:
We could even go one step further: remove SearchEngine completely as it only reorders the results according to an assumed importance.
From my perspective: a lot of code with very limited actual functionality.

rgmf:
> We could even go one step further: remove SearchEngine completely as it only reorders the results according to an assumed importance.
> From my perspective: a lot of code with very limited actual functionality.

I've used it for a couple of times or so. Don't you use this feature? It seems useful, doesn't it?

PS/ I've never read the code and I don't know that part.

dennisguse:
I meant to remove the reordering the results and do it on DB level only. The search would still provide the same  results, but the code would be faster and easier.

dennisguse:
@rgmf @pstorch I guess, I know the answer, but just to be sure :)
So far our search functionality searched for _tracks and markers_ (in name, description, and category).
What about separating the search functionality?
If search is triggered from the track list, then search for tracks.
If search is triggered from the marker list, then search for markers.

This would simplify our code a lot (no changes to CustomContentProvider), but we lose some functionality.
For marker+track search, I need to create a really relatively complicated SQL union in CustomContentProvider.
And SQLiteQueryBuilder is not helpful (or i did not understood it - very likely ><).

[...]

PR #496, opened 2020-11-03
Search now uses database cursor (separate for tracks and markers)

dennisguse:
Working on #492.

Plan is to use a DB cursor directly to show the present the data.


dennisguse:
It is already working for tracks alone.

next step: create a union query for tracks & markers.

rgmf:
It only works if you type the exact name. For instance, if I have a track called "Bigotilla, Romeral y Pouel" and type "bigo" then it doesn't appear in the result (and in previous versions it appeared).

dennisguse:
@rgmf While cleaning up, I accidentially removed this functionality.
Thanks!

PR #511, opened 2020-11-09
Activity icon uses now an image instead of a spinner (which was only …

dennisguse:
Back in the good old day the activity icon was a spinner.
But since ages ago, the spinner was actually not used, but it's UI replaced with an dialog.
So, let's get rid of the spinner ;)


dennisguse:
My initial goal was to get EspressoUI running and I up refactoring the buttons while EspressoUI is still not working for the start/stop button ;)
@pstorch Can you take a look?
I basically replaced the thread by using the handler and moved the animations into XML.

pstorch:
@dennisguse I'll have a look tomorrow evening.

dennisguse:
No worries; not urgent :)

pstorch:
@dennisguse now the animation is the other way around from no red to full red. It's hardly seen that there is something happening. Also the feedback with the vibration is at the end.

dennisguse:
Then I have to check how to get the animation interverted. I was happy about the fadein as everything is now in one xml and we don't do anything in the code. Let me check.

Vibration: will change it.

dennisguse:
@pstorch This animation stuff is crazy. Or it was build to make people crazy.
Anyhow: in https://github.com/OpenTracksApp/OpenTracks/pull/511/commits/901dbe7ba96beec02e4941d46f793ae5b42b4294 I change the background color of the TrackController-view.
Now the animation is really visible as the whole space is used instead of only the button, which is likely below the finger.

pstorch:
> @pstorch This animation stuff is crazy. Or it was build to make people crazy.

I feel your pain. It also caused me a few hours plus this strange bug we had on some devices.
If it is simpler, we could replace the animation with a simpler switch of the color.

dennisguse:
And I just (finally) figured out, how to create screen recordings in the emulator ><.
That's how my proposal:
output

@pstorch What do you think?

pstorch:
@dennisguse that is definitely better :+1:

Issue #514, opened 2020-11-11
KMZ files Import: markers's images aren't imported


rgmf:
**Describe the bug**
Images from markers in KMZ files aren't imported.

Working on #473  issue I stumple upon with this (potencial) bug.

I'm trying to fix this one together #473 issue.

**To Reproduce**
1. Go to import all.
2. Select a folder with one KMZ file with images.
3. Images aren't not imported with its markers.

rgmf:
Delete marker crashes app when user select more than one marker.

PR #519, opened 2020-11-14
Import marker photos

rgmf:
**Describe the pull request**
Three bugs fixed:
1. Import marker with photos didn't saved the photos.
2. Edit marker could leave not linked photos in track directory.
3. Delete marker didn't work.

**Link to the the issue**
#514 

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/MarkerListActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/MarkerListActivity.java
@@ -237,9 +237,6 @@ public class MarkerListActivity extends AbstractActivity implements DeleteMarker
                 }
                 return true;
             case R.id.list_context_menu_delete:
-                if (markerIds.length > 1 && markerIds.length == viewBinding.markerList.getCount()) {
-                    markerIds = null;
-                }
                 DeleteMarkerDialogFragment.showDialog(getSupportFragmentManager(), markerIds);
                 return true;
             case R.id.list_context_menu_select_all:
--- a/src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
@@ -68,7 +68,6 @@ abstract class AbstractFileTrackImporter extends DefaultHandler implements Track
     private final ContentProviderUtils contentProviderUtils;
     private final int recordingDistanceInterval;
 
-    private Track.Id importTrackId;
     private final List<Track.Id> trackIds = new ArrayList<>();
     private final List<Marker> markers = new ArrayList<>();
 
@@ -102,11 +101,6 @@ abstract class AbstractFileTrackImporter extends DefaultHandler implements Track
         this.context = context;
         this.contentProviderUtils = contentProviderUtils;
         this.recordingDistanceInterval = PreferencesUtils.getRecordingDistanceInterval(context);
-    }
-
-    AbstractFileTrackImporter(Context context, ContentProviderUtils contentProviderUtils, Track.Id importTrackId) {
-        this(context, contentProviderUtils);
-        this.importTrackId = importTrackId;
     }
 
     @Override
@@ -181,6 +175,10 @@ abstract class AbstractFileTrackImporter extends DefaultHandler implements Track
                         // No more markers
                         return;
                     }
+                    // If marker had photo it must be translated to internal photo url (depend on track id)
+                    if (marker.hasPhoto()) {
+                        marker.setPhotoUrl(getInternalPhotoUrl(marker.getPhotoUrl()));
+                    }
                 }
 
                 if (trackPoint == null) {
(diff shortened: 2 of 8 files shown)


src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
```diff
@@ -69,20 +68,18 @@ public KmzTrackImporter(Context context, Uri uriFile) {
...
-        if (!copyKmzImages()) {
-            cleanImport(context, importTrackId);
-            return null;
-        }
-
         try {
             trackId = findAndParseKmlFile(inputStream);
         } catch (Exception e) {
-            cleanImport(context, importTrackId);
             throw e;
```
dennisguse, line 74: This try-catch is not doing anything.
Is it not necessary to cleanup the failed import on exception?

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
```diff
@@ -69,20 +68,18 @@ public KmzTrackImporter(Context context, Uri uriFile) {
...
         }
 
         if (trackId == null) {
```
dennisguse, line 77: I guess this should not happen.
We could improve this by making findAndParseKmlFile to always throw an exception if it would not return a trackId.

src/main/java/de/dennisguse/opentracks/util/FileUtils.java
```diff
@@ -289,6 +289,30 @@ public static File getPhotoFileIfExists(Context context, Track.Id trackId, Uri u
...
 
+    /**
+     * Builds interval photo file object for fileNameUri Uri.
+     *
+     * @param context     the Context.
+     * @param trackId     the id of the Track.
+     * @param fileNameUri the file name uri.
+     * @return            file object or null.
+     */
+    public static File buildInternalPhotoFile(Context context, Track.Id trackId, Uri fileNameUri) {
```
dennisguse, line 300: fileNameUri should be @NonNull and the null check should be done by the caller (if needed).

src/main/java/de/dennisguse/opentracks/io/file/importer/AbstractFileTrackImporter.java
```diff
@@ -104,11 +104,6 @@
         this.recordingDistanceInterval = PreferencesUtils.getRecordingDistanceInterval(context);
     }
 
-    AbstractFileTrackImporter(Context context, ContentProviderUtils contentProviderUtils, Track.Id importTrackId) {
-        this(context, contentProviderUtils);
-        this.importTrackId = importTrackId;
```
dennisguse, line 109: Attribute must be removed completely (it is always null) and we have an `if (importTrackId == null)` on this class.

dennisguse (changes requested):
Not yet sure if this works....

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
rgmf: The cleanImport needs the track id but if findAndParseKmlFile fails then trackId is not known (now the track is created after import is finishes).

dennisguse: Then we still don't need to catch the exception and throw the same one - so we can remove this try catch, right?

src/main/java/de/dennisguse/opentracks/MarkerListActivity.java
```diff
@@ -237,9 +237,6 @@ private boolean handleContextItem(int itemId, long... longMarkerIds) {
                 }
                 return true;
             case R.id.list_context_menu_delete:
-                if (markerIds.length > 1 && markerIds.length == viewBinding.markerList.getCount()) {
```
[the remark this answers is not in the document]
dennisguse: markerIds must be null, so that DeleteMarkerDialogFragment will show the "delete all message".
Don't know if the current is code is working without this if - markerIds might be an initialized array with length=0.
And then the deletion will not happen anymore.

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
rgmf: What exception should it throw if not trackId won't be returned? RuntimeException?

src/main/java/de/dennisguse/opentracks/MarkerListActivity.java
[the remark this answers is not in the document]
rgmf: If markerIds is null then OT crashes. Look at 82 DeleteMarkerDialogFragment's line (it uses markerIds but it can be null):

```java
for (Marker.Id markerId : markerIds) {
    contentProviderUtils.deleteMarker(getContext(), markerId);
}
```

All that code was only for show "Delete all markers?" instead of "Delete selected?" in the title of the Dialog.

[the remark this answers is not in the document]
rgmf: In my last commit is my proposal (it can be undo :D)

dennisguse:
@rgmf Can you squash and merge?
Code looks good, but how we handle images is way too complicated.
Don't know what to do about it, but I am not feeling to comfortable about it.
But for now it is good enough ;)

rgmf:
> @rgmf Can you squash and merge?

Of course.

> ...but how we handle images is way too complicated.
> Don't know what to do about it, but I am not feeling to comfortable about it.

I agree with you (I'm not really happy with the result). I think we should split MarkerEditActivity in Model/View because there are too much logic about data in the activity. This week I'm a little busy. Next week I have much more time to check it out.

dennisguse:
What about moving the images into the DB?
one blob column + one mime type column?
Files are anyhow not accessible from file system; we "would" just need to extend the contentprovider to provide proper access, so that the view show the images.

rgmf:
> What about moving the images into the DB?
I've always thought that it's not a good idea have images in the database like blob... Is that typical?

I'll think about it and the next week I'll try to do it.

PS/ Should we crop photos for markers or do something with large photos? I have a problem when I go to marker's list with 20 or 30 large photos.

dennisguse:
About DB: there are a lot of opinions on the Internet - don't know if it would be good for us.
Performance might be fine or not. That is hard to figure out and might depend on the actual device.
https://www.sqlite.org/intern-v-extern-blob.html

About the slow loading: we could go for https://developer.android.com/reference/android/media/ThumbnailUtils

PR #526, opened 2020-11-21
UI tuning: enable right-to-left, interval headings, and ChartView's fontSize

dennisguse:
We have some translations for RTL languages, so why not provide also handle RTL properly (we get rid of some warnings).


rgmf (commented):
Good for me. Texts on ChartView works perfect when I change font size configuration.

Only one thing. In `interval_stats_list_item.xml` in rtl languages we see something like this:



Like you can see the up arrow is next to speed instead of gain. If you add `android:layoutDirection="ltr"` in `LinearLayout` that contains the arrow and the gain then we'll have something like this:



Anyway, I don't know how rtl language works but unit should be in the left of the number, the other way around that in ltr languages, right? Seem logical but I don't have idea about it :sweat_smile:

dennisguse:
@rgmf 
Thanks for this tiny detail: I have a simplification for this (need to test, no time before work).

About the units - the answer seems to be: it depends according to https://material.io/design/usability/bidirectionality.html#mirroring-layout (somewhere in the middle) ;)

I guess, we should use those:
```
    <string name="value_float_feet">%1$.2f ft</string>
    <string name="value_float_kilometer">%1$.2f km</string>
    <string name="value_float_meter">%1$.2f m</string>
    <string name="value_float_mile">%1$.2f Meilen</string>
    <string name="value_integer_feet">%1$d Fuß</string>
    <string name="value_integer_kilometer">%1$d km</string>
    <string name="value_integer_kilometer_hour">%1$d km/h</string>
    <string name="value_integer_meter">%1$d m</string>
    <string name="value_integer_mile">%1$d mi</string>
    <string name="value_integer_mile_hour">%1$d Meile/h</string>
    <string name="value_integer_minute">%1$d min</string>
    <string name="value_integer_percent">%1$d %%</string>
    <string name="value_integer_second">%1$d s</string>
```

and replace those: `viewHolder.rate.setText(speedParts.first + " " + speedParts.second);`

rgmf (approved):

dennisguse:
@rgmf Thanks for testing 👍

PR #538, opened 2020-11-28
fix #537 ignore waypoint markers with invalid coordinates on import

### LINKED ISSUE #537: Ignore deprecated statistics waypoint on import
It still bugs me that I cannot import my old tracks.
I once tried to ignore invalid coordinates, but then the discussion on how to handle the pause/resume trackpoints in the database and the im-/export files came up. See #316 

Pause and Resume use latitude 100 and 200 with longitude 0, but the invalid coordinates from my old kmz exports contain latitude 100 and longitude 180. So I dig a little bit deeper. From the kmz file I saw it is a Placemark with "Split 0" in the name:

```
<Placemark>
<name><![CDATA[Split 0]]></name>
<description><![CDATA[Total distance: 0.00 km (0.0 mi)
Total time: 00:00
Moving time: 00:00
Average speed: 0.00 km/h (0.0 mi/h)
Average moving speed: 0.00 km/h (0.0 mi/h)
Max speed: 0.00 km/h (0.0 mi/h)
Average pace: 0:00 min/km (0:00 min/mi)
Average moving pace: 0:00 min/km (0:00 min/mi)
Fastest pace: 0:00 min/km (0:00 min/mi)
Max elevation: -9223372036854775808 m (-9223372036854775808 ft)
Min elevation: 9223372036854775807 m (9223372036854775807 ft)
Elevation gain: 0 m (0 ft)
Max grade: 0 %
Min grade: 0 %
Recorded: 6/23 11:29
]]></description>
<TimeStamp><when>1970-01-01T00:00:00Z</when></TimeStamp>
<styleUrl>#waypoint</styleUrl>
<Point>
<coordinates>180.0,100.0</coordinates>
</Point>
</Placemark>
```

Newer files don't contain this Placemark anymore.

Then I digged a little bit in the git history and found this commit:
https://github.com/OpenTracksApp/OpenTracks/commit/09d960c7a1228e4822410abedf3cef9e45440ef5

So this is not used anymore. 

Can we ignore those Markers while importing? Markers are not relevant for the Track itself, right? So ignoring markers with invalid coordinates wouldn't hurt. If someone thinks we should still stop the import on invalid marker coordinates, I can offer to only ignore those obsolete statistic markers.

pstorch:


src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
```diff
@@ -194,7 +194,7 @@ private boolean hasImageExtension(String fileName) {
      * @param trackId the id of the Track.
      */
     private void deleteOrphanImages(Context context, Track.Id trackId) {
-        if (!trackId.isValid()) {
+        if (trackId != null && !trackId.isValid()) {
```
dennisguse, line 197: This should not be necessary as `findAndParseKmlFile()` either throws an exception or returns a trackid, right?

dennisguse (approved):
Looks good.

Love the test :)

PS: I am just not sure about the null check (I removed this some month ago).
Do we need to add it again?

src/main/java/de/dennisguse/opentracks/io/file/importer/KmzTrackImporter.java
pstorch: I came across this when I first experimented with a kmz file for the Unit test. But I had problems with it an the while ZipEntry didn't find any files in the zip input stream. Then it resulted in a null trackId.

dennisguse: Then we should throw an exception in `findAndParseKmlFile` as the file is broken or am I mistaken?

pstorch: Yes, you are right.

pstorch:
Updated the PR with a second commit to handle the "No doc.kml file found in KMZ" case.

PR #541, opened 2020-12-04
BLE sensors: Miband3's is now considered a heart rate sensor during discovery

### LINKED ISSUE #447: Since Bluetooth discovery dialog was removed, my HR sensor is no longer listed
Thank you for all your great work!

The BT discovery dialog was removed some time ago (about a year ago). As i had my HR sensor (00002a37... GATT characteristics supported device) already set in OT, all was working even after that change. Unfortunately, after uninstall and new installation, newly installed OT does not allow me to select my device, in the Settings → Bluetooth → Heart rate → i have " * None ".

I can confirm, however, with nRF toolbox:

that my sensor works. The sensor is Amazfit Bip with HR access enabled for 3rd party software (making the sensor available as generic GATT device). So let me repeat - this device used to work normally with OT, but i have no longer change to actually select it to be able to use it.

Maybe i have missed something? Thank you very much.

dennisguse:
Some BLE devices do not announce all there supported services in the BLE advertisement packages.
Android's ScanFilter decides based upon this information only.
When connecting to such a device (and trigger service discovery), the service can still be used.

Fixes #447.


dennisguse:
Wait for testing by @vanous.

vanous:
This is awesome and works, thank you very much! I will follow up in the issue tracker.

PR #543, opened 2020-12-05
Viewpager2 zoom

dennisguse:
Almost not tested.

Approach: ChartView handles all touchEvents if
* more than one pointer is down (i.e., zooming in progress), or
* the view is in zoomed mode (and the user can scroll)
otherwise the parent may react to the events.

The following case felt weird:
1. Go to "By time"
2. Zoom
3. Scroll from right to left until moving to "By distance"
4. Swipe back to "By time"
Then the ChartView is reset and thus the user needs to zoom/scroll again.
Also: The user has no indication to see when he might swipe to the next tab as s/he doesn't 
know how much is left on the x-axis.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/chart/ChartView.java
+++ b/src/main/java/de/dennisguse/opentracks/chart/ChartView.java
@@ -115,18 +115,10 @@ public class ChartView extends View {
     private int effectiveWidth = 0;
     private int effectiveHeight = 0;
 
-    private boolean chartByDistance;
+    private final boolean chartByDistance;
     private boolean metricUnits = true;
     private boolean reportSpeed = true;
     private boolean showPointer = false;
-
-    // To handle disallow intercept touch event in parent.
-    private static final int SCROLL_X_NONE = 0;
-    private static final int SCROLL_X_LEFT = 1;
-    private static final int SCROLL_X_RIGHT = 2;
-    private boolean isOnLeftEdge = true;
-    private boolean isOnRightEdge = true;
-    private int scrollXDirection = SCROLL_X_NONE;
 
     private final GestureDetectorCompat detectorScrollFlingTab = new GestureDetectorCompat(getContext(), new GestureDetector.SimpleOnGestureListener() {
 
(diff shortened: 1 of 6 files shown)


src/main/java/de/dennisguse/opentracks/chart/ChartView.java
```diff
@@ -527,6 +542,10 @@ public void computeScroll() {
     public boolean onTouchEvent(MotionEvent event) {
         boolean isZoom = detectorZoom.onTouchEvent(event);
         boolean isScrollTab = detectorScrollFlingTab.onTouchEvent(event);
+
+        // ChartView handles zoom gestures (more than one pointer) and all gestures when zoomed itself
+        requestDisallowInterceptTouchEventInParent(event.getPointerCount() != 1 || zoomLevel != MIN_ZOOM_LEVEL);
```
rgmf, line 547: In that way only can swip if there are not zoom.

If you prefer that behaviour, what would you think add double tap to zoom out?

rgmf (commented):
> The following case felt weird:
> 
>     1. Go to "By time"
> 
>     2. Zoom
> 
>     3. Scroll from right to left until moving to "By distance"
> 
>     4. Swipe back to "By time"
>        Then the ChartView is reset and thus the user needs to zoom/scroll again.
>        Also: The user has no indication to see when he might swipe to the next tab as s/he doesn't
>        know how much is left on the x-axis.

Yes, but I think that is another problem.

I'd add ViewPager2 with swip feature and then I'd fix that weird case.

rgmf (approved):

src/main/java/de/dennisguse/opentracks/chart/ChartView.java
dennisguse: Why not :)

dennisguse:
@rgmf You can merge this, if you like it.
If you like your approach more than close it.
I am not really using the zoom-feature, so I should not complain too loud :)
Was just an idea.

rgmf:
So, we are going to merge it. Your changes become code simplier and I don't so sure about double tap feature because it would be a little bit hidden feature.

Issue #545, opened 2020-12-06
Use ThumbnailUtils for marker's photos


rgmf:
**Is your feature request related to a problem? Please describe.**
Marker's list has trouble when there are several large photos (slow loading).

**Describe the solution you'd like**
Use ThumbnailUtils (see https://github.com/OpenTracksApp/OpenTracks/pull/519#issuecomment-729231165)
https://developer.android.com/reference/android/media/ThumbnailUtils

rgmf:
Seem that using ThumbnailUtils.extractThumbnail can be problematic because you are actually loading in memory a copy of the large bitmap, so this is not a good way of managing big images.

In this official doc explain how to do it:
https://developer.android.com/topic/performance/graphics/load-bitmap?hl=es

rgmf:
Seem that the problem comes from the moment where image is loaded (from storage). The size of the image is not the problem (altought it doesn't help).

Cache strategy could be the solution: https://developer.android.com/topic/performance/graphics/cache-bitmap

dennisguse:
We could also create thumbnails and store those beside the actual pictures.

rgmf:
> We could also create thumbnails and store those beside the actual pictures.

Yes, I thought on that possibility but the actual issue is that we have to load photo's file on a separate thread. Doesn't matter if photo is large or small in size: the marker's list is really slow if images are not loaded in a separate thread.

So, I think the best approach if what I've proposed in #546

PR #546, opened 2020-12-08
Marker's list with photos are now loaded efficiently.

rgmf:
**Describe the pull request**
Photos from marker's list are loaded asynchronously and Adapter uses a cache strategy to speed up the loading. It's using 1/8 from free memory (read code).

**Link to the the issue**
#545 

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/MarkerListActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/MarkerListActivity.java
@@ -17,7 +17,6 @@
 package de.dennisguse.opentracks;
 
 import android.app.SearchManager;
-import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.SharedPreferences.OnSharedPreferenceChangeListener;
@@ -30,11 +29,12 @@ import android.view.View;
 
 import androidx.annotation.NonNull;
 import androidx.appcompat.widget.SearchView;
-import androidx.cursoradapter.widget.ResourceCursorAdapter;
 import androidx.loader.app.LoaderManager;
 import androidx.loader.content.CursorLoader;
 import androidx.loader.content.Loader;
 
+import de.dennisguse.opentracks.adapters.MarkerResourceCursorAdapter;
+import de.dennisguse.opentracks.adapters.ScrollVisibleViews;
 import de.dennisguse.opentracks.content.data.Marker;
 import de.dennisguse.opentracks.content.data.MarkerColumns;
 import de.dennisguse.opentracks.content.data.Track;
@@ -44,8 +44,6 @@ import de.dennisguse.opentracks.fragments.DeleteMarkerDialogFragment;
 import de.dennisguse.opentracks.fragments.DeleteMarkerDialogFragment.DeleteMarkerCaller;
 import de.dennisguse.opentracks.util.ActivityUtils;
 import de.dennisguse.opentracks.util.IntentUtils;
-import de.dennisguse.opentracks.util.ListItemUtils;
-import de.dennisguse.opentracks.util.MarkerUtils;
 import de.dennisguse.opentracks.util.PreferencesUtils;
 
 /**
@@ -81,7 +79,7 @@ public class MarkerListActivity extends AbstractActivity implements DeleteMarker
         }
     };
     private Track track;
-    private ResourceCursorAdapter resourceCursorAdapter;
+    private MarkerResourceCursorAdapter resourceCursorAdapter;
 
     private MarkerListBinding viewBinding;
 
@@ -123,33 +121,17 @@ public class MarkerListActivity extends AbstractActivity implements DeleteMarker
 
         viewBinding.markerList.setEmptyView(viewBinding.markerListEmpty);
         viewBinding.markerList.setOnItemClickListener((parent, view, position, id) -> {
+            resourceCursorAdapter.markerInvalid(id);
             Intent intent = IntentUtils.newIntent(MarkerListActivity.this, MarkerDetailActivity.class)
                     .putExtra(MarkerDetailActivity.EXTRA_MARKER_ID, new Marker.Id(id));
             startActivity(intent);
         });
-        resourceCursorAdapter = new ResourceCursorAdapter(this, R.layout.list_item, null, 0) {
-            @Override
-            public void bindView(View view, Context context, Cursor cursor) {
-                int nameIndex = cursor.getColumnIndex(MarkerColumns.NAME);
-                int timeIndex = cursor.getColumnIndexOrThrow(MarkerColumns.TIME);
-                int categoryIndex = cursor.getColumnIndex(MarkerColumns.CATEGORY);
-                int descriptionIndex = cursor.getColumnIndex(MarkerColumns.DESCRIPTION);
-                int photoUrlIndex = cursor.getColumnIndex(MarkerColumns.PHOTOURL);
 
-                int iconId = MarkerUtils.ICON_ID;
-                String name = cursor.getString(nameIndex);
-                long time = cursor.getLong(timeIndex);
-                String category = cursor.getString(categoryIndex);
-                String description = cursor.getString(descriptionIndex);
-                String photoUrl = cursor.getString(photoUrlIndex);
-
-                ListItemUtils.setListItem(MarkerListActivity.this, view, false, true, iconId, R.string.image_marker, name, null, null, 0, time, false, category, description, photoUrl);
-            }
-        };
+        resourceCursorAdapter = new MarkerResourceCursorAdapter(this, R.layout.list_item);
+        ScrollVisibleViews scrollVisibleViews = new ScrollVisibleViews(resourceCursorAdapter);
+        viewBinding.markerList.setOnScrollListener(scrollVisibleViews);
         viewBinding.markerList.setAdapter(resourceCursorAdapter);
         ActivityUtils.configureListViewContextualMenu(viewBinding.markerList, contextualActionModeCallback);
-
-        loadData(getIntent());
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


rgmf:
@dennisguse no hurry, obviously :D

dennisguse:
Looks good.
Did some tiny changes: if you like them please squash and merge; my changes are not worth the commit ><
PS: Could not really test as I don't have too many markers with pictures (lack of test data basically).

dennisguse (approved):
@rgmf If it works for, then please squash and merge.

rgmf:
> Looks good.
> Did some tiny changes: if you like them please squash and merge; my changes are not worth the commit ><
> PS: Could not really test as I don't have too many markers with pictures (lack of test data basically).

Only a tiny thing:
I've modified the code that calculate the cache to be used. It now is calculated from actual free memory available.

imagen
(source: https://stackoverflow.com/a/18375641)

dennisguse:
Thanks - didnot know that!

rgmf:
I had tested it with about 50 markers with images I have and all work perfectly. But before merge I tested with more than a hundred of markers with images and it crashes if you scroll quickly. That's because the number of thread it creates (I think). So, I have to re-think it: another async strategy for loading images (thread pool?)

Mental note: follow the next algorithm before create a new PR:

```asm
begin:
   test
   jmp begin
```

dennisguse:
All these tiny details :)

Just looking at the code again and there might even be situations in which the multiple threads are started to generate one thumbnail.
Those might try to do the same thing in parallel and whoever finishes last will be in the `LruCache`.
ThreadPool sounds like a good idea and we should check if there is already (ongoing) work in progress for a thumbnail.

One more thing: we could wrap the Thumbnails into a WeakReference before putting them into the LruCache (or even drop the LruCache completely).
In this way the JVM will remove the thumbnails if it needs more RAM - only thing that we should handle a situation were the JVM trashes things and we try to regenerate it.

rgmf:
The bug has been catched... I don't know why but `getContentResolver().openInputStream(..)` cannot called in parallel.

If you call `openInputStream` serveral times in a time then memory usage grow up a lot (according to Android Studio Profiler).

dennisguse:
? I dont understand

rgmf:
> ? I dont understand

Me neither, but it's what is happening... for example, doing scroll more than 40 open input stream are opened and this error is triggered before any of these 40 open input stream were closed.

```
2020-12-16 18:11:02.005 1355-1555/system_process E/InputDispatcher: channel 'c12ac68 de.dennisguse.opentracks.debug/de.dennisguse.opentracks.MarkerListActivity (server)' ~ Channel is unrecoverably broken and will be disposed!
```

Also, with deprecated AsyncTask in `doInBackground` I can see that two input stream are not opened at the same time (like a sempahore was there), and this error is not triggered.

dennisguse:
Can we just usw one thread only?

rgmf:
> Can we just usw one thread only?

Yes, we can. I'll do it that.

dennisguse:
That's actually odd; just by reading the code, there should be at most 10 openinputstreams (one per thread in the executor).
And they should be closed automatically after the image was set. 
Could you check if you can close the inputstream manually?

```
InputStream i;
try {
  i = ....
} finally {
 Log.w(TAG, "inputstream closed");
 if (i != null) i.close();
}
```

And two more things:
1. the executor service must be shutdown the parent object reaches its end of the lifecycle.
2. we still may start multiple thumbnail generations for the same image in parallel (I scroll very fast from the top to the end and back). This might enqueue the same image multiple times and waste a lot of energy.

And one more thing:
What about using a LIFO for the executor service? assuming that the image most recently requested is should be loaded first?
Which actually brings up the next aspect: should be we stop generating thumbnails for images that are not visible anymore?

rgmf:
> That's actually odd; just by reading the code, there should be at most 10 openinputstreams (one per thread in the executor).

Yes, the 40 was with old version, now it opens at most 10 inputstream. But with 10 it also crashes.

> And they should be closed automatically after the image was set.
> Could you check if you can close the inputstream manually?
> 
> ```
> InputStream i;
> try {
>   i = ....
> } finally {
>  Log.w(TAG, "inputstream closed");
>  if (i != null) i.close();
> }
> ```

All inputstreams are closed but before the first inputstream is closed all 10 inputstreams already are opened.

> 
> And two more things:
> 
>     1. the executor service must be shutdown the parent object reaches its end of the lifecycle.

:+1: 

> 
>     2. we still may start multiple thumbnail generations for the same image in parallel (I scroll very fast from the top to the end and back). This might enqueue the same image multiple times and waste a lot of energy.

:+1:  this must be fixed, you are right.

> 
> And one more thing:
> What about using a LIFO for the executor service? assuming that the image most recently requested is should be loaded first?
> Which actually brings up the next aspect: should be we stop generating thumbnails for images that are not visible anymore?

:+1:

rgmf:
@dennisguse only FYI, if you called `openAllPhotos` OT crashes (I think, the code below show what I suspected: opening several inputstreams using the content resolver results in trouble - but no idea why). 

PS/ It's a simple code that I've used to be sure that we cannot open more than X inputstream at the same time.

```java
private String[] photoUrls = new String[] {
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/8/images-24.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/12/images-11.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/31/images-28.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/31/images-29.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/31/images-30.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/33/images-18.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/33/images-19.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/39/images-32.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/70/images-17.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/81/images-8.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/81/images-10.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/98/images-12.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/98/images-13.jpeg",
            "content://de.dennisguse.opentracks.debug.fileprovider/pictures/Pictures/98/images-14.jpeg"
    };

    private void openAllPhotos() {
        for (int i = 0; i < photoUrls.length; i++) {
            asyncOpenPhoto(photoUrls[i]);
        }
    }

    private void asyncOpenPhoto(String photoUrl) {
        executorService.execute(() -> {
            try (InputStream inputStream = activity.getContentResolver().openInputStream(Uri.parse(photoUrl))) {
                Bitmap bitmap = BitmapFactory.decodeStream(inputStream);
            } catch (IOException e) {
                Log.e(TAG, "Failed to image " + e);
            }
        });
    }
```

rgmf:
@dennisguse I CANNOT BELIEVE IT!!!! :angry:

I was using a smartphone and the error is also triggered in the main branch when I enter to Markers. So all the errors I could see in this branch has nothing to do with this branch but with the smartphone (Lineage 16, Sony Xperia).

No problems with any AVD and my current smartphone (Pixel 2 XL). It didn't have any sense.

dennisguse:
That hurts..... real bad.

But one more thing: why do we use the FileProvider at all (I guess the bug is in Android's FileProvider on your device)?
i mean it is our file in our space - so we can access it directly without the need to to the use FileProvider at all.
This alone might improve performance by a lot.
But we store the FileProvider URI in the photoUri column, so it might not be straight forward to convert...

rgmf:
> That hurts..... real bad.
> 
> But one more thing: why do we use the FileProvider at all (I guess the bug is in Android's FileProvider on your device)?
> i mean it is our file in our space - so we can access it directly without the need to to the use FileProvider at all.
> This alone might improve performance by a lot.
> But we store the FileProvider URI in the photoUri column, so it might not be straight forward to convert...

Well... seem that it's not that difficult. We have the name of the file (the last part of the URI) and the trackId... I'll take a look.

rgmf:
- [x] the executor service must be shutdown the parent object reaches its end of the lifecycle.
- [x] we still may start multiple thumbnail generations for the same image in parallel (I scroll very fast from the top to the end and back). This might enqueue the same image multiple times and waste a lot of energy.

dennisguse:
@rgmf About the last point (fileprovider): I guess so, as you already saw a crash on your device and there might be others having trouble with this as well.
From a functional standpoint: this list is not too essential, so if it crashes the application users can still import/export/record stuff.
So, we can also ship it and see what happens and check how many people complain // use GooglePlay reportings.

Should we put this into the upcoming release?

PS: The Marker has the trackId as an attribute.

rgmf:
> @rgmf About the last point (fileprovider): I guess so, as you already saw a crash on your device and there might be others having trouble with this as well.

My faulty device crashes with File / FileInputStream too :(

> From a functional standpoint: this list is not too essential, so if it crashes the application users can still import/export/record stuff.
> So, we can also ship it and see what happens and check how many people complain // use GooglePlay reportings.

I would leave it like it is now.

> Should we put this into the upcoming release?

Yes, I think we can add it to the upcoming release.

With my actual phone it works well with more than 150 images.

dennisguse:
@rgmf Should we include this in v3.15.0?

rgmf:
@dennisguse Yes, we can merge it. It's not as good as Glide is but at least you can move around markers with a lot of photos smoothly :smile: 

I'm going to "press the button"...

PR #549, opened 2020-12-16
ActionBar now provides back functionality

dennisguse:
The app icon is now only shown in the TrackListActivity.
In all other activities the actionBar icon now triggers the back action (same as Android navigation bar).

This feature was actually in place quite a while ago, but implemented differently.
Back then the actual target for the back action of each activity was defined in the manifest.
The new solution just uses android's activity stack to go back.


rgmf (approved):
Great for me and needed option... I didn't realize because I always use back button.

Issue #316, opened 2020-07-12
Concept: how to store pause/resume


dennisguse:
At the moment, we store the pause/resume and signal lost (aka max distance) events as TrackPoints with a specific latitude (100.0 and 200.0).
In case of pause/resume even with two TrackPoints.
On export these are used as separators for individual track segments.

However, we have trouble when exporting and importing data; especially as we try to re-create the same database values (main issue is time).
Also we may run into issues that _we_ think a coordinate is a specific one, but it was reported this way by a GPS unit (see #315 where lat: 180 was in the database).

I guess, we need a cooler concept for this kind of data:
1. How do we store this kind of data?
2. How do we export/import this?

Follow-up to #315.

pstorch:
When I read the KML spec correctly, a `<gx:MultiTrack>` can have several `<gx:Track>` elements which are used to split a tour into several segments when interrupted by a break (a.k.a. pause/resume).

dennisguse:
@pstorch The actual question is: how do we store this in OpenTracks?
Multitrack is already used for GPX/KML (I also just read the spec ><) - better late than never.

pstorch:
The "clean" solution would be to also model it as segments in the database. But introducing a 1:n relationship from Track to Segments and then another 1:n relationship to TrackPoints might break a lot.

What about adding a SegmentNumber column in the TrackPoints table? And increase the SegmentNumber on every pause/resume.

What exactly are the "Lost Signal" TrackPoints used for?

dennisguse:
@pstorch It is not really signal lost - it is actually a segment breaker.
When you receive a GPS coordinate that farther away than _max distance_ (see settings) then a Pause-TrackPoint is inserted.
I guess the actual info is for drawing the map to not a draw a straight for a part where actually no useful measurements were received.
It is also used to compute the distance.
I just called it signal lost as in case of cycling it usually means that the GPS is not really precise/available.

pstorch:
Pulling the discussion from https://github.com/OpenTracksApp/OpenTracks/pull/315#issuecomment-657267366 to here.

> Pause/Resume is restored, but the timestamps are off.
> 
> And even one more thing: all GPS-based TrackPoints use GPS time, but pause/resume use device time (also start and stop time of a track).

Why do we need the extra timestamps of the pause/resume TrackPoints? Is it much different to the last/first valid TrackPoint of a segment?

dennisguse:
[...]

PR #550, opened 2020-12-17
Remove Track.Id and Marker.Id.isValid()

dennisguse:
isValid() should not be necessary.
We can just use null as Track.Id or Marker.Id if the marker/track was not loaded from the database.

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/content/data/TestDataUtil.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/content/data/TestDataUtil.java
@@ -7,6 +7,8 @@ import android.util.Pair;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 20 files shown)


dennisguse:
Merge _after_ v3.14.1 as this might break stuff.

rgmf:
> Merge _after_ v3.14.1 as this might break stuff.

Good idea :D

I'm going to install this branch on my phone and use it. I'll inform you of any issue ;)

rgmf (approved):
I've not been got problems  for these days using this branch... Good for me.

dennisguse:
@rgmf Great; I will just add some smaller (non-relevant) cleanups and merge it then.
I stumpled upon some mess in the tests while working on #316 and will put these in here.

PR #552, opened 2020-12-19
Share markers from list

### LINKED ISSUE #495: Markers should also be shareable
We only support sharing of tracks at the moment.
There might a benefit to share a marker directly (incl. it's picture).

dennisguse:
Marker images can now be shared from the list directly via the contet menu (like tracks).

Fixes #495.


rgmf (approved):
Good for me... only one little bit thing: when you share several images the message says "I think you might be interested in this image." instead of "I think you might be interested in these images."

I don't know if this worth it or it's important ;)

dennisguse:
We do the same for tracks (aka only singular); i guess, we have some other stuff there to do as well.
For now, I guess it is sufficient ;)

Issue #479, opened 2020-10-29
Tables: set `NOT NULL`


dennisguse:
In the database layer (i.e., TracksColumn, MarkersColumn, and TrackPointsColumn) none of the columns are set to NOT NULL.
However (almost?) all these fields are mandatory and should be therefore be marked on database layer as NOT NULL.

dennisguse:
Don't know if this is really needed. Probably not.

PR #553, opened 2020-12-22
VERY EXPERIMENTAL: refactoring storage of pause and resume

dennisguse:
For be able to recreate the TrackStatistics (export/import), we need to store some additional TrackPoints (aka track started, track ended) and these need to use the same base for timestamps (here device time; NOT GPS-provided time).

* [x] pause/resume TrackPoint type is stored in separate column (DB version 30)
* [x] (to be decided) downgrading from DB version 30 to 29 restores data correctly
* [x] automatic pause TrackPoints migrated to SEGMENT_END_AUTOMATIC (attached to next trackpoints)
* [x] device-time is used (instead of GPS time), but existing data is _not_ updated
* [x] (only new tracks) add track start TrackPoint (type SEGMENT_START_MANUAL)
* [x] (only new tracks) add track start TrackPoint (type SEGMENT_END_MANUAL) on pressing `pause`/`stop`
* [x] export _all_ SEGMENT_START_MANUAL (has no coordinates)
* [x] import _all_ SEGMENT_START_MANUAL (the first location of a segment; if no coordinates are present otherwise SEGMENT_START_AUTOMATIC)
* [x] update OSMDashboard to use TrackPoint.type (as well as old interface; if not present for now)
* [x] finally: write documentation of internal data structures
* [x] TrackStatistics are now computed only using TrackPoints
* [x] Check using data set (@dennisguse) 
* [x] Export only xml structures for data (e.g., speed) if it contains not only null

Cleanup:
* [x]  `gpsStatus.onLocationChanged(trackPoint);` is called for _each_ TrackPoint

Upgrade strategy:
1. Release new version of OSMDashboard (Dashboard API changes w/ versioning)
2. Release new version containing database upgrade/downgrade code only (first commit in this branch; version v3.14.2)
3. Release new version with all changes (version v3.15.0)
    Add in release notes:
    * that this upgrade may be harmful and tracks should be exported first
    * that OSMDashboard needs to be upgraded to v2.9.0


dennisguse:
@rgmf @pstorch 
Happy new year ;)

If you have some time (and data to lose), here is a first version to change the internal storage of TrackPoints, so that we can restore the TrackStatistics completely (export/import).
So, if you tests this using a real database, please get a backup first.
Hope nothing burns....

PS: Downgrading from DB version 30 to version 29 _does not restore_ deleted RESUME trackpoints.
So, upgrade->downgrade will result in changed data.
Don't know if this is really worth doing it.

Anyhow: :crossed_fingers:, but keep a :fire_extinguisher: handy :sunglasses:

pstorch:
I think the update of the trackpoints table doesn't work. The latitude column is an integer not a double, so the where clause with `100.0` or `200.0` doesn't find anything.
When I look into the SQLite DB I see the type is still 0:

Screenshot from 2021-01-02 18-53-22
Screenshot from 2021-01-02 18-53-45

dennisguse:
@pstorch Thx; one of those face palm moments ;)

src/main/java/de/dennisguse/opentracks/content/provider/CustomSQLiteOpenHelper.java
```diff
@@ -345,8 +345,8 @@ private void upgradeFrom29to30(SQLiteDatabase db) {
 
         // TrackPoints
         db.execSQL("ALTER TABLE trackpoints ADD COLUMN type TEXT CHECK(type IN (-2, -1, 0, 1))");
-        db.execSQL("UPDATE trackpoints SET type = -2, latitude = NULL, longitude = NULL WHERE latitude = 200.0");
-        db.execSQL("UPDATE trackpoints SET type = 1, latitude = NULL, longitude = NULL WHERE latitude = 100.0");
+        db.execSQL("UPDATE trackpoints SET type = -2, latitude = NULL, longitude = NULL WHERE latitude = 200");
```
pstorch, line 348: That's not gonna work either. It should be 200 * 1E6 :wink:

pstorch:
> PS: Downgrading from DB version 30 to version 29 _does not restore_ deleted RESUME trackpoints.

I still don't get that "Downgrading". In which situation does this work? When I install the version from main branch (DB version 29), then the downgrade code from 30 to 29 is missing and the App is crashing. When I install the version from this branch it knows DB version 30 and doesn't downgrade either.

src/main/java/de/dennisguse/opentracks/content/provider/CustomSQLiteOpenHelper.java
dennisguse: Grummel ;)

dennisguse:
@pstorch Valid point, for a downgrade we would need to release a new version with version=29 that has the downgrade code for version 30. Then we could allow a real downgrade for a latter version.
So, technically we never allowed downgrades (with db changes) so far...
Thanks for pointing that out! Did not occur to me until know.

pstorch:
The db migration to version 30 works now.

I did some tests with my 318 tracks in my OpenTracks (debug).

Before I started:

image

- 246 trackpoints with latitude 200
- 248 trackpoints with latitude 100
- 738161 "normal" trackpoints

Then I stumbled upon the migration bug, so I started over with the old version reimporting the 318 tracks in a clean db. Interestingly it was a bit different now:

image

- 219 trackpoints with latitude 200
- 219 trackpoints with latitude 100
- 738156 "normal" trackpoints

So I guess this export/reimport "cleaned" something up. 

I installed the pauseResume branch and let the database migrate.

After that I got the same overall statistics:

image

and the trackpoints looked good:

- 219 of type 1
- 219 of type -2
- 738156 "normal" trackpoints.

I exported the 318 track after the migration.

Then I cleaned up the data and started over with importing the 318 files before the migration:

image

Now I got:

- 551 of type -1
- 737605 of type 0

:thinking: what happened here?

The last test was reimporting the migrated files back to a clean database.

Looks like something bad happening here:

image

- 219 of type -2
- 219 of type 1
- 46127 of type 0

The pause/resume trackpoints match the initial values, but why are so many regular trackpoints lost?

pstorch:
I've started a similar branch on the OSMDashboard https://github.com/OpenTracksApp/OSMDashboard/tree/pauseResume to read this new column.

Unfortunately the ContentProvider PROJECTION isn't backward compatible with older OT versions. As soon as I add the new `type` column I don't get any trackpoints from the old OT anymore. So I had to set the PROJECTION to `null` for the time being.

We could add a `version` attribute into the API (as an Intent extra), so that the OSMDashboard can decide which PROJECTION to use.

dennisguse:
@pstorch Thanks for the test.
The migration does not update the trackstatistics (nor deletes tracks), so the migration in itself is only visible in the number of trackpoints and thus the aggregated stats will remain the same.
But there is an issue - I guess in the exporter; let me check.

About OSMDashboard integration: do you need to send a projection (i.e., non null)?
If yes, we can add a version attribute to the URI - no problem.

pstorch:
> About OSMDashboard integration: do you need to send a projection (i.e., non null)?

No, it's not needed. It works with null. The only drawback is that then all columns are transferred, more than needed. Don't know if this has a performance/memory impact.

dennisguse:
@pstorch Adding a version as extra sounds good!

About the data import->export->import: 
I checked the exporter and it looks fine.
And I am not really sure, how to proceed as in my simple test cases it is working as expected and with the original data, the imports look fine.
So, it must be some weird "feature" of export+import together and I am not sure, how to figure that one out.
Can you provide some test data?

pstorch:
Very strange: I tried to reproduce the last testcase (reimporting the migrated and exported tracks). Now it worked and the aggregated stats and the db trackpoints had the correct values.

Sorry for the confusion.

rgmf:
I get this error in 40 of my 161 tracks. I've not checked all the 40 files but the error is triggered in the files where there is a pause/restart:
```
Parser error: org.xml.sax.SAXException: Parsing error at line: 4380 column: 0.
Invalid location detected: time=1583051351000 (type=SEGMENT_START_AUTOMATIC(-1)): lat=100.0 lng=0.0
```

Here is the line when the error is triggered (with the context):
```
<trkpt  lat="100" lon="0">
<time>2020-03-01T08:29:11Z</time>
</trkpt>
<trkpt  lat="200" lon="0">
<time>2020-03-01T08:33:49.166Z</time>
</trkpt>
```

To reproduce:
1. Install this branch.
2. Export all tracks to GPX format.
3. Import all tracks.
4. Get this error.

Also, when you are recording then the intervals doesn't work. Maybe it needs to be refactored because intervals uses TrackPoints directly.

dennisguse:
@pstorch If you can reproduce it again (or some other weird behavior), let me know.

@rgmf Those trackpoints should not actually not have been exported - they should have become segment open/close. We can now either ignore these - or try to restore the segment. But those trackpoints have device time while the rest has GPS time.

rgmf:
> @rgmf Those trackpoints should not actually not have been exported - they should have become segment open/close. We can now either ignore these - or try to restore the segment. But those trackpoints have device time while the rest has GPS time.

I see.

I guess it's not that important for now but I'd try to restore segments altough stats are not the same: better that than end up with errors.

I'll test import/export from this branch then ;)

dennisguse:
@rgmf the old pause/resume markers are now transformed into a new segment.
I hope we can remove this workaround by the end of the year.
Makes the import code somehow ugly, but backward compatibility should be  a real thing.

@pstorch I added a version field to the Dashboard-Intent.
Does this work for you?

pstorch:
I'll check

pstorch:
@dennisguse it works, but I'm thinking if we should use a simple integer instead of a String. If (in the future) we have even more versions, we can check with a comparison like `protocolVersion > x`.

pstorch:
@dennisguse works like a charm with the integer protocol version :+1:

dennisguse:
@pstorch that is a really great improvement.
Will save us some hazzle in the future - thanks!

dennisguse:
@pstorch @rgmf I did a testing hike today and I saw that moving time was not working (fixed just now).
So, from my side this PR is ready! 
Can you give this another spin and break it? ><

pstorch:
I'll so some tests tomorrow

rgmf:
@dennisguse I've tested the app this morning and I got this error (crash OT) when OT was pause and I was switching between STATS and INTERVALS tabs:

```
2021-01-17 10:00:55.307 22824-22824/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.dennisguse.opentracks.debug, PID: 22824
    java.lang.NullPointerException: Attempt to invoke virtual method 'double java.lang.Double.doubleValue()' on a null object reference
        at de.dennisguse.opentracks.content.data.TrackPoint.getAltitude(TrackPoint.java:254)
        at de.dennisguse.opentracks.fragments.StatisticsRecordingFragment.setLocationValues(StatisticsRecordingFragment.java:435)
        at de.dennisguse.opentracks.fragments.StatisticsRecordingFragment.updateUI(StatisticsRecordingFragment.java:259)
        at de.dennisguse.opentracks.fragments.StatisticsRecordingFragment.lambda$onTrackUpdated$3$StatisticsRecordingFragment(StatisticsRecordingFragment.java:164)
        at de.dennisguse.opentracks.fragments.-$$Lambda$StatisticsRecordingFragment$qTCJqbKQ9ik7N-JSKTl9FPND1dg.run(Unknown Source:4)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6718)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:491)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

dennisguse:
@rgmf That is odd as there is a null check +hasAltitude check.
https://github.com/OpenTracksApp/OpenTracks/blob/4714171e2163e818bc486967509f585186b4bc44/src/main/java/de/dennisguse/opentracks/fragments/StatisticsRecordingFragment.java#L445
Is this deterministic (I could not reproduce it)?

pstorch:
From my 318 Tracks I got 1 import error while reimporting tracks, which where exported from before refactoring:

Screenshot_20210117-121714_OpenTracks (Debug)

I you like, I can send the problematic track to you via email.

pstorch:
Reimporting the migrated tracks is even worse:

Screenshot_20210117-122416_OpenTracks (Debug)
Screenshot_20210117-122410_OpenTracks (Debug)

dennisguse:
@pstorch I think I got and by guessing created  a test.
Don't know exactly why, but the tracks does not seem to have data attached.

rgmf:
> Is this deterministic (I could not reproduce it)?

I've not been able to reproduce it again and like you point out that's not make sense... some error updating OT from Android Studio or something weird ¿?

I don't get any error importing my 149 tracks which were exported from before refactoring. But, aggregated stats has huge differences. Some examples:

**Mountain Biking**
| Actual Stats | Stats after importing in this branch |
---|----
| Total distance: 875.31km | Total distance: 522.23km |
| Total time: 50:19:50 | Total time: 28:00:16 |

**Trail Running**
| Actual Stats | Stats after importing in this branch |
---|----
| Total distance: 5.77km | Total distance: 0km |
| Total time: 28:55 | Total time: 00:00 |

pstorch:
After importing exports from before the change and opening the AggregatedStats I get this error:

```
2021-01-17 17:34:38.531 32504-19818/de.dennisguse.opentracks.debug E/AndroidRuntime: FATAL EXCEPTION: Thread-6
    Process: de.dennisguse.opentracks.debug, PID: 32504
    java.lang.NullPointerException: Attempt to read from field 'long java.time.Instant.seconds' on a null object reference
        at java.time.Instant.compareTo(Instant.java:1249)
        at java.time.Instant.isBefore(Instant.java:1279)
        at de.dennisguse.opentracks.stats.TrackStatistics.merge(TrackStatistics.java:86)
        at de.dennisguse.opentracks.viewmodels.AggregatedStatistics$AggregatedStatistic.add(AggregatedStatistics.java:80)
        at de.dennisguse.opentracks.viewmodels.AggregatedStatistics.aggregate(AggregatedStatistics.java:39)
        at de.dennisguse.opentracks.viewmodels.AggregatedStatistics.<init>(AggregatedStatistics.java:23)
        at de.dennisguse.opentracks.viewmodels.AggregatedStatisticsModel.lambda$loadAggregatedStats$0$AggregatedStatisticsModel(AggregatedStatisticsModel.java:36)
```

pstorch:
There is no NullPointerException anymore at importing, but as desribed at Aggregated Stats.

dennisguse:
@pstorch Fits to the no data hypothesis; so on import, we need to prevent importing tracks that do not contain any data, because then we cannot compute the TrackStatistics at all.
And we need to fix #479.

pstorch:
Ah, now the empty file is skipped: 

Screenshot_20210117-185547_OpenTracks (Debug)

:+1:

pstorch:
There is still something wrong. I reimported the tracks I've exported after upgrading the app. So the ones which got migrated during app upgrade.

Screenshot_20210117-210410_OpenTracks (Debug)

dennisguse:
@pstorch I found (and fixed) an issue happening while computing the TrackStatistics.
A SEGMENT_START_AUTOMATIC trackpoint resulted in ignoring the statistics of the previous segment (the merging of the segments was only done on SEGMENT_END_MANUAL).

The other export/import issues, I could not reproduce.
Looks okayish to me ;)

pstorch:
I tried to import the migrated dataset again and get a lot of error `Track did not contain any locations."

But as you see in the debugger, there are 2051 TrackPoints:
image

pstorch:
I think there was an error with the previous export. The files contain MultiTrack elements without Track child elements.
I have to do the export again.

pstorch:
No the tracks are exported and imported correctly again. I think it looks good now.

pstorch (approved):

dennisguse:
I missed to include the elevation_loss column in the db upgrade/downgrade for version 29/30.

dennisguse:
I fixed this and:
* released v3.14.3
* removed the tag for v3.14.2 (so F-Droid does not build it); see #576
* merged this branch locally

PR #591, opened 2021-02-07
GPX: only export TrackPoints with lat/lng.

dennisguse:
As GPX does not support exporting trackpoints without coordinates, we cannot export manually created trackpoints like (START and END).
This PR changes the GPX import/export to do not export these trackpoints (also excl. there timestamp).

This also affects segmentation on import: all segments are considerd AUTOMATIC.
Thus, restoring TrackStatistics if a track was  paused is totally off (if one moved while track was paused).


src/main/java/de/dennisguse/opentracks/io/file/TrackFileFormat.java
```diff
@@ -64,16 +62,12 @@ public String getExtension() {
...
 
         @Override
-        public TrackWriter newTrackWriter(Context context) {
-            return new KmlTrackWriter(context, false, false, exportPhotos);
-        }
-
-        public TrackExporter newTrackExporter(Context context) {
-            return newKmzTrackExporter(context, this.newTrackWriter(context), exportPhotos);
+        public TrackExporter createTrackExporter(Context context) {
+            KMLTrackExporter exporter = new KMLTrackExporter(context, false, false, exportPhotos);
```
rgmf, line 69: You use KML for KMLTrackExporter and GPX for GPXTrackExporter but Kmz for KmzTrackExporter.
When you use KmzTrackExporter, do you wanted to use KMZTrackExporter?

dennisguse: Yes, as the KMLTrackExporter is used to generate the doc.xml for the KMZ.
It is in the following line passed to a KmzTrackExporter, which uses it to create the file and then zipping everything together.

rgmf (approved):
Great refactoring and coding!
It works for me...

src/main/java/de/dennisguse/opentracks/io/file/exporter/GPXTrackExporter.java
```diff
@@ -62,28 +75,119 @@
...
     @Override
+    public boolean writeTrack(Track track, @NonNull OutputStream outputStream) {
+        try {
+            prepare(outputStream);
+            writeHeader(track);
+
+            writeMarkers(track);
+
+            Instant startTime = track.getTrackStatistics().getStartTime();
+            Duration offset = Duration.between(track.getTrackStatistics().getStartTime(), startTime);
```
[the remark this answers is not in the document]
dennisguse: Nice catch :)
As GPX only can export one track, it does not make any sense.
Even for KML I was wondering what this is (still in there).
Will check this again for multiple tracks per file export/import.

PR #598, opened 2021-02-13
Use distance from cycling speed sensors.

dennisguse:
For cycling, we already support wheel rotation (aka speed) sensors.
These can also be used to distance (actually easier than computing the speed).

This PR contains:
* [x] distance derive distance between two sensor measurements (like speed)
* [x] store distance in an extra column (sensor_distance)
* [x] adapt trackStatistics to either use sensor_distance or distance from previous trackpoint (old behavior)
* [x] export distance to KML/KMZ
* [x] (kind ugly) we need to store the distance between two  trackpoints somehow (SensorData only supports between two sensor measurements).
  Actually, we need some integration (e.g., averaging?) of sensor measurements as so far we only store the last provided value; see https://github.com/OpenTracksApp/OpenTracks/pull/592#issuecomment-775488089
* [x] check Garmin Speed 2
* [x] extend trackStatisticsUpdater
* [ ] do some actual testing

### DIFF
--- a/README_TESTED_SENSORS.md
+++ b/README_TESTED_SENSORS.md
@@ -19,6 +19,9 @@ However, often only one value is provided.
 ### Speed
 
 * Garmin Speed Sensor 2
+  This sensor reports speed data as cadence.
+  A workaround is in place.
+
 * Wahoo Speed (Model: WFRPMSPD)
 
 ### Cadence
--- a/README_TESTED_SENSORS.md
+++ b/README_TESTED_SENSORS.md
@@ -19,8 +19,11 @@ However, often only one value is provided.
 ### Speed
 
 * Garmin Speed Sensor 2
-  This sensor reports speed data as cadence.
-  A workaround is in place.
+  * Has updatable firmware that requires an account for garmin.com
+  * This sensor reports speed data as cadence.
+    A workaround is in place.
+  * This sensor sometimes counts backwards (unknown software version).
+    A workaround is in place.
 
 * Wahoo Speed (Model: WFRPMSPD)
 
--- a/doc/opentracks-schema-1.0.xsd
+++ b/doc/opentracks-schema-1.0.xsd
@@ -1,8 +1,32 @@
 <?xml version="1.0"?>
-<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+<xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema"
     targetNamespace="http://opentracksapp.com/xmlschemas/v1">
 
-    <xs:element name="trackuuid"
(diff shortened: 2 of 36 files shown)


dennisguse:
@ChristophWurst You can try this if you want (it should also work indoors).
The distance is reported between to sensor measurements, so it will be completely off - but it would be interesting if the rest is already working. I don't have any test equipment here - I "just" implemented it. ;)

I have not (yet) a good concept on how to gather the distance correctly, but it is not rocket science and we have another related issue with the other sensor data (we only store the last one, but we should do some averaging or so).
I guess enhancing SensorDataSet will be a good way...

ChristophWurst:
> @ChristophWurst You can try this if you want (it should also work indoors).

Will test soonish :pray:

dennisguse:
@ChristophWurst Implementation should be complete, but (due to lack of sensors) completely untested.
If you have some time to test it, it would be great.
Also works in an indoor situation (except that is only once stored at the end if no GPS signal is available).

PS: I, finally, ordered one today :sunglasses:

dennisguse:
@ChristophWurst This is good to go (from my side) - would be great to see if it also works for you.

ChristophWurst:
I put the Garmin Speed Sensor 2 onto my front wheel, connected it in the settings and gave it a manual spin.

The speed and distance is negative for me:

<details>

ot

</details>

Wheel spun backwards:

<details>

ot2

</details>

Summary:

<details>

ot3

</details>

ChristophWurst:
The second screenshot is from running the wheel backwards. It always reported negative numbers. I don't think this sensor has a right or wrong direction.

dennisguse:
@%#%#%&@ why can those sensor manufacturers not just build something according to the *standard*. Wahoo sends cadence as speed. Garmin sends cadence as speed. And Garmin sometimes counts backward. 

The hacky workaround is just to compute the absolute value of the wheel rotation difference.
But still.
Standards, right?

dennisguse:
@ChristophWurst I added a workaround that hopefully works.
But can you check the firmware on your sensor?
I just found that you actually need other Garmin equipment to do this; seemingly the mobile cannot do this :man_shrugging: 
https://support.garmin.com/de-DE/?partNumber=010-12843-00&tab=software

ChristophWurst:
Well, tried that. I'm generally trying to avoid any proprietary services with mandatory sign-up, hence I didn't have the Connect app before. In any case it fails pairing the sensor, while Android successfully lists it in the bluetooth settings. I can't tell what firmware I have. It must be the one that came with the device when I bought it.

dennisguse:
@ChristophWurst Does the workaround work for you?
(I would like to merge this soonish as it blocks some refactoring in the importer code).

Issue #601, opened 2021-02-14
Normalize sensors stats (avg)


rgmf:
**Is your feature request related to a problem? Please describe.**
Following with https://github.com/OpenTracksApp/OpenTracks/pull/592 the sensors stats averages has to be normalized.

**Describe the solution you'd like**
The solutions are in the follow comments:
https://github.com/OpenTracksApp/OpenTracks/pull/592#issuecomment-775488089
https://github.com/OpenTracksApp/OpenTracks/pull/592#issuecomment-778795768

rgmf:
I take notes about the progress of this issue:

I've been playing with sqlite and the possibility of normalize the avg. This SQL sentence is the nearly I have but it's difficult because the several types of the trackpoints.

I don't know if it worths or it's better doing the normalization in java code instead of sql.

```sql
select avg(x.sensor_heartrate * (x.time - coalesce(y.time, x.time))) / 1000 from (
       select w.time, max(d.time) previous, w.sensor_heartrate
       from trackpoints w left join trackpoints d on d.time < w.time
       where w.type=0
       group by w.time) x
left join trackpoints y on y.time=previous
where y.type=0
order by x.time
```

Basically what this SQL sentence does is ponderate the heart rate based on the time this heart rate last.

dennisguse:
(about code vs SQL)
Don't know; both work. Back in 2012 I would have said the DB might be faster, but nowadays :man_shrugging: 

Tiny fixe to the SQL:
* ORDER BY  x._id 

Now comes the ugly part.
In theory, all trackpoints may have sensor data (even SEGMENT_MANUAL_START, e.g., PAUSE/RESUME case). 
I guess the only exception is the very first SEGMENT_MANUAL_START as the sensor connection is established afterwards.
So, we have the following cases:
* first trackpoint of a track (ignore likely has no data)
* first tackpoint in segment (SEGMENT_AUTOMATIC_START) time difference to previous trackpoint as it is just for segment splitting (i.e., rendering and distance computation if not present via sensor).
* all other trackpoints: use.

And another ugly one: what happens if not _all_ trackpoints have sensor data (e.g., sensor got disconnected; i.e., sensor data null).

(side note)
SQL (for me) is easier to read if SQL keywords are upcase.
Then it also works without code highlighting (for example as String in Java) ;)

dennisguse:
[...]

PR #606, opened 2021-02-21
Normalize hr, cadence and power averages

rgmf:
**Describe the pull request**
It computes time weighted heart rate, cadence and power averages. So the more time it takes the more important is.
Also, it computes heart rate and cadence maximum.
Finally, it ignores pause time for whatever reason (SEGMENT_START_MANUAL, SEGMENT_START_AUTOMATIC).

For a track with about 7000 trackpints the query takes about 140ms.

**Link to the the issue**
#601 

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/content/data/TrackPointsColumns.java
+++ b/src/main/java/de/dennisguse/opentracks/content/data/TrackPointsColumns.java
@@ -54,6 +54,13 @@ public interface TrackPointsColumns extends BaseColumns {
     String ELEVATION_GAIN = "elevation_gain";
     String ELEVATION_LOSS = "elevation_loss";
 
+    // Alias for sensor statistics
+    String ALIAS_AVG_HR = "avg_hr";
+    String ALIAS_MAX_HR = "max_hr";
+    String ALIAS_AVG_CADENCE = "avg_cadence";
+    String ALIAS_MAX_CADENCE = "max_cadence";
+    String ALIAS_AVG_POWER = "avg_power";
+
     String CREATE_TABLE = "CREATE TABLE " + TABLE_NAME + " ("
             + _ID + " INTEGER PRIMARY KEY AUTOINCREMENT, "
             + TRACKID + " INTEGER NOT NULL, "
--- a/src/main/java/de/dennisguse/opentracks/content/provider/CustomContentProvider.java
+++ b/src/main/java/de/dennisguse/opentracks/content/provider/CustomContentProvider.java
@@ -36,6 +36,7 @@ import androidx.annotation.VisibleForTesting;
 import java.util.Arrays;
 
 import de.dennisguse.opentracks.content.data.MarkerColumns;
+import de.dennisguse.opentracks.content.data.TrackPoint;
 import de.dennisguse.opentracks.content.data.TrackPointsColumns;
 import de.dennisguse.opentracks.content.data.TracksColumns;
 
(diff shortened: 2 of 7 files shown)


rgmf:
@dennisguse if you think it's okay then I'll write tests to ensure all this is working well (we cannot trust on manual tests for this).

src/main/java/de/dennisguse/opentracks/content/data/TrackPointsColumns.java
```diff
@@ -31,6 +31,7 @@
     String TABLE_NAME = "trackpoints";
     Uri CONTENT_URI_BY_ID = Uri.parse(ContentProviderUtils.CONTENT_BASE_URI + "/" + TABLE_NAME);
     Uri CONTENT_URI_BY_TRACKID = Uri.parse(ContentProviderUtils.CONTENT_BASE_URI + "/" + TABLE_NAME + "/trackid");
+    Uri CONTENT_URI_SENSOR_STATS = Uri.parse(ContentProviderUtils.CONTENT_BASE_URI + "/" + TABLE_NAME + "/sensorstats");
```
dennisguse, line 34: I would consider this computation rather part of the track than of the trackpoints.
Technically, you are completely right as the data is derived from the trackpoints.
However, in the track-table, we have all the other statistics - only difference is that those are precomputed.

src/main/java/de/dennisguse/opentracks/content/provider/CustomContentProvider.java
```diff
@@ -356,11 +362,46 @@ private Uri insertMarker(Uri url, ContentValues contentValues) {
         throw new SQLException("Failed to insert a marker " + url);
     }
 
+    /**
+     * Return the string representing the query that compute sensor stats from trackpoints table.
```
dennisguse, line 366: * Return the string representing the query that compute sensor stats from trackpoints table.
* It computes the average for heart rate, cadence, and power (duration-based average) and the maximum for heart rate and cadence.

src/main/java/de/dennisguse/opentracks/content/provider/CustomContentProvider.java
```diff
@@ -356,11 +362,46 @@ private Uri insertMarker(Uri url, ContentValues contentValues) {
...
+     * It computes time weighted heart rate, cadence and power averages. So the more time it takes the more important is.
+     * Also, it computes heart rate and cadence maximum.
+     * Finally, it ignores pause time for whatever reason (SEGMENT_START_MANUAL, SEGMENT_START_AUTOMATIC).
```
dennisguse, line 369: I would argue that a START_AUTOMATIC should be considered as it only denotes a gap in the GPS track (due to fast movement, low reception etc).
Example: 
If I use a cycling sensor (distance/speed), this sensor data is stored for all trackpoints of type TRACKPOINT, START_AUTOMATIC, and (I guess even) END_MANUAL. 
If I am cycling in the mountains, GPS reception is likely to be varying a lot and this will likely result in START_AUTOMATIC (like going downhill with a racing bike at 60km/h).
And in my opinion, these should be part of the AVG/MAX computation.

In data:
t=0, START_MANUAL
t=10, TRACKPOINT, speed=5
t=20, TRACKPOINT, speed=10
t=25, START_AUTOMATIC speed=60
t=27, START_AUTOMATIC speed=60
t=30, TRACKPOINT, speed=15
t=30, END_MANUAL, speed=15

I was thinking about this while working on the "distance from sensor" task and for the overall sum to match, we need to consider all trackpoints that have distance (relative to previous).
So, basically everything except START_MANUAL, which should not have any sensor data (as no connection to sensor is yet established) and it has not duration to previous trackpoint.

dennisguse (changes requested):
Code looks good; some comments.
(almost good to go!)

src/main/java/de/dennisguse/opentracks/content/provider/CustomContentProvider.java
rgmf: Umm, okay, I understand. I was thinking on those moments when you take a rest in your route and forget pause the activity. I guess we cannot do anything here.

So, yes, you are right.

src/main/java/de/dennisguse/opentracks/content/provider/CustomContentProvider.java
```diff
@@ -356,13 +362,47 @@ private Uri insertMarker(Uri url, ContentValues contentValues) {
         throw new SQLException("Failed to insert a marker " + url);
     }
 
+    /**
+     * Return the string representing the query that compute sensor stats from trackpoints table.
+     * It computes the average for heart rate, cadence, and power (duration-based average) and the maximum for heart rate and cadence.
+     * Finally, it ignores manual pause (SEGMENT_START_MANUAL).
+     */
+    private final String getSensorStatsQuery() {
```
dennisguse, line 370: Can be a property aka private final String SENSOR_STATS_QUERY = ""

dennisguse (approved):
Code looks good!

src/main/java/de/dennisguse/opentracks/content/provider/CustomContentProvider.java
rgmf: Oops!!! :see_no_evil: I read it quickly... Fixed!

rgmf:
Ready to go...

I've added tests to check query is computing averages well.

dennisguse (approved):
Looks good!
@rgmf Merge at your discretion.

PR #618, opened 2021-02-28
Performance: statistics fragment use sharedPreferenceListener

dennisguse:
While doing some cleanup, I was annoyed by the slowness of the statistics fragments.
Those repeatedly access  the sharedpreferences in a very non-performant manner.
Now, a listener is used and startup is way faster + the unwanted animation is gone.


rgmf (approved):
What a stupid thing I did with Preferences in fragment statistics!
Great catch and cleanup!!!

dennisguse:
@rgmf Don't worry, wasn't you - the initial code was like this.
And the two of us just copied it and re-used it :)

rgmf:
@dennisguse I didn't realize that pace of all activities are showed in min/km regardless the activity type. But if you move off from OT and back again then OT shows pace correctly.

dennisguse:
@rgmf You are right: the callback from the activityChooser is handled by the TrackRecordingActivity, but this info is _not_ passed to the StatisticsRecordingFragment. We now update it if the track's category changes (which is triggered by the user's selection).

rgmf:
> @rgmf You are right: the callback from the activityChooser is handled by the TrackRecordingActivity, but this info is _not_ passed to the StatisticsRecordingFragment. We now update it if the track's category changes (which is triggered by the user's selection).

Also for TrackRecordedActivity / StatisticsRecordedFragment...

dennisguse:
@rgmf Done;
Also removed use of onActivityCreated (deprecated) and reordered the methods.

rgmf (approved):
It works for me.
PS/ Mental note: I need to pay attention to declare final the variables has to be final.

dennisguse:
@rgmf I use AndroidStudio's static code analyzer every now and then - it shows such things and even fixes it almost automatically.
:)

rgmf:
> @rgmf I use AndroidStudio's static code analyzer every now and then - it shows such things and even fixes it almost automatically.
> :)

Ummm, okay, I'm going to add that in my routine :smile:

dennisguse:
@rgmf And at some point I would love to get rid of all the warnings AndroidStudio shows, but somehow they are not getting less :shrug:

rgmf:
> @rgmf And at some point I would love to get rid of all the warnings AndroidStudio shows, but somehow they are not getting less shrug

:rofl: :rofl: :rofl: 

I'm going to try to help you on that.

PR #643, opened 2021-03-11
Performance: Keep reference to sharedPreferences.

### LINKED ISSUE #642: Keep reference to sharedpreference if used more than once
We are opening the sharedpreferences again and again.
This is really slow and costly.

We need to replace this everywhere where we call something like `PreferenceUtils.isXXX(context)`.

dennisguse:
So far, we opened the SharedPreferences very often, which is very slow; also happened in the UI thread a lot.

Fixes #642.

PS: Expect crashes....

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
@@ -2,6 +2,7 @@ package de.dennisguse.opentracks.io.file.importer;
 
 import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.location.Location;
 import android.os.Looper;
 import android.util.Log;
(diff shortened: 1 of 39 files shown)


rgmf:
@dennisguse two little issues that result in crashes: when you open markers and when you start a new activity.

In `TrackRecordingActivity.java`, line `209':

```java
sharedPreferenceChangeListener.onSharedPreferenceChanged(null, null);
```

It should be this, right?

```java
sharedPreferenceChangeListener.onSharedPreferenceChanged(sharedPreferences, null);
```

And the same In `MarkerListActivity.java`, line 141.

These two changes will avoid crashes when you go to Markers and when you start a new activity.

I've installed this branch to test it ;)

dennisguse:
@rgmf exaclty one of those uygly bugs - I probably created some more.
Yesterday after work, I was somehow in the mood to refactor this - it was bothering me since some weeks when I realized what is happening there and we just copied the approach :)

rgmf (approved):
I've not found any more bugs these days. Okay for me ;)

PR #644, opened 2021-03-12
GPX: export and import power data.

### LINKED ISSUE #638: GPX: add sensor data
We could export all sensor data to GPX.

An approach by Strava:
https://developers.strava.com/docs/uploads/

Alternative to #637

dennisguse:
Uses Garmin extensions to export power data.
This is compatible with Golden Cheetah.

This is NOT supported by Strava.
They would require to use "power" instead of "PowerInWatts"./
See https://developers.strava.com/docs/uploads/

Fixes #638.


rgmf:
This is great :D

Tomorrow I'll check it out!!!

rgmf (approved):
Good for me.

I've found a bug introduced on here (I think): https://github.com/OpenTracksApp/OpenTracks/commit/503cba13150ebd4ab63abf745752df81e48c6227

I'm fixing it.

PS/ I'm becoming in a bug hunter (I've always hated them and now I'm one of them :rofl: ).

dennisguse:
@rgmf You are doing a great job there ;)

rgmf:
> @rgmf You are doing a great job there ;)

Thank you very much :blush:

Issue #512, opened 2020-11-10
Automatic export with default track name


slackline:
Thanks you for the on-going development of this App, its improving with every iteration.

I was particularly pleased to see the option to automatically export a track in a desired format on completion of recording, saves a lot of time doing it manually.

Having now recorded a couple of tracks with this new feature enabled I see the track is exported with its Number rather than the user specified "Default track name" which in my case I have configured to "Date (ISO8601)".

This is in v3.12.0/v3.12.0-d16cbe04a (Version code 3704).

Would it be possible to have the automatic exports use the "Default track name" as configured?

Thanks and keep up the excellent work.

enilkcals

rgmf:
@slackline you can read some discussion about this in #159 

I think is related ;)

slackline:
@rgmf Thanks for the pointer.

One of the reasons I prefer ISO 8601 and use it for manual exports is that it facilitates subsequent usage of the GPX files.  If I want to find one for an activity I did on a certain date I can get the from the filename without having to look inside each file.  In that sense ISO 8601 is human readable and adds context compared to track number.

rgmf:
@slackline you are right, I have about 200 tracks for now and I cannot find nothing through a File Manager in my computer :sweat_smile:

Someone in #159 pointed out that track's name would be useful in the exported file's name too. It's another possibility. Anyway, #159  is in stand by nowday.

slackline:
@rgmf Thanks, the thing with #159 seems to have actually been resolved, manually exporting files we have the option to choose the `Settings > Advanced Settings > Default Track Name > [Date (local) | Date (ISO8601) | Number]`.  I have `Date (ISO8601)` selected and when I export manually the GPX has the ISO8601 filename stub which is great.

I actually am planning on doing a bit more than Importing/Exporting the GPX tracks as I'd like to get a [Panel Dashboard](https://panel.holoviz.org/) written that looks at a directory using [pathlib](https://docs.python.org/3/library/pathlib.html) and reads in all files using the filename (ideally ISO8601) as keys to the dictionary with values as the GPX tracks ([gpxpy](https://pypi.org/gpxpy) makes reading GPX tracks simple). Work and life get in the way but when I get it running I'll share it so others can visualise their workouts.

Anyway, back on topic, as #159 works for manual exports I was surprised that the "Default" filename wasn't used by the "Automatic export".

dennisguse:
[...]

PR #646, opened 2021-03-13
Track exported names: UUID, track's name and extension

rgmf:
**Describe the pull request**
Exported track's name like proposed in #512 (see comment: https://github.com/OpenTracksApp/OpenTracks/issues/512#issuecomment-766437523)

Also, it checks if file that result of export process match with the following regular expression:
`<UUID first 8 bytes>.*\.<extension>`

PS/ If it is accepted I'd like to add/update tests, but I need to be sure that it's what was expected in #512.

**Link to the the issue**
#512

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/io/file/exporter/ExportActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/exporter/ExportActivity.java
@@ -91,7 +91,7 @@ public class ExportActivity extends FragmentActivity implements ExportServiceRes
     private ContentProviderUtils contentProviderUtils;
 
     // List of tracks to be exported.
-    private final ArrayList<Track.Id> trackIds = new ArrayList<>();
+    private final ArrayList<Track> tracks = new ArrayList<>();
 
     private final LinkedBlockingQueue<PendingConflict> conflictsQueue = new LinkedBlockingQueue<>();
     private final Handler conflictsHandler = new Handler();
@@ -187,7 +187,7 @@ public class ExportActivity extends FragmentActivity implements ExportServiceRes
     protected void onDestroy() {
         super.onDestroy();
         conflictsQueue.clear();
-        trackIds.clear();
+        tracks.clear();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


src/main/java/de/dennisguse/opentracks/io/file/exporter/ExportActivity.java
```diff
@@ -91,7 +92,7 @@
     private ContentProviderUtils contentProviderUtils;
 
     // List of tracks to be exported.
-    private final ArrayList<Track.Id> trackIds = new ArrayList<>();
+    private final ArrayList<UUID> uuids = new ArrayList<>();
```
dennisguse, line 95: That is the only place where we are passing the UUID around.
What is the advantage over passing the Track.Id or the Track?

dennisguse (commented):
@rgmf Cool that you picked up this task!

I was just confused by the UUID that is now passed around instead of the Track.Id.
What is the benefit?
So, we never used anywhere to identify a track except in the import stage.

src/main/java/de/dennisguse/opentracks/io/file/exporter/ExportActivity.java
rgmf: Because export method in `ExportActivity` (line 236) needs UUID to check if file already exists. Before it used Track.Id but now it needs UUID.

Also, PendingConflict calls that export method and the only thing it knows is UUID (before Track.Id).

PS/ I thought the same is to use Track.Id than to use UUID to identify tracks. I you are not sure about it I could re-think it.

dennisguse: It will work; it would just be the first time we use it internally.
What about using ArrayList<Track> instead?
The track is loaded anyhow to get the UUID, right?

dennisguse: Just realized that `<Track>` was hidden as I did not put it into backticks.

dennisguse:
That was fast :)
Looks good!

dennisguse (approved):

rgmf:
> That was fast :)
> Looks good!

I was working on it at the same time you were writing :rofl:

PR #649, opened 2021-03-13
Import: allow import multiple select files.

### LINKED ISSUE #559: import multiple GPX files via intent at once
Importing multiple GPX files at once would eventually be a nice addition. So far i have only researched and tested the manifest changes, as per snippet below. `ImportActivity` would need to receive multiple files most probably via `getParcelableArrayListExtra`, but i didn't manage to research this further.

```xml
<intent-filter>
    <action android:name="android.intent.action.SEND_MULTIPLE" />
    <category android:name="android.intent.category.DEFAULT" />
    <data android:mimeType="application/gpx+xml" />
</intent-filter>
<intent-filter>
  <action android:name="android.intent.action.SEND_MULTIPLE" />
  <category android:name="android.intent.category.DEFAULT" />
  <data android:mimeType="application/gpx" />
</intent-filter>
```

dennisguse:
Fixes #559.

@vanous Could you check that this works as you expect it to work?

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -73,11 +73,13 @@ limitations under the License.
 
             <intent-filter>
                 <action android:name="android.intent.action.SEND" />
+                <action android:name="android.intent.action.SEND_MULTIPLE" />
                 <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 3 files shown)


vanous:
@dennisguse this is great, works well! thank you very much!

PR #672, opened 2021-03-22
Added EGM2008 undulation data.

dennisguse:
* [x] add EGM2008 data (5 minute resolution from GeographicLib, PGM format)
* [x] parse EGM2008 data
* [x] use EGM2008 corrected in UI (@dennisguse)
* [x] bilinear ~~or bicubic interpolation~~ + fix tests expectations
* [x] add EGM2008 data source reference to about page 
* [x]  use EGM2008 corrected in charts (@dennisguse) (more complicated than it looks like
* [x] show that we use EGM2008 in the StatisticsRecordingFragment? Or make it configurable: WGS84 or EGM2008.

PS: EGM2008 lookup is cached, but happens currently in the UI thread.
For now that is okay, but is not cool.
Lookup should only happens once per track at a 5' minute resolution.

Adds about ~~20MB~~ 16MB as data to the APK.

More infos: https://geographiclib.sourceforge.io/html/geoid.html

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/util/EGM2008Utils.java
+++ b/src/main/java/de/dennisguse/opentracks/util/EGM2008Utils.java
@@ -1,0 +1,176 @@
+package de.dennisguse.opentracks.util;
+
+import android.content.Context;
+import android.location.Location;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
+
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.util.Objects;
+
+import de.dennisguse.opentracks.R;
+
+/**
+ * Converts WGS84 altitude to EGM2008 (should be close to height above sea level).
+ * <p>
+ * Uses <a href="https://geographiclib.sourceforge.io/">GeographicLib</a>] EGM2008 5minute undulation data.
+ * https://geographiclib.sourceforge.io/html/geoid.html
+ * <p>
+ * File starts at 90N, 0E (North pole) and is encoded in parallel bands as unsigned shorts.
+ */
+public class EGM2008Utils {
+
+    static final int EGM2008_5_DATA = R.raw.egm2008_5;
+
+    private static final int HEADER_LENGTH = 404;
+
+    private static final int RESOLUTION_IN_MINUTES = 60 / 5;
+
+    private static final int LATITUDE_CORRECTION = 360 * RESOLUTION_IN_MINUTES;
+
+    private EGM2008Utils() {
+    }
+
+    public static EGM2008Correction createEGM2008Correction(Context context, Location location) throws IOException {
+        Indices indices = getIndices(location);
+
+        try (DataInputStream dataInputStream = new DataInputStream(context.getResources().openRawResource(EGM2008_5_DATA))) {
+            return new EGM2008Correction(indices, dataInputStream);
+        }
+    }
+
+    @VisibleForTesting
+    static int getUndulationRaw(DataInputStream dataInputStream, Indices indices) throws IOException {
+        dataInputStream.reset();
+        int absoluteIndex = indices.getAbsoluteIndex();
+        return getUndulationRaw(dataInputStream, absoluteIndex);
+    }
+
+    private static int getUndulationRaw(DataInputStream dataInputStream, int undulationIndex) throws IOException {
+        dataInputStream.reset();
+        int index = HEADER_LENGTH + undulationIndex * 2;  //byte size is 2
+        long ignored = dataInputStream.skip(index);
+
+        return dataInputStream.readUnsignedShort();
+    }
+
+    @VisibleForTesting
+    static Indices getIndices(Location location) {
+        double latitude = -location.getLatitude() + 90;
+        int latitudeIndex = (int) (latitude * RESOLUTION_IN_MINUTES);
+
+        double longitude;
+        if (location.getLongitude() >= 0) {
+            longitude = location.getLongitude();
+        } else {
+            longitude = 180 + Math.abs(location.getLongitude());
+        }
+        int longitudeIndex = (int) (longitude * RESOLUTION_IN_MINUTES);
+
+        if (longitudeIndex >= 360 * RESOLUTION_IN_MINUTES) {
+            longitudeIndex = 0;
+        }
+        return new Indices(latitudeIndex, longitudeIndex);
+    }
+
+    public static class EGM2008Correction {
+
+        protected final Indices indices;
+
+        protected final int v00;
+        protected final int v10;
+        protected final int v01;
+        protected final int v11;
+
+        public EGM2008Correction(Indices indices, DataInputStream dataInputStream) throws IOException {
+            this.indices = indices;
+
+            v00 = getUndulationRaw(dataInputStream, indices);
+            if (!isSouthPole()) {
+                v10 = getUndulationRaw(dataInputStream, indices.offset(0, 1));
+                v01 = getUndulationRaw(dataInputStream, indices.offset(1, 0));
+                v11 = getUndulationRaw(dataInputStream, indices.offset(1, 1));
+            } else {
+                v10 = 0;
+                v01 = 0;
+                v11 = 0;
+            }
+        }
+
+        public boolean canCorrect(@NonNull Location location) {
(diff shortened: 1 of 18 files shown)


src/main/java/de/dennisguse/opentracks/util/EGM2008Utils.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final int HEADER_LENGTH = 403;
+
+    private static final int RESOLUTION_IN_MINUTES = 60 / 5;
+
+    private static final int LATITUDE_CORRECTION = 180 * RESOLUTION_IN_MINUTES;
+
+    private EGM2008Utils() {
+    }
+
+    public static double correctedAltitude(Context context, Location location) throws IOException {
```
gdt, line 33: This looks like it is opening the file every time.    I would expect a pattern of opening the file and reading it once - but I am not used to android and it may be better to do more file ops and be less piggy with RAM as you are doing.

gdt: Actually, it may make sense to keep the file open, saving that time, and still seek to find things.  It would be really good to seek, vs reading and discarding.  (I am of course speaking in POSIX API, not Android..)

gdt:
Given that this is for fitness, not geodesy, I think some attention to accuracy is warranted but that you probably don't want to go down the path of more accuracy for more complexity too far.   So I would be inclined to go for bilinear.  The basic concept is that for a given coordinate pair, you round down and up in lat and lon which in 4 combinations gets you 4 points that form a square of the grid resolution such that the desired point is inside.   I am pretty sure that you just sum the 4 values where the weight of each is "( (1-dlat)/latgrid + (1-dlon)longrid ) /2" where dlat and dlon are the difference between the corner coordinate and the desired location.

I think this is equivalent to first finding the points on the two constant-longitude sides of the squares at the desired latitude by simple linear interoplation between the two points (doing this twice) and then doing one more linear interpolation in longtitude.

src/main/java/de/dennisguse/opentracks/util/EGM2008Utils.java
dennisguse: (I hope) it is not that problematic as we don't need to much undulation data.
Usually, the user will stay in one cell only and we can cache this value in the UI.
If some people track their car/airplane rides that might be different, but we will fix it then.
Then we only do this lookup once. No need to keep the whole file in memory.

And we only get an InputStream from Android to access assets.
Let's see if this works ;)

PS: I need to that only the undulation offset is returned and the getAlltitude() modified by UI only.

gdt: I guess there are two separate questions.    One is that given a  location do you want to interpolate or just pick closest.   The other is how long to use that location's result -- meaning for other  nearby points -- before refetching.   You could perhaps store [original_location, offset] and then as long as new_location is close use it, else look up again.    It sounds like you are just going to skip that for now  and assume all locations in a  track are close enough - which is still a vast improvement over the old way.

dennisguse:
Bilinear according to GeographicLib: GeoId.cpp:
```
        v00 = rawval(ix    , iy    );
        v01 = rawval(ix + 1, iy    );
        v10 = rawval(ix    , iy + 1);
        v11 = rawval(ix + 1, iy + 1);
```
and:
```
      real
        a = (1 - fx) * v00 + fx * v01,
        b = (1 - fx) * v10 + fx * v11,
        c = (1 - fy) * a + fy * b,
        h = _offset + _scale * c;
```
with PGM index computation:
```
      if (ix < 0)
        ix += _width;
      else if (ix >= _width)
        ix -= _width;
        if (iy < 0 || iy >= _height) {
          iy = iy < 0 ? -iy : 2 * (_height - 1) - iy;
          ix += (ix < _width/2 ? 1 : -1) * _width/2;
        }
``` 

Let's presume that they know what they are doing :)

gdt:
> Let's presume that they know what they are doing :)

That looks good to me and was what I was trying to say.   And yes, the GeographicLib author is a very reliable source for getting things far harder than this exactly right.

dennisguse:
@gdt you seem to have some experience with this kind of data.
Could you check if the implementation is correct?
That would be awesome!

rgmf (commented):
It doesn't work for me. When OT is recording all fields are 0. Only time is working.

dennisguse:
@rgmf Weird - the updating while recording is not happening.
The data is collected and after stopping the recording, it is shown.
Thanks for testing!

dennisguse:
@rgmf I found the issue: I "optimized" TrackDataHub and the contentobservers were not registered correctly.
I removed all unnecessary changes (besides EGM2008) and force pushed.
Those changes would have clashed with #680 anyhow and were therefore of no use.
Thanks!

rgmf:
> @rgmf I found the issue: I "optimized" TrackDataHub and the contentobservers were not registered correctly.
> I removed all unnecessary changes (besides EGM2008) and force pushed.
> Those changes would have clashed with #680 anyhow and were therefore of no use.
> Thanks!

Great!!! I'm going to install this branch to check it oit tomorrow :D

rgmf:
@dennisguse I couldn't see corrections of altitude on my area, I mean I see the same altitude as up to now. Let me figure out what happend and I'll let you know.

dennisguse:
@rgmf The difference might be rather small depending on your location (in Berlin it's 35m).
You can get your local the undulation value from here: https://geographiclib.sourceforge.io/cgi-bin/GeoidEval

I needed to do an actual test walk as the horizontal GPS accuracy is not that great - especially in cities with large buildings. Started at a 100m and then went to the actual 35m (took almost 10min) - so depending on the horizontal accuracy it may vary as well. :shrug:

dennisguse:
Ready to go.
Wait for v3.16.0 availability on F-Droid.

dennisguse:
For the charts, I removed the smoothing of the altitude to be able to get it EGM2008 corrected.
But I just realized that it looks ugly, so we need the smoothing back.
Also these chart-changes will conflict with the refactoring of IntervalStatistics #703.

@rgmf Following options:
A) merge this PR without smoothing but with EGM2008 in charts (as it is now)
B) merge this PR with smoothing but without EGM2008 in charts
C) wait until #703 is done

I would prefer to do B) as this will allow to merge the functionality, and we can then implement #703 and then figure out how to do the EGM2008 in smoothed charts.
This has the disadvantage that the charts will be in WGS84 while atitude while recording will be in EGM2008.
For me #703, is more important as it simplifies the code base and complicating it beforehand is not really helpful (IMHO).

rgmf:
@dennisguse I was going to merge #703... It's squashed now and ready for merge. Do what you think the better option is. If I have to fix conflicts on #703 don't worry, I'll do it ;)

dennisguse:
@rgmf while checking the feasibility, I realized that we should do C).
My concept was that we do the EGM computation in the TrackDataHub alone and then pass it around in the UI from there.
But this also requires a change in the TrackStatisticsUpdater (for altitude smoothing), but TrackStatisticsUpdater is also used during import (here we don't need EGM) and in the TrackRecordingService (here we don't need EGM as well). We might need it for voice announcements (rather unlikely).

Long story short: we need may need to use two slightly different TrackStatisticsUpdater...
But it will be easier to decide when #703 is integrated.

dennisguse:
@rgmf We need to finish #703 first, #701should not be of any conflict :)

rgmf:
@dennisguse it's #701 what is ready (the PR) not #703 (the issue)... my mistake. Sorry for the confusion.

Now I undertand everything :smile: 

If you want, you can finish EGM2008 because I've not already started with #701 (I was planing to check it out tomorrow).

dennisguse:
@rgmf I have no concept for this at the moment.
So, I guess moving to TrackStatisticsUpdater everywhere first, makes the EGM integration easier (or at least getting a plan).
It's complicated as we have to cache the EGM lookups - I/O-Operations on 15MB stored data are kind of expansive.
No need to hurry, the math is done and the integration can also wait for some more days.

gdt:
I am still intending to pay attention bere, but haven't and do not wait for me!

dennisguse:
Removed conflicting TrackDataHub/Statistics changes.
Wait until #736 is merged.

PR #680, opened 2021-03-25
Track recording service handle and offer status (trackId, pause and GPS status)

rgmf:
**Describe the pull request**

- [x] Add service callback for trackId and recording state

I've rewritten the TrackRecordingServiceCallback to create an interface through which the other components can receive recording status in general: trackId, pause and gps (now it's easy to add more information to that status in the future).

- [x] Replace SharedPreference recordingTrackId, recordingTrackPaused with service callback (everywhere)

Preferences doesn't handle nothing about it anymore.

- [x] Replace SharedPreference recordingTrackId, recordingTrackPaused with service callback (everywhere)

Preferences doesn't handle nothing about it anymore.

- [x] Remove RECORDING_TRACK_ID_DEFAULT
- [x] Remove PreferencesUtil methods.
- [x] Remove TrackRecordingService restart feature (after getting killed by Android OS / it dies)

If TrackRecordingService crashes or somethink like this then user could resume the track that was recording.

**Really Important Notes**
- Not ready to merge.
- Ready to test, test and test and decide if it's a good approach.
- @dennisguse if this was what you expected let me know because there are a lot of things to do on tests... If it's okay I'll do it ;)

**Link to the the issue**
#388 

**License agreement**
By opening this pull request, you are providing your contribution under the _Apache License 2.0_ (see [LICENSE.md](LICENSE.md)).

PR #692, opened 2021-04-02
Use a non-static Binder for TrackRecordingService.

### LINKED ISSUE #98: Use non-static service binder possible?
There is a warning in the `TrackRecordingService` that the service binder needs to be static.
Is this still necessary?

```
    /**
     * TODO: There is a bug in Android that leaks Binder instances. This bug is
     * especially visible if we have a non-static class, as there is no way to
     * nullify reference to the outer class (the service). A workaround is to use
     * a static class and explicitly clear service and detach it from the
     * underlying Binder. With this approach, we minimize the leak to 24 bytes per
     * each service instance. For more details, see the following bug:
     * http://code.google.com/p/android/issues/detail?id=6426.
     */
```

See: https://github.com/dennisguse/OpenTracks/blob/1750236474f72ec714f744102408e1873be4e72e/src/main/java/de/dennisguse/opentracks/services/TrackRecordingService.java#L928

dennisguse:
Fixes #98.

* [x] Wait until #680 is merged and also remove TrackRecordingServiceInterface.


rgmf (approved):
How simple service is now!!!!

I've not found any issue in a couple of recordings I've done. Look great!!

PR #693, opened 2021-04-02
Rename elevation to altitude

### LINKED ISSUE #678: Rename elevation to altitude
We are using both words interchangeably, but the Android API uses altitude.
So, that's the one to use.

PS: Do not rename DB column as this is too much effort.

dennisguse:
* [x] absolute altitude
* [x] elevation gain / elevation loss
* [x] translation keys (pull first from Weblate, then push to main)
* [x] Rename TrackRecordingService methods
Fixes #678.

Wait for #680.


dennisguse:
@rgmf Altitude (instead of elevation) is now used everywhere throughout the code base.
If you are working on something larger, it would avoid merge conflicts if you rebase to current main.

(wanted to be fast with this PR as it contains changes of translation keys and thus changes in weblate would introduce merge conflicts that are ugly to deal with).

PR #697, opened 2021-04-05
While recording show coordinates by default.

dennisguse:
Show the coordinates by default, so new users are aware that they can get them if needed.


rgmf (approved):
Great idea. It's a field I never use but yo are right about the new users.

PR #708, opened 2021-04-10
Upgrade to Gradle 7.0

dennisguse:
Released on 2021-04-09.

We may need to wait some time until this can be used on F-Droid.
I recall that we had some trouble with failing builds when upgrading to 6.5 (for some time).

Edit: it was Gradle 5.6.
https://github.com/OpenTracksApp/OpenTracks/commit/fa6329c1575d68ec954f75c9a820c6d9b4086463


dennisguse:
Available on F-Droid build server:  https://github.com/f-droid/fdroidserver/commit/0298eef46fd660aef8811708aaaacd558d60158a

PR #712, opened 2021-04-16
Distance interval bugfixed: interval options are km/mi values

rgmf:
@dennisguse maybe we need a new release because intervals are computed using 1m, 2m, 5m, 10m, 20m, 50m instead of 1km, 2km, 5km, 10km, 20km, 50km.

Sorry, but I couldn't test new Distance code until now :(

This PR fixes that.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/content/data/Distance.java
+++ b/src/main/java/de/dennisguse/opentracks/content/data/Distance.java
@@ -20,9 +20,13 @@ public class Distance {
         return of(distance_mile * UnitConversions.MI_TO_M);
     }
 
+    public static Distance ofKilometer(double distance_km) {
+        return of(distance_km * UnitConversions.KM_TO_M);
+    }
+
     public static Distance one(boolean metricUnit) {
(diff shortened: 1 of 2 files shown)


rgmf:
@dennisguse the fixes is simple so if you think there is a better approach simply do it and discard this PR. At least, this PR shows you what the issue is ;)

dennisguse:
My plan was to do all computations within Distance.
PS: Distance.one(metricUnits) was actually broken...

@rgmf What about this?
(I added a commit)
If you like, just squash and merge ;)

rgmf:
> My plan was to do all computations within Distance.
> PS: Distance.one(metricUnits) was actually broken...

I take note ;)

> @rgmf What about this?
> (I added a commit)
> If you like, just squash and merge ;)

It's simple awesome.

I did it, I don't matter whose commit is but this should be yours. Anyway, bug is fixed that it's the important thing. Next time do it you please, it's strange see my name in this commit :rofl: :rofl: :rofl:

PR #715, opened 2021-04-17
registerListener for barometric sensor uses SENSOR_DELAY_FASTEST

### LINKED ISSUE #714: Gain and loss don't work
**Describe the bug**
Recording a track show dashes on gain and loss fields.

**Technical information**
 - Version 3.17 (F-Droid)

rgmf:
**Describe the pull request**
`sensorManager.registerListener` didn't work because `samplingPeriodUs` arguments was in nanosecons. Seem that delay must be in microseconds. According to the documentation:

The rate `android.hardware.SensorEvent sensor events` are delivered at. This is only a hint to the system. Events may be received faster or slower than the specified rate. Usually events are received faster. The value must be one of `#SENSOR_DELAY_NORMAL`, ` #SENSOR_DELAY_UI`, `#SENSOR_DELAY_GAME`, or `#SENSOR_DELAY_FASTEST` or, the desired delay between events in microseconds. Specifying the delay in microseconds only works from Android 2.3 (API level 9) onwards. For earlier releases, you must use one of the `SENSOR_DELAY_*` constants.

So, I propose to use SENSOR_DELAY_FASTEST.

**Link to the the issue**
This PR fixes #714

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/services/sensors/AltitudeSumManager.java
+++ b/src/main/java/de/dennisguse/opentracks/services/sensors/AltitudeSumManager.java
@@ -40,7 +40,7 @@ public class AltitudeSumManager implements SensorEventListener {
             Log.w(TAG, "No pressure sensor available.");
             isConnected = false;
         } else {
-            isConnected = sensorManager.registerListener(this, pressureSensor, (int) SAMPLING_RATE.toNanos());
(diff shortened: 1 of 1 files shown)


dennisguse:
+1
Thanks for this!

Do you recall why we used 3s in the first place?

rgmf:
> Do you recall why we used 3s in the first place?

I have no idea 😄 

But reading the docs seem that Android does what it wants so it's not that important.

PR #731, opened 2021-04-24
Some improvements for #719

### LINKED ISSUE #719: MemoryLeak as TrackRecordingListeners are not unregistered
**Describe the bug**
When we refactored to use listener to communicate the TrackRecordingService, we are only registering but __we never unregister__.
Therefore, we (a) keep objects in memory that are not used anymore until TrackRecordingService shuts down and (b) we inform objects about changes that are already finished.

__Update:__ should not be too bad as the TrackRecordingService is terminated if stopped.
We might have some trouble using pause and voice announcements (see below NPE).

**To Reproduce**
I saw this issue while executing tests that resulted in NPE as the TrackRecordingService was re-used for a recording.
The NPE was triggered in the AnnouncementPeriodTask as it was already shutdown, but received an update from TrackRecordingService.

I got this issue due to 772dad48372c4fba9e4de2afc89fb7ab42ee712c
(not the actual reason, but not correct solution anyhow).

**Technical information**
 - Version v3.17.1

dennisguse:
Does not fix #719.

TrackDataHub does not need to register for TrackRecordingService status changes as this can be handled by TrackRecordingActivity alone.

* reduces impact of memory leak as TrackDataHub keeps references to fragments; now only the activities are referenced by TrackRecordingServices and those at least remove all there references.
* (so far) TrackDataHub started TrackRecordingSevice due to binding to the service (even if only showing already recorded tracks)

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/chart/ChartPointTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/chart/ChartPointTest.java
@@ -1,0 +1,94 @@
+package de.dennisguse.opentracks.chart;
+
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.time.Duration;
+
+import de.dennisguse.opentracks.content.data.Distance;
+import de.dennisguse.opentracks.content.data.Speed;
+import de.dennisguse.opentracks.content.data.TrackPoint;
+import de.dennisguse.opentracks.fragments.TrackStubUtils;
+import de.dennisguse.opentracks.stats.TrackStatistics;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+@RunWith(AndroidJUnit4.class)
+public class ChartPointTest {
+
+    @Test
+    public void create_by_time() {
+        // given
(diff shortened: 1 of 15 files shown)


rgmf (changes requested):
Looks like it isn't resume properly. When you are recording an activity you can see speed and location but if you force a TrackRecordingActivity (StatisticsRecordingFragment) resume then `isSelectedTrackRecording()` returns false and you cannot see speed and location.

To reproduce:
- Start a new recording. You can see all okay.
- Press back button so you are now in TrackListActivity.
- Click on track that is recording so you back to TrackRecordingActivity.
- You cannot see speed and location anymore.

dennisguse:
@rgmf Thanks!
i have a hard time reproducing this
Could you check if this helps?
https://github.com/OpenTracksApp/OpenTracks/pull/731/commits/1bcd77abeb5d9e69c4ae5e3e5daee5d53c0371a9

rgmf:
> @rgmf Thanks!
> i have a hard time reproducing this
> Could you check if this helps?
> [1bcd77a](https://github.com/OpenTracksApp/OpenTracks/commit/1bcd77abeb5d9e69c4ae5e3e5daee5d53c0371a9)

It works now ;)

rgmf (approved):

Issue #705, opened 2021-04-08
Reported via Google Play Store: crash while deleting


dennisguse:
Happened once; don't know why.

```
java.lang.IllegalThreadStateException: 
  at java.lang.Thread.start (Thread.java:868)
  at de.dennisguse.opentracks.TrackDeleteActivity.onStart (TrackDeleteActivity.java:72)
  at android.app.Instrumentation.callActivityOnStart (Instrumentation.java:1433)
  at android.app.Activity.performStart (Activity.java:7923)
  at android.app.ActivityThread.handleStartActivity (ActivityThread.java:3332)
  at android.app.servertransaction.TransactionExecutor.performLifecycleSequence (TransactionExecutor.java:221)
  at android.app.servertransaction.TransactionExecutor.cycleToPath (TransactionExecutor.java:201)
  at android.app.servertransaction.TransactionExecutor.executeLifecycleState (TransactionExecutor.java:173)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:97)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2044)
  at android.os.Handler.dispatchMessage (Handler.java:107)
  at android.os.Looper.loop (Looper.java:224)
  at android.app.ActivityThread.main (ActivityThread.java:7562)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:539)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:950)
```

rgmf:
Reproduce:
1. Delete an activity.
2. Move quickly to another app.
3. Wait for a while.
4. Back to OpenTrack that have finished the deleting.
5. And you will see the crash.

PR #734, opened 2021-04-25
Bugfix: crash deleting when activity restarts

rgmf:
**Describe the pull request**
Thread starts on onCreate (once).
Activity's onStart checks if thread was terminated and activity finishes in that case.
Activity's onStop interrupts the thread.

All this avoids crashing when activity restarts while thread is deleting.

**Link to the the issue**
#705 

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/TrackDeleteActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/TrackDeleteActivity.java
@@ -64,24 +64,27 @@ public class TrackDeleteActivity extends AbstractActivity {
 
             runOnUiThread(TrackDeleteActivity.this::onAsyncTaskCompleted);
         });
-    }
-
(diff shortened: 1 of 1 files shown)


dennisguse:
Nice work!

One follow-up is the actual question if deletion is actually properly implemented (according to Android lifecycle).
In theory the deletion might be long running (which lead to the crash) and should be rather implemented as a service?

dennisguse (approved):

dennisguse:
@rgmf Can we merge this? I would then tag the next (bugfix) release in the v3.17-series.

rgmf:
@dennisguse we can merge it to fix the issue, but we should improve it to use properly the thread with android lifecycle so I'm going to open an issue for that.

dennisguse:
Thanks :)

Issue #42, opened 2019-10-13
Bug: sensor data diagram handles missing data wrongly


dennisguse:
As the heart rate sensor is disconnected (and not reconnected; see #41), the diagram should not draw the last point in the heart rate diagram (0bpm).

![Screenshot_20191013-094812_OpenTracks](https://user-images.githubusercontent.com/2820825/66712818-df660200-eda2-11e9-9801-673cb7a3b2f6.png)

dennisguse:
The issue seems to originate from `ChartView.drawPaths()`.
However, the whole chart infrastructure (data passing as object instead of `double[]`) should be fixed first.
The code is rather complicated...

dennisguse:
This also affects cadence and speed Bluetooth LE sensor

PR #736, opened 2021-04-27
TrackRecordingService: provide status updates via LiveData.

dennisguse:
Just some work in progress.

* [x] TrackRecordingActivity.
* [x] MarkerListActivity
* [x] TrackListActivity
* [x] TrackRecordedActivity
* [x] SettingsActivity.PrefsFragment

TODO: 
* [x] check if LiveData reference should be kept in subscribers
* [x] check what happens when subscribing multiple times: we get the event multiple times; not that problematic
* [x] check if object reference leaks
* [x] remove TrackRecordingServiceStatus


dennisguse:
Stale references are not an issue: https://developer.android.com/topic/libraries/architecture/livedata
```
You can register an observer paired with an object that implements the LifecycleOwner interface. This relationship allows the observer to be removed when the state of the corresponding Lifecycle object changes to DESTROYED. This is especially useful for activities and fragments because they can safely observe LiveData objects and not worry about leaks—activities and fragments are instantly unsubscribed when their lifecycles are destroyed.
```

dennisguse:
@rgmf @pstorch Could you give this a spin?

I hope I did not break too much....

rgmf:
> @rgmf @pstorch Could you give this a spin?
> 
> I hope I did not break too much....

Installing it... and ready to check it out :smile: 

I'll use it this weekend and I'll tell you how it's going.

pstorch:
I think I can also give it a try at the weekend.

pstorch:
Anything special to look at?
A recording while a short walk to the bakery worked without issue.

rgmf:
It works well for me, not any bug found, but only a strange issue in the  graph drawn. I don't know if it's related with this issue but looks like it is not:

Screenshot_20210430-173443_OpenTracks_(Debug)

dennisguse:
@rgmf That is not related to this change.
Some weeks ago I was fixing this https://github.com/OpenTracksApp/OpenTracks/issues/42
This looks related.

And the elevation is broken: I can reproduce this; issue seems to be that the conversion is incorrect; this happened in this PR (I guess).

rgmf:
@dennisguse after recording, back to list and GPS status remains with this icon:

Screenshot_20210501-204134_OpenTracks_(Debug)_1

dennisguse:
@rgmf I added this commit to always update the GpsStatus, but I am not sure that this is the best solution.
And we seemingly get a delayed callback after TrackRecordingService.onDestroy() is called. I just catch this with a null check - not really cool: https://github.com/OpenTracksApp/OpenTracks/pull/736/commits/cc775655299a04fd087f7f42e4c40bcce433768d

dennisguse:
@rgmf @pstorch Any more things that you found while testing?

rgmf:
> @rgmf @pstorch Any more things that you found while testing?

I've not found any more things.

pstorch:
> @rgmf @pstorch Any more things that you found while testing?

Nothing else from my side

dennisguse:
@rgmf @pstorch Thanks! :partying_face:

PR #738, opened 2021-04-29
Reproducible build: disable PNG crunching.

dennisguse:
Increases APK size by 4kB.


dennisguse:
Let's see if the next release via F-Droid is reproducible...

PR #743, opened 2021-05-03
Bugfix: voice announcements need to use KM/Miles (instead of m/feet).

### LINKED ISSUE #722: newest bugs after update // voice announcements by distance not working
- can't export my runs anymore. I get the message that their was an error for all the exports...
- I was using voice announcements for every 5km but it kept on announcing without pause. So I had to turn off the announcements.
- downgrade to older version viva Fdroid always resulted in an error that the installation was aborted.
- app crashed several times after the update to 3.17, 

Thank you for your good work!

dennisguse:
Fixes #722.

While introducing Distance, I used m/feet instead of KM/Miles to trigger the voice announcements based upon distance.

Reason: lack of test coverage.


dennisguse:
Just did a test walk (literally) and it is working - I used a Nexus 7 to test it.

PR #746, opened 2021-05-05
[Does not work with 4.2.1; only with 7.0.0] Upgrade to JDK11

dennisguse:
Android gradle 4.2 allows use of JDK11.

Sadly, not `InputStream.transferTo()`...
But we get `List.of()`!

There might be more things, we can use; this is just AndroidStudio's code analyzer.


pstorch:
Interesting, so AS and Gradle produce compatible bytecode for older devices?

dennisguse:
@pstorch At least on an Android 7.1.2 it was working, so I guess yes.
On CI it fails and I guess the JDK11 is not bundled into the CI that we use.

dennisguse:
AndroidStudio 4.2.1 itself can compile the code.

__Issue:__ This is blocked as `./gradlew build` fails: ` error: package android.util does not exist`.
Seemingly, the android.jar is not found if using JDK11 as source/target.
Tested on Linux/Windows using AndroidStudio 4.2.1 / gradle 7.0.2.

On Windows: using the JDK11 bundled with AndroidStudio 
On Linux: also with local OpenJDK11

dennisguse:
Got it: it is an issue with Android Gradle Plugin in version 4.2.X.
Using 7.0.0-X works like charm, but then AndroidStudio itself cannot build anymore.

dennisguse:
Not valid anymore - we will do the JDK11 upgrade when the next AndroidStudio get's released.

Issue #735, opened 2021-04-27
Implement deletion track according to Android lifecycle


rgmf:
**Is your feature request related to a problem? Please describe.**
Follow-up #734.

Read comment: https://github.com/OpenTracksApp/OpenTracks/pull/734#issuecomment-826386091

Basically we are using a Thread that will update UI if Activity is started/resumed. We currently have a collateral issue: when user tap back button while track is deleting then the ActivityList show the track already delete in background by the thread. Usually this is difficult because deletion is fast but it can be occur.

Sum up: we need another approach for this.

**Describe the solution you'd like**
Using an approach that match properly with Android Lifecycle.

rgmf:
I think the best solution should be a service so the TrackListActivity can bind to it and receive delete result by the service. But I have some doubts:

TrackListActivity will has to try to bind to the service. Can this be a problem? I mean, every time TrackListActivity is resumed or started it'll has to check if delete service is running to bind to it.

@dennisguse do you think that can be a problem? Am I missing something?

In that case some classes have to be edited:
- TrackListActivity.
- TrackRecordedActivity.
- AbstractListActivity.
- TrackDeleteActivity.

dennisguse:
Don't know - but I guess that won't be a (big) problem.
One data point from TrackDataHub: it was always trying to bind the service until recently.

Can we somehow get rid of TrackDeleteActivity by moving this to a separate service?
And I guess only TrackListActivity would need to check the progress - assuming that TrackRecordedActivity triggers deletion and then finishes. If TrackListActiivty then prevents opening Tracks that are in deletion, we could avoid this.

rgmf:
> Can we somehow get rid of TrackDeleteActivity by moving this to a separate service?

Yes, that's the idea.

dennisguse:
@rgmf Forgot: i am working on #736 (memoryLeak/liveData - hopefully done until the weekend) and this might create some conflicts with this issue.

rgmf:
@dennisguse okay, then I guess it'll be better to wait.

PR #752, opened 2021-05-09
Track deletion is handled by a service

rgmf:
**Describe the pull request**
Track deletion is handled by a started/binded service.
The service shows a notification on the status bar with a progress bar so user can use OpenTracks tough tracks are deleting. Nothing is blocked. If deletion is slow then user can see the progress.

Captura de pantalla_2021-05-09_17-20-25

**Link to the the issue**
#735


.idea/misc.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" default="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
dennisguse, line 3: Can you move this to a separate commit? It is related to AndroidStudio 4.2 update.
(incl runConfigurations.xml)

src/main/java/de/dennisguse/opentracks/AbstractTrackDeleteActivity.java
```diff
@@ -38,21 +37,46 @@
...
+    private static final String TAG = AbstractTrackDeleteActivity.class.getSimpleName();
 
     protected static final int GPS_REQUEST_CODE = 6;
-    private static final int DELETE_REQUEST_CODE = 3;
+
+    private TrackDeleteServiceConnection trackDeleteServiceConnection;
+
+    private ArrayList<Track.Id> trackIds;
+
+    private final Runnable bindChangedStatusCallback = new Runnable() {
```
dennisguse, line 51: We should change this interface as the ServiceConnection should only call this when it get's connected.
Also in the other activities this code is not suitable for "service died notifications".
And then we can actually pass the service as a parameter.

src/main/java/de/dennisguse/opentracks/AbstractTrackDeleteActivity.java
```diff
@@ -38,21 +37,46 @@
...
+        @Override
+        public void run() {
+            if (trackDeleteServiceConnection == null) {
```
dennisguse, line 54: should not happen

src/main/java/de/dennisguse/opentracks/AbstractTrackDeleteActivity.java
```diff
@@ -38,21 +37,46 @@
...
     @Override
-    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
-        if (requestCode == DELETE_REQUEST_CODE) {
-            onTrackDeleted();
-        } else {
-            super.onActivityResult(requestCode, resultCode, data);
+    protected void onStop() {
+        super.onStop();
+        if (trackDeleteServiceConnection != null) {
+            trackDeleteServiceConnection.unbind(this);
```
dennisguse, line 79: also `trackDeleteServiceConnection = null`

null check not needed.

src/main/java/de/dennisguse/opentracks/AbstractTrackDeleteActivity.java
```diff
@@ -38,21 +37,46 @@
...
+            }
+
+            service.getDeletingStatusObservable().observe(AbstractTrackDeleteActivity.this, status -> onTrackDeleteStatus(status));
+        }
+    };
+
+    @Override
+    protected void onStart() {
+        super.onStart();
+        if (trackDeleteServiceConnection == null) {
```
dennisguse, line 69: always create a new one

src/main/java/de/dennisguse/opentracks/TrackRecordedActivity.java
```diff
@@ -248,9 +249,16 @@ public boolean onOptionsItemSelected(MenuItem item) {
         return recordingStatus.getTrackId();
     }
 
+    @Override
+    protected void onTrackDeleteStatus(TrackDeleteService.DeleteStatus deleteStatus) {
```
dennisguse, line 253: why twice?
onDeleteConfirmed does something very similar, right?

src/main/java/de/dennisguse/opentracks/AbstractTrackDeleteActivity.java
```diff
@@ -69,6 +93,8 @@ protected void deleteTracks(Track.Id... trackIds) {
     public void onConfirmDeleteDone(Track.Id... trackIds) {
```
dennisguse, line 93: Why not pass a list?

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
```diff
@@ -0,0 +1,159 @@
...
+        public TrackDeleteService getService() {
+            return TrackDeleteService.this;
+        }
+    }
+
+    public static class DeleteStatus {
+        private int progress;
+        private int max;
+
+        public DeleteStatus(int progress, int max) {
```
dennisguse, line 142: private

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        public TrackDeleteService getService() {
+            return TrackDeleteService.this;
+        }
+    }
+
+    public static class DeleteStatus {
+        private int progress;
+        private int max;
```
dennisguse, line 140: max is count of tracks to be deleted and progress the current one?

src/main/java/de/dennisguse/opentracks/services/TrackDeleteServiceConnection.java
```diff
@@ -0,0 +1,68 @@
...
+import androidx.annotation.NonNull;
+
+import java.util.ArrayList;
+
+import de.dennisguse.opentracks.BuildConfig;
+import de.dennisguse.opentracks.content.data.Track;
+
+import static de.dennisguse.opentracks.services.TrackDeleteService.EXTRA_TRACK_IDS;
+
+public class TrackDeleteServiceConnection implements ServiceConnection {
```
dennisguse, line 18: Same as TrackRecordingServiceConnection, right?
Why not use ServiceConnection<T extends Service> as a base class?

dennisguse (commented):
Did not yet test (was too late).

dennisguse:
And on my personal whishlist: an EspressoUI test that checks that deletion works.
Just in case somebody (usually me) tries to "improve" something ;)

src/main/java/de/dennisguse/opentracks/AbstractTrackDeleteActivity.java
rgmf: Okay, I followed the same approach than TrackRecordingService. In fact I loved that approach the first time I saw it :roll_eyes: 
Do you know another approach for this?
I'm going to look into it.

src/main/java/de/dennisguse/opentracks/AbstractTrackDeleteActivity.java
rgmf: I didn't want to touch it.

rgmf: This should be fixed/changed in other separated issue because it affects a lot of parts: `Track.Id... trackIds` are everywhere.

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
rgmf: That's right.
I've used the same names that Android in `BuilderNotification.setProgress(int max, int progress, int indeterminate)`

src/main/java/de/dennisguse/opentracks/services/TrackDeleteServiceConnection.java
rgmf: > Same as TrackRecordingServiceConnection, right?

Yes, the approach is the same, but the implementation of both connections are different.

rgmf:
> And on my personal whishlist: an EspressoUI test that checks that deletion works.
> Just in case somebody (usually me) tries to "improve" something ;)

Not usually you but everyone :rofl: :rofl: :rofl: 

That makes sense, I should implement some EspressoUI tests... so I'm going to do it.

src/main/java/de/dennisguse/opentracks/AbstractTrackDeleteActivity.java
rgmf: null check if needed because the connection could be unbind before onStop (when deletion service finishes).
Look at line 120.

src/main/java/de/dennisguse/opentracks/AbstractTrackDeleteActivity.java
dennisguse: Creating an interface and implementing on the activity/fragment and pass it to the TrackRecordingService.
(side note: I am working on making TrackController a fragment and might need another callback there aka service disconnected).

rgmf:
@dennisguse Espresso test added to check that a track is delete when user select a item in the list view and click on delete action menu item.

PS/ This afternoon has been funny learning Espresso :smile:  I've never used it an I still have a lot of things to learn. I'll learn more about it because it looks important.

PS1/ I'll try to make an effort to add more tests to OT... It will worth. But it's so boring... :laughing:

dennisguse:
@rgmf Awesome!
i also promise to write some more tests (especially UI)!

rgmf:
It's ready for me but we have a problem with the vacuum when deleting. In a phone I have it takes too long to delete a track because the vacuum.

When that occurs then if you try to do something while deleting an ANR can showed (if that action imply a query) or other not desired behaviour. For example, if you try to show aggregated stats while deleting then they won't show until deleting is ended.

Because of this problem I've added a workaround: avoid all actions until deletion is finished and user sees a screen like this:

imagen

rgmf:
Another possible solution would be create a transparent Dialog and the UI would be like before but the lifecycle would be more complicated :thinking:

dennisguse:
@rgmf Could it be that we are loading in the UI thread and we shouldn't do that?
(I guess, you see the issue mainly in the loaderCallback in the two listActivities)

PS: It reliably gets more complicated than it initially looks like :)

rgmf:
> @rgmf Could it be that we are loading in the UI thread and we shouldn't do that?
> (I guess, you see the issue mainly in the loaderCallback in the two listActivities)
> 
> PS: It reliably gets more complicated than it initially looks like :)

That's right... We are querying database on UI thread. I'm going to re-think it. Because if a query is blocked until vacuum is done then UI shouldn't be freeze.

dennisguse:
@rgmf Do you know at where this is happening?
I just briefly checked the lists and those calls should be in a background thread using LoaderManager.

rgmf:
> @rgmf Do you know at where this is happening?
> I just briefly checked the lists and those calls should be in a background thread using LoaderManager.

It happends when deleting gets stuck in the vavuum and user opens a track. In TrackRecordedActivity: contentProviderUtils.getTrack gets freeze because is in the UI thread and other sql execution is running.

dennisguse:
My proposal: Let's live this (for now) aka let's merge without tackling these issues in this PR.

TL;DR 

I was worried that you would find such things.
I guess, then it even affects more places like MarkerDetailActivity as well as the *EditActivities (while loading and storing).

The question is how lilkely/bad is this?
From my code perspective, it should be okay (for now) as you cleaned up one part and just found some other hidden gems.
And for the user: it shouldn't happen that often, right?
* Deletion is not the daily business
* Depends on actual database size (mainly would affect power users)
* Depends on device performance

rgmf:
We should to do all database operations out of UI thread, right? We could do that with LiveData, what do you think? I can follow up this on this PR for completing it.

I know it's a lot of work but... I can do it... or not... will see :rofl: 

Deleting now is working, it's not hurry for this PR.

dennisguse:
@rgmf That would be awesome!
Do you need help?

PS: I have no trouble merging this PR even it may produce ANRs, as it cleans up other things and makes our live already easier.

Issue #745, opened 2021-05-05
While recording: ChartDiagram drawing is not correct when a new TrackPoint is added


dennisguse:
Likely introduced in https://github.com/OpenTracksApp/OpenTracks/commit/7be4ac436c3f6a41d57ca1aaa555088608ce9579

This happens while recording and being on a diagram tab.
If a new trackpoint is received, the diagram updates to this:

![Screenshot_20210430-173443_OpenTracks_(Debug)](https://user-images.githubusercontent.com/4819612/116725393-21d14d00-a9e2-11eb-95db-4e7f6bbd435f.png)

_Originally posted by @rgmf in https://github.com/OpenTracksApp/OpenTracks/issues/736#issuecomment-830214492_

dennisguse:
Introduced in: https://github.com/OpenTracksApp/OpenTracks/commit/887d4700d4f8fbf10563834e7498388a3213007e

TrackDataHub needs use the same TrackStatisticsUpdater for the whole track.
When receiving updates (while recording) a new one was created for every received TrackPoint resulting in wrong statistics and these weird diagrams. Did not affect StatisticsRecordingFragment as this get the TrackStatistics from the TrackRecordingService.

PR #754, opened 2021-05-09
ChartView: improve performance.

dennisguse:
While working on #745 I also re-read the rendering code in ChartView.
And here is an improvement: easier to read and should be slightly faster.


dennisguse:
Still works ;)

PR #755, opened 2021-05-10
TrackController is now ControllerFragment.

dennisguse:
TrackController did not handle it's own lifecycle and thus making TrackRecordingActivity and TrackListActivity slightly more complicated.

Now it does.

PS: On search the ControllerFragment is not hidding anymore.
I hope that is not too problematic.
The activities are now self-contained and the fragment only referenced via XML.
If we need this hiding feature we would need to use a FragmentManager and do it dynamically.

__REMOVED__  see #757 
Update: Search was not working on TrackListActivity as it only works if it is shown as an action menu (i.e., not hidden in the overflow).
Was not working for a very long time (July 2020)...
__REMOVED__


src/main/java/de/dennisguse/opentracks/ControllerFragment.java
```diff
@@ -0,0 +1,297 @@
...
+            handlerUpdateTotalTime.removeCallbacks(updateTotalTimeRunnable);
+        }
+        if (recordingStatus.isRecordingAndNotPaused()) {
+            handlerUpdateTotalTime.postDelayed(updateTotalTimeRunnable, UI_UPDATE_INTERVAL.toMillis());
+        }
+
+        viewBinding.controllerRecord.setImageResource(recordingStatus.isRecordingAndNotPaused() ? R.drawable.ic_button_pause : R.drawable.button_record);
+        viewBinding.controllerRecord.setContentDescription(getContext().getString(recordingStatus.isRecordingAndNotPaused() ? R.string.image_pause : R.string.image_record));
+
+        viewBinding.controllerStop.setEnabled(recordingStatus.isRecording());
```
rgmf, line 282: Stop button doesn't get disable until service is started. You have to start a recording and then stop it to see the stop button disabled on track list activity.

rgmf (commented):
How clear the code is now!!! Only a comment.

src/main/java/de/dennisguse/opentracks/ControllerFragment.java
dennisguse: Nice catch!
I forgot to initialize the UI - now it is happening!
Thanks!

PR #768, opened 2021-05-13
TrackRecordingService uses LiveData to transfer data.

dennisguse:
StatisticsRecordingFragment and ControllerFragment get the data via LiveData from the TrackRecordingService (event every second).
All computation is done within the TrackRecordingService.

Replaces pull from UI layer.

Not really tested yet: only by holding the phone out of a window ;)
Cycling sensor handling must be tested in detail.

Next step: simplify TrackDataHub interface as it is only used by Charts now.


rgmf (approved):
I tested it driving to work and all looks okay.

Issue #418, opened 2020-09-07
Can we get reproducible builds?


dennisguse:
Follow-up to the discussion on Mastodon https://chaos.social/@dennisguse/104818490105850432

Open tasks:
- [x] https://github.com/OpenTracksApp/opentracksapp.github.io/pull/12
~- [ ] Add build infos to OpenTracks (different splashscreen?)~
~- [ ] Add functionaliity to irreproducible build users aware of the reproducible build~
~- [ ] (fastlane) Add link to reproducible build into irreproducible description~

NOTE: both variants will stay on F-Droid and user can choose which they prefer.
No need to additionally communicate something more.

dennisguse:
Waiting for new builds on https://verification.f-droid.org

dennisguse:
A lot of details: https://f-droid.org/docs/Reproducible_Builds/

dennisguse:
OSMDashboard and also the Bahnhofsfoto App are now reproducible (thanks @pstorch)

Things that were done: 
1. https://github.com/RailwayStations/RSAndroidApp/commit/411426ebe54e32538aa0f0ac8285ef38a2a30dfb
2. https://gitlab.com/fdroid/fdroiddata/-/merge_requests/16486

For OpenTracks, we will likely publish reproducible builds under a new package name and keep both versions in the F-Droid store.
This allows current user to migrate whenever they want while new users can directly use the reproducible version.
Otherwise, we would force existing users to re-install incl. backup and restore - this has the tendency to be error prone...

dennisguse:
Could it be that the builds are already reproducible?
At least https://verification.f-droid.org reports that everything since versionCode 5683 could be verified.
Neat :D

For example: https://verification.f-droid.org/de.dennisguse.opentracks_5683.apk.json

dennisguse:
PR for F-Droid created: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/17921

However, F-Droid seems to have trouble with checkupdates by identifying the new package name.
In OpenTracks this is implemented using buildTypes (with applicationIdSuffix).
OSMDashboard uses product flavors - and this works.

dennisguse:
@pstorch I saw that multiple fastlane translations were used for OSMDashboard.
Do you know where to find the documentation for this?

I would love to change the short summary for the irreproducible flavor.

pstorch:
@dennisguse no real documentation, but I got this ticket back then to fix the fastlane structure:
[...]

PR #773, opened 2021-05-15
Launcher icon: use vector (layer-list for API21+ and adaptive-icon for API26+).

dennisguse:
The launcher icon used PNGs below API26.
We can use a layer-list vector icon instead of PNGs for such devices.


pstorch:
Does the new AS generate the adaptive launcher icons now like this? I have to try it. Back then when I created the adaptive launcher icons with AS, it was AS generating all the PNGs. I guess for older devices, but I don't have enough background knowledge about which version supports which graphic format.

dennisguse:
Me neither.
I just wanted to remove the PNGs, so they are not an issue for reproducible builds #418.
https://f-droid.org/de/docs/Reproducible_Builds/

In the emulator using API21 it works.

dennisguse:
Tested using Android 7.1 (LineageOS) and emulator with Android 5.0.
Looks okay :)

PR #793, opened 2021-05-24
Cleanup TrackDataHub

dennisguse:
See commtis.


rgmf (approved):
It works for me :tada: 

PS/ TrackDataHub will end up being removed? Is that your plan? I see only Intervals and Charts are using it.

dennisguse:
@rgmf that was my plan three month ago as it did not do too much and just connected different components.
With the EGM2008 implementation, TrackDataHub is now also acting as a cache for the raw data and also it computes the TrackStatistics.
And as the StatisticsRecordingFragment is not using it anymore, the complexity is greatly reduced already.
So, I guess my "plan" changed ;)

PR #794, opened 2021-05-25
Cleanup unused code.

rgmf:
Look at the commit (is simple).

That removed code is never execute. There aren't any component using TrackRecordedActivity.EXTRA_MARKER_ID...

Am I missing anything?


dennisguse:
I guess it was used to navigate from a marker to the track.
Don't know that we had this functionality :)
Guess it is not needed :)

dennisguse (approved):

Issue #779, opened 2021-05-17
All sql operations out of main thread


rgmf:
**Is your feature request related to a problem? Please describe.**
There are sql operations running on the main UI thread. In most cases there are not problems but they should be done in other threads.

**Describe the solution you'd like**
Every method in `ContentProviderUtils` return a `LiveData` and the rest of components can observe them to get the results. An example:

In ContentProviderUtils:
```java
    public MutableLiveData<Track> getTrack(@NonNull Track.Id trackId) {
        MutableLiveData<Track> liveData = new MutableLiveData<>();

        new Thread(() -> {
            try (Cursor cursor = getTrackCursor(TracksColumns._ID + "=?", new String[]{Long.toString(trackId.getId())}, null)) {
                if (cursor != null && cursor.moveToNext()) {
                    liveData.postValue(createTrack(cursor));
                } else {
                    liveData.postValue(null);
                }
            }
        }).start();

        return liveData;
    }
```

In StatisticsRecordedFragment:
```java
contentProviderUtils.getTrack(trackId).observe(this, track -> {
	if (track == null) {
	    Log.e(TAG, "track cannot be null");
	    getActivity().finish();
	    return;
	}
	
	sensorStatistics = contentProviderUtils.getSensorStats(trackId);
	
	boolean prefsChanged = this.track == null || (!this.track.getCategory().equals(track.getCategory()));
	this.track = track;
	if (prefsChanged) {
	    sharedPreferenceChangeListener.onSharedPreferenceChanged(sharedPreferences, getString(R.string.stats_rate_key));
	}
	
	loadTrackDescription(track);
	updateUI();
	updateSensorUI();
    });
```

dennisguse:
This also fixes:
```
D/StrictMode: StrictMode policy violation; ~duration=25 ms: android.os.strictmode.DiskReadViolation
        at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1556)
        at android.database.sqlite.SQLiteConnection.applyBlockGuardPolicy(SQLiteConnection.java:1138)
        at android.database.sqlite.SQLiteConnection.executeForCursorWindow(SQLiteConnection.java:939)
        at android.database.sqlite.SQLiteSession.executeForCursorWindow(SQLiteSession.java:838)
        at android.database.sqlite.SQLiteQuery.fillWindow(SQLiteQuery.java:62)
[...]

PR #801, opened 2021-05-29
Activity and Fragment classes do SQL operations out of UI thread. Fix…

rgmf:
**Describe the pull request**
I've created an static class inside ContentProviderUtils with methods that return LiveData so Activity/Fragment classes can observe them to do SQL operations out of UI thread.

This PR is only open for discussion because I don't know if it's the best approach. A lot of work to do yet (mainly tests).

**Link to the the issue**
#779 

**License agreement**


src/main/java/de/dennisguse/opentracks/io/file/exporter/ExportActivity.java
```diff
@@ -204,26 +203,30 @@ public void onBackPressed() {
     }
 
     private void initExport(int from) {
-        try (Cursor cursor = contentProviderUtils.getTrackCursor(null, null, TracksColumns._ID)) {
-            if (cursor == null) {
-                onExportEnded();
-                return;
-            }
+        ContentProviderUtils.RunOutUIThread.build(contentProviderUtils).getTrackCursor(null, null, TracksColumns._ID).observe(this, cursor -> {
```
dennisguse, line 206: somehow handing out a cursor feels dangerous in terms of closing it.

src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
```diff
@@ -802,4 +804,97 @@ public SensorStatistics getSensorStats(@NonNull Track.Id trackId) {
         }
         return sensorStatistics;
     }
+
+
+    public static class RunOutUIThread {
```
dennisguse, line 809: should only return LiveData<> instead of MutableLiveData<>

src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
```diff
@@ -351,10 +353,10 @@ public Marker createMarker(Cursor cursor) {
         return marker;
     }
 
-    public void deleteMarker(Context context, Marker.Id markerId) {
+    public int deleteMarker(Context context, Marker.Id markerId) {
```
dennisguse, line 356: Why not return Marker.Id?

src/main/java/de/dennisguse/opentracks/TrackEditActivity.java
```diff
@@ -65,13 +62,17 @@ protected void onCreate(Bundle bundle) {
         }
 
         contentProviderUtils = new ContentProviderUtils(this);
-        track = contentProviderUtils.getTrack(trackId);
-        if (track == null) {
-            Log.e(TAG, "No track for " + trackId.getId());
-            finish();
-            return;
-        }
+        ContentProviderUtils.RunOutUIThread.build(contentProviderUtils).getTrack(trackId).observe(this, track -> {
```
dennisguse, line 65: We need to do something about the syntax: 
```
ContentProviderUtils.RunOutUIThread.build(contentProviderUtils)
```

a contentProviderUtils that gets a contentProviderUtils?
Confusing ;)

dennisguse (commented):
Not sure yet; looks kind of okayish (not yet used to the amount of brackets).
Do we need to unsubscribe from the LiveData?

rgmf:
> Not sure yet; looks kind of okayish (not yet used to the amount of brackets).
> Do we need to unsubscribe from the LiveData?

According to the documentation:

"LiveData is an observable data holder class. Unlike a regular observable, LiveData is lifecycle-aware, meaning it respects the lifecycle of other app components, such as activities, fragments, or services. This awareness ensures LiveData only updates app component observers that are in an active lifecycle state"

So, I guess it's not needed to unsubscribe them.

src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
rgmf: It returns an integer (the number of rows deleted) because it's what delete in ContentResolver returns.

rgmf:
@dennisguse this PR sucks. The more I read it the more I don't like it :rofl: :roll_eyes: 
Also, I've read the database code from other projects and seem nobody does sql operations out of main thread.
It looks like a recommendation/demanding that nobody follow ¿?

dennisguse:
I am also not sure if this is important or just a real nice to have.
Then let's skip it for now and wait until somebody complains ;)

We have enough features that we can build - if we have time ><

dennisguse:
@rgmf
PS: anyhow nice that you worked on this!

PR #806, opened 2021-06-06
Speed and distance each only use one formatting rule.

### LINKED ISSUE #648: Speed / distance: rendering
We have two different ways to render the speed and the distance.
See `StringUtils.getSpeedParts()` and `StringUtils.formatSpeed()` as well as `StringUtils.getDistanceParts()` and `StringUtils.formatDistance()`

And both format the speed/distance differently, especially regarding the number of decimal digits.

dennisguse:
Fixes #648.

So far, we used separate rules for rendering distance and speed either (a) with unit or (b) without unit.
This lead to situations, were the decimal digits were different.
This is now unified.

PS: Also we get rid of a lot of copied translations for these cases.


rgmf (approved):
Everything looks good. Great for me.

PR #811, opened 2021-06-08
Fixed: using a wheel rotation sensor for speed and distance did not show on UI

### LINKED ISSUE #810: UI+sensor: speed and distance not shown properly
While using a speed sensor, the UI should show all overall distance and speed even if the GPS is off.

dennisguse:
While refactoring StatisticsRecordingFragment, the UI update was not implemented correctly, when using wheel rotation sensor.

PS: There also seems to be an issue computing the speed.
Distance seems fine. Not sure yet.

Fixes #810.

### DIFF
--- a/README_TESTED_SENSORS.md
+++ b/README_TESTED_SENSORS.md
@@ -23,8 +23,6 @@ However, often only one value is provided.
   * Has updatable firmware that requires an account for garmin.com
   * This sensor reports speed data as cadence.
     A workaround is in place.
-  * This sensor sometimes counts backwards (unknown software version).
-    A workaround is in place.
 
 * Wahoo Speed (Model: WFRPMSPD)
 
--- a/src/androidTest/java/de/dennisguse/opentracks/util/UintUtilsTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/util/UintUtilsTest.java
@@ -1,0 +1,23 @@
+package de.dennisguse.opentracks.util;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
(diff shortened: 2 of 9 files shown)


dennisguse:
One of these things - this PR is actually embarrassing.

TL;DR: the sensor decoding was broken and when it worked, the sensor was not connected and speed+distance taken from the GPS :facepalm:.

# Post mortem
While implementing the wheel rotation sensor for cycling, I overlooked that the data is an UINT32 and instead decoded using UINT16. I guess I copied the decoding from the cadence sensor instead of checking the documentation.
It was working as the sensor I tested with never got that many wheel rotation (I guess I tested for 2-3 hours). Then I got a sensor on my own and saw that it was not working properly. And while testing, I saw that the wheel count difference to the previous measurement was `-1`. I was testing at home (aka rotation the wheel of my bike while it was hanging at the wall). So, I _assumed_ that the sensor counts backwards, so I added `Math.abs(wheelDiff)`.
Then I went cycling and often speed+distance was working and often not.

Sometimes, I just saw speed around 5-8 kmh and the distance was way too little.
I disconnected from the sensor and re-connected and sometimes it worked.

So, I started digging (and discovered + fixed some other stuff). The issue was that the computation of the difference of wheel count (incl. accounting for overflow) was limited to UINT16. When trying this for anything above 65535, the diff computation returned `-1`. And as I added `Math.abs(wheelDiff)`, the wheel rotation was always 1. And the speed differences occurred due to timing differences (sometimes a notification is received after .5s and sometimes at .9).

Now we just throw an exception if the data is invalid. No time to mess around with missing error handling.
Now it works and I checked the documentation........

Issue #808, opened 2021-06-07
Intervals doesn't work properly with TrackDataHub


rgmf:
I'm improving Intervals using TrackStatisticsUpdater to compute the intervals and I found that Intervals are re-created every time the IntervalsFragment is resumed. It doesn't make sense because we're using ViewModel with LiveData that retain the intervals. That's because TrackDataHub is resumed every time IntervalsFragment is resumed and then TrackDataHub sends all track points again and again.

It's potentially inefficient because if you change ten times from stats tab to intervals tab then intervals are re-computed ten times, for instance.

I'm working on it (https://github.com/OpenTracksApp/OpenTracks/tree/intervalsWithStatsUpdater%23703)

dennisguse:
Sounds good! 
Does this only affect the IntervalFragment?
I am pretty sure that the ChartFragment is doing it something very similar...

rgmf:
> I am pretty sure that the ChartFragment is doing it something very similar...

Yes, I think so too.

I'll check it out too.

Issue #703, opened 2021-04-07
Intervals distance don't match with stats


rgmf:
**Describe the bug**
The addition of all intervals don't match with total distance in stats.

I realized of this while I was testing #701 but it's not related with it. I guess this issue is because intervals are not considering the new types of track points (TrackPoint.Type).

I think the addition of all intervals were nearly the same that total distance in stats, weren't they?

I'm going to check it out.

**To Reproduce**
1. Record a new track.
2. Look at stats distance and intervals distance: they not match.

dennisguse:
Don't know ;)

Why not use the TrackStastistics via TrackStatisticsUpdater (similar to the charts)?
Then all computation/presentation is done this combination and we are consistently correct/wrong :)

rgmf:
> Don't know ;)
> 
> Why not use the TrackStastistics via TrackStatisticsUpdater (similar to the charts)?
> Then all computation/presentation is done this combination and we are consistently correct/wrong :)

Yes, I was thinking on it.

It has should been done like this... I'm going to try to do it like this.

dennisguse:
Is this related with the IntervalStatistics vs TrackStatistics?
If yes, then this "imprecision" also affects the voice announcements, right?

rgmf:
> Is this related with the IntervalStatistics vs TrackStatistics?
> If yes, then this "imprecision" also affects the voice announcements, right?

Yes, I think so. We are using IntervalStatistics for getting the current interval data to announce them.

rgmf:
It's strange because the distance should be exactly the same and I'm fairly sure that one day it was :thinking: 

Anyway, it's a good moment to use TrackStatistics via TrackStatisticsUpdater :smile:

rgmf:
@dennisguse  I've fixed this... but:

I don't understand very well an important detail. I have a very little track:
- 0.95 km
- 46 seconds

![image](https://user-images.githubusercontent.com/4819612/114205416-915f9980-995a-11eb-8c85-01d8cb30c1ba.png)

But those 46 seconds should be 49 seconds:
- First trackPoint (from database) has a time of: 1617571177713
- Last trackPoint (from database) has a time of: 1617571227232

So `1617571227232 - 1617571177713 = 49519 millis`

Is there an explanation on this? Could this be a potencial bug?

PS/ All trackPoints type are 0 (unless the first one, -2, and the last one, 1).

dennisguse:
@rgmf That was fast!

[...]

PR #812, opened 2021-06-09
Intervals: performance improvement and TrackStatisticsUpdater to compute them

rgmf:
**TrackStatisticsUpdater to compute Intervals**
#703 is fixed in this PR.

**Performance improvement**
- When recording every time a new track point were gotten the intervals were created. So intervals were built again and again. Now, intervals add new track points to complete them.
- Related to the above TrackDataHub have been removed from IntervalsFragment. With TrackDataHub every time the fragment was resumed all TrackPoints were sent to the ViewModel but ViewModel.

**Work to be done after this PR**
- Improve announcement performance (last interval).
- Review ChartFragment (maybe the same issue that with the IntervalsFragment).

**Why draft PR**
Because I want to test them in a long route (next weekend). But this can be already reviewed.

**Link to the the issue**
#703 
#808 

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
@@ -745,6 +745,16 @@ public class ContentProviderUtils {
         return new TrackPointIterator(this, trackId, startTrackPointId);
     }
 
+    public List<TrackPoint> getTrackPoints(final Track.Id trackId, final TrackPoint.Id startTrackPointId) {
+        List<TrackPoint> trackPoints = new ArrayList<>();
+        try (TrackPointIterator trackPointIterator = getTrackPointLocationIterator(trackId, startTrackPointId)) {
+            while (trackPointIterator.hasNext()) {
+                trackPoints.add(trackPointIterator.next());
+            }
+        }
+        return trackPoints;
+    }
+
     @Deprecated
     private TrackPoint findTrackPointBy(String selection, String[] selectionArgs) {
         try (Cursor cursor = getTrackPointCursor(null, selection, selectionArgs, TrackPointsColumns._ID)) {
--- a/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
@@ -37,6 +37,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.StreamSupport;
 
 import de.dennisguse.opentracks.BuildConfig;
 import de.dennisguse.opentracks.content.data.Altitude;
@@ -743,16 +744,6 @@ public class ContentProviderUtils {
      */
     public TrackPointIterator getTrackPointLocationIterator(final Track.Id trackId, final TrackPoint.Id startTrackPointId) {
         return new TrackPointIterator(this, trackId, startTrackPointId);
-    }
-
-    public List<TrackPoint> getTrackPoints(final Track.Id trackId, final TrackPoint.Id startTrackPointId) {
-        List<TrackPoint> trackPoints = new ArrayList<>();
-        try (TrackPointIterator trackPointIterator = getTrackPointLocationIterator(trackId, startTrackPointId)) {
-            while (trackPointIterator.hasNext()) {
-                trackPoints.add(trackPointIterator.next());
-            }
-        }
-        return trackPoints;
     }
 
     @Deprecated
--- a/src/main/java/de/dennisguse/opentracks/fragments/IntervalsFragment.java
+++ b/src/main/java/de/dennisguse/opentracks/fragments/IntervalsFragment.java
@@ -140,11 +140,14 @@ public class IntervalsFragment extends Fragment implements TrackDataListener {
     @Override
     public void onResume() {
         super.onResume();
+
+        resumeTrackDataHub();
+
         sharedPreferences.registerOnSharedPreferenceChangeListener(sharedPreferenceChangeListener);
         sharedPreferenceChangeListener.onSharedPreferenceChanged(sharedPreferences, null);
+
         viewModel = new ViewModelProvider(getActivity()).get(IntervalStatisticsModel.class);
         loadIntervals();
-        resumeTrackDataHub();
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
```diff
@@ -745,6 +745,16 @@ public TrackPointIterator getTrackPointLocationIterator(final Track.Id trackId,
         return new TrackPointIterator(this, trackId, startTrackPointId);
     }
 
+    public List<TrackPoint> getTrackPoints(final Track.Id trackId, final TrackPoint.Id startTrackPointId) {
+        List<TrackPoint> trackPoints = new ArrayList<>();
+        try (TrackPointIterator trackPointIterator = getTrackPointLocationIterator(trackId, startTrackPointId)) {
+            while (trackPointIterator.hasNext()) {
+                trackPoints.add(trackPointIterator.next());
```
dennisguse, line 752: there might be a nicer way: https://www.baeldung.com/java-convert-iterator-to-list

(Nr. 3 or Nr. 4)

src/main/java/de/dennisguse/opentracks/fragments/IntervalsFragment.java
```diff
@@ -70,12 +64,14 @@
     /**
      * Creates an instance of this class.
      *
+     * @param trackId          track's id.
```
dennisguse, line 67: not really needed as it is clear by the type :)

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -9,44 +10,64 @@
...
+        this.minGPSDistance = minGPSDistance;
 
+        interval = new Interval();
+        lastInterval = new Interval();
+        intervalList = new ArrayList<>();
+        intervalList.add(lastInterval);
+    }
+
+    public void addTrackPoints(@NonNull List<TrackPoint> trackPoints) {
         if (trackPoints.size() == 0) {
```
dennisguse, line 38: trackPoints.isEmpty()

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -9,44 +10,64 @@
...
-                interval.time = interval.time.plus(Duration.between(prevTrackPoint.getTime(), trackPoint.getTime()));
-                interval.gain_m += trackPoint.hasAltitudeGain() ? trackPoint.getAltitudeGain() : 0;
-                interval.loss_m += trackPoint.hasAltitudeLoss() ? trackPoint.getAltitudeLoss() : 0;
+                double adjustFactor = distanceInterval.dividedBy(interval.distance);
+                Interval adjustedInterval = new Interval(interval);
+                adjustedInterval.adjust(adjustFactor);
 
-                if (interval.distance.greaterThan(distanceInterval)) {
-                    Interval adjustedInterval = new Interval(interval, distanceInterval.dividedBy(interval.distance));
+                intervalList.remove(intervalList.size() - 1);
```
dennisguse, line 53: this is replacing the last element?
Then `intervalList.set(intervalList.size() - 1, adjustedInterval)` can be used.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -111,5 +144,23 @@ public float getGain_m() {
         public float getLoss_m() {
             return loss_m;
         }
+
+        private void update(TrackStatistics trackStatistics, @Nullable TrackPoint lastTrackPoint) {
```
dennisguse, line 148: void add()?

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -111,5 +144,23 @@ public float getGain_m() {
...
+            distance = distance.plus(trackStatistics.getTotalDistance());
+            time = time.plus(trackStatistics.getTotalTime());
+            gain_m = trackStatistics.hasTotalAltitudeGain() ? trackStatistics.getTotalAltitudeGain() : gain_m;
+            loss_m = trackStatistics.hasTotalAltitudeLoss() ? trackStatistics.getTotalAltitudeLoss() : loss_m;
+            if (lastTrackPoint != null) {
+                gain_m = lastTrackPoint.hasAltitudeGain() ? gain_m - lastTrackPoint.getAltitudeGain() : gain_m;
```
dennisguse, line 154: should that not be part of the trackstatistics already?
Or was the lastTrackPoint not yet added?

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
```diff
@@ -1,81 +1,123 @@
...
     }
 
-    public void add(TrackPoint trackPoint) {
-        synchronized (trackPoints) {
-            trackPoints.add(trackPoint);
-        }
+    private void loadIntervalStatistics(Track.Id trackId) {
+        executor.execute(() -> {
+            ContentProviderUtils contentProviderUtils = new ContentProviderUtils(getApplication());
+            List<TrackPoint> trackPoints = contentProviderUtils.getTrackPoints(trackId, lastTrackPointId);
```
dennisguse, line 86: That is a memory intensive operation; it would be better to pass use the cursor/TrackPointIterator and initialize the intervalStatistics from there.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
```diff
@@ -103,6 +145,10 @@ public Distance getDistance(boolean metricUnits) {
                     .multipliedBy(multiplier);
         }
 
+        public boolean equals(IntervalOption intervalOption) {
+            return intervalOption != null && this.multiplier == intervalOption.multiplier;
```
dennisguse, line 149: the method name is misleading as equals is otherwise used to check for equality with another object.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
```diff
@@ -1,81 +1,123 @@
...
+            intervalStatistics = new IntervalStatistics(distanceInterval, minGPSDistance);
+
+            loadIntervalStatistics(trackId);
         }
-    }
 
-    private void loadIntervalStatistics() {
-        IntervalStatistics intervalStatistics = new IntervalStatistics(trackPoints, distanceInterval);
-        intervalsLiveData.postValue(intervalStatistics.getIntervalList());
+        registerTrackPointsObserver(trackId);
```
dennisguse, line 78: Why move this into a separate method?
(I was somehow confused as I did not realized this method should only be called once).

dennisguse (commented):
Looks kind of nice!

Did not test, but it looks way better than the TrackDataHub use.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
rgmf: That's because every time a new Interval is added trackStatisticsUpdater is reset to begin from the last track point, so last track point is added two times, so we need to substract the gain and the loss in that case.

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatisticsModel.java
rgmf: You're right.
I'm surprised how quickly you see all these stuffs!!! :open_mouth:

rgmf:
Ready for me @dennisguse 

What do you think?

dennisguse:
@rgmf If it is good for, it is good for me :sunglasses: 

PS: Okay to create a release afterwards?

rgmf:
> PS: Okay to create a release afterwards?

Good moment for that :+1:

PR #818, opened 2021-06-16
EspressoCategoryUnitTest needs to set metricUnits.

dennisguse:
i am running the tests only in emulators as some of them delete data or modify preferences.

And my emulator starts with Locale.US, but this tests assumes that we are a non-US device.
So, we set it ;)


rgmf (approved):
Good for me!

PR #831, opened 2021-06-21
Bugfix: TrackRecordingService needs to send data from lastTrackPoint …

### LINKED ISSUE #828: [v3.18.4] distance, time, speed, and geolocation values flicker between two values very fast (and shown data is incorrect)
I am currently trying to add Moofit Cadence + Speed sensors to the app but it did not work so far. Currently, only the cadence value is displayed correctly. The speed value is way too low (like 3 km/h). Yesterday I saw that a new version of the app was available via f-droid. Version number is v3.18.4-9234da49.
After download I tried to pair both sensors again but it still doesn't work. Additionally, now I have rapidly changing values for the fields mentioned in the title.
Like "current speed": 25km/h and then 0km/h the next second. Geolocation is first correct and in the next moment only "-" is displayed. Maybe a problem with my GPS? Anyway, I tried to resolve it by restarting app and device but it did not change the outcome.

Maybe anyone has some input on either of the two issues I am facing with the app. Otherwise it is a real great app!

**Technical information**
 - Device: [Oneplus 6t]
 - OS: [LineageOS Fajita]
 - Version [v3.18.4]

dennisguse:
…in it's 1s updates.

This also removes UI updates if a GPS measurement is received as 1s update should be sufficient.

Fixes #828.

BUT: elevation is shown in WGS84 as the EGM2008 transformation is happening in the TrackDataHub.
TrackRecordingService needs to do the same.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/content/data/Altitude.java
+++ b/src/main/java/de/dennisguse/opentracks/content/data/Altitude.java
@@ -1,5 +1,7 @@
 package de.dennisguse.opentracks.content.data;
 
+import de.dennisguse.opentracks.R;
+
 public abstract class Altitude {
 
     private final double altitude_m;
@@ -12,12 +14,18 @@ public abstract class Altitude {
         return altitude_m;
     }
 
+    public abstract int getLabelId();
+
     public static class WGS84 extends Altitude {
 
         private WGS84(double altitude_m) {
             super(altitude_m);
         }
 
+        @Override
+        public int getLabelId() {
+            return R.string.wgs84;
+        }
 
         public static Altitude of(double altitude_m) {
             return new WGS84(altitude_m);
@@ -30,6 +38,10 @@ public abstract class Altitude {
             super(altitude_m);
         }
 
+        @Override
+        public int getLabelId() {
+            return R.string.egm2008;
+        }
 
         public static Altitude of(double altitude_m) {
             return new EGM2008(altitude_m);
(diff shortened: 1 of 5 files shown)


dennisguse:
@montagsmodell @sebastianha Can you give this a spin?

sebastianha:
I am sorry but I still don't have a development device. I need definitely get one to help out!

dennisguse:
@sebastianha Don't worry :)

montagsmodell:
Not sure how I can help you. I dont know how to create the apk to test it with my phone. if you send me instructions I will try though!

dennisguse:
@montagsmodell don't worry for now.
I will create in the following days.

rgmf:
@dennisguse speed stays at 0.0km/h all time.

dennisguse:
@rgmf Great catch!
I tested with a speed sensor and that worked :)
Now speed is working :)

rgmf:
@dennisguse works in the emulator but I suspect the speed is jumping between actual speed and a high speed (I see how the speed flickering between 10km/h and 40km/h in a route when I was running around 12km/h). Let me test it in a real route.

dennisguse:
@rgmf the speed is now most recent data (aka not averaged via TrackStatistics).
So, you might see this. Let me check if should take it from the TrackStatistics instead.

dennisguse:
@rgmf I just checked v3.16.0.
There we already used the speed from the lastTrackPoint, so it was not smoothed using TrackStatistics.
The smoothed speed is only used for avg speed and max speed.
So, we are back now to the "original" implementation (not sure if this is the perfect solution).

rgmf:
@dennisguse anyway let me test it tomorrow morning because I cannot trust emulator.

rgmf:
@dennisguse speed works like a charm. Sorry for worry you. I was emulator fault ;P

PS/ tested on bike and walking with muy dog :D

rgmf (approved):
Great!!

dennisguse:
Will do a final test after work and then release v3.18.5 :partying_face:

PR #834, opened 2021-06-22
Fix typo in German translation

pgeorgi:
It's Geschwindigkeit, not Geshwindigkeit.


dennisguse:
@pgeorgi Thanks!

You may also use Weblate: https://weblate.bubu1.eu/engage/OpenTracks/
Before every release (and every now and then), I pull the translations from there.

Issue #820, opened 2021-06-16
Customizable Recording Layout


rgmf:
**Description**
We have a grid to show data fields in the recording layout:

![imagen](https://user-images.githubusercontent.com/4819612/122245056-ac540a00-cec5-11eb-8436-d33cd1218ba4.png)

From settings, users will been able to select the data fields they want to see and the order in the recording layout grid.

To do that we can use a “Preference action” that start a new Activity. (https://developer.android.com/guide/topics/ui/settings/customize-your-settings#actions). This Activity will use a RecyclerView with the drag and drop feature to offer to the user an UI to set their preferences (what to see and the order in the grid the fields data will be seen).

![imagen](https://user-images.githubusercontent.com/4819612/122245347-f210d280-cec5-11eb-9960-22cb7832a9be.png)

**Preferences Changes**
For every data field we save two pairs of key-value in shared preferences:
    • One of them to save the position in the grid.
    • The other one to set (true or false) the visibility.

So, we’ll define all these preferences keys in the values/settings.xml file:

stats_total_time_data_position_key
stats_total_time_data_show_key

stats_distance_data_position_key
stats_distance_data_show_key

stats_moving_total_time_position_key
stats_moving_total_time_show_key

stats_speed_position_key
stats_speed_position_show

stats_avg_speed_position_key
stats_avg_speed_show_key

stats_avg_moving_speed_position_key
stats_avg_moving_speed_show_key

stats_max_speed_position_key
stats_max_speed_show_key

stats_altitude_position_key
stats_altitude_show_key

stats_gain_position_key
stats_gain_show_key

stats_loss_position_key
stats_loss_show_key

stats_sensors_position_key
stats_sensors_show_key

Sensors clarification: sensors will add all user sensors connected. For simplification we offer to the user to show all or none of sensors.

**SettingsActivity Changes**
Add an option that will open a new activity ( https://developer.android.com/guide/topics/ui/settings/customize-your-settings#actions)

**StatisticsRecordingFragment Changes**
The layout will be a RecyclerView with a grid (GridLayoutManager). It’ll be like current stats sensors recycler view.

This new RecyclerView will get preferences information and put all stats where they should be according to the user preferences.

**Issues related**
https://github.com/OpenTracksApp/OpenTracks/issues/659
https://github.com/OpenTracksApp/OpenTracks/issues/302

**Discussion**
https://github.com/OpenTracksApp/OpenTracks/discussions/813

[...]

Issue #659, opened 2021-03-17
Make shown data configurable


sebastianha:
I would really like to see that I can configure the data shown on the screens.

**Example 1 -  Main screen:** Currently time and km is shown. I would be handy to also show average speed or heart rate information here. There should be enough space in the second row.

**Example 2 - Interval screen:** Currently on the right there is elevation. It would be handy to change this to e.g. heart rate information (avg/max) or other data.

**Example 3 - Graph screens:** It could be handy to hide shown graphs, e.g. if you are not interested in elevation.

rgmf:
Interesting idea. I like it.

We could apply something similar than we are doing now for sensor data (dynamically loaded).

I don't like icon on recording either... it's not that important there.

Issue #302, opened 2020-06-24
Show data source for speed and distance


dennisguse:
Show from were we get the speed (Bluetooth sensor vs GPS).

Suggested here: https://github.com/OpenTracksApp/OpenTracks/issues/278#issuecomment-648845402

PR #842, opened 2021-06-24
[Wait for v3.19.1] Settings activity for customize order and visible stats fields.

rgmf:
**What do you think**
@dennisguse @edern-appere @vsevjednom-cz what  would you change? is this what you expected? You all have been participated on this so I mention you because you can participate on the final result...

**Why draft?**
Because I want to use it in real routes (real testing).

**Describe the pull request**
Added new setting so the users can customize theirs recording layout:
- Order of data fields.
- Visible of data fields.

Recording layout is now a RecyclerView with a set of CardViews that results in a grid layout with selected visible fields (in the order selected).

The grid layout has 2 columns in portrait mode and 3 columns in landscape mode (they can be configured on values/dimens).

**Internal data structure detail**
I added the following string not translatable to handle all settings stuff surronding layout customization:
```xml
<string name="stats_custom_layout_order_key" translatable="false">statsCustomLayoutOrderKey</string>
    <string-array name="stats_custom_layout_order_values">
        <item>@string/stats_total_time</item>
        <item>@string/stats_moving_time</item>
        <item>@string/stats_distance</item>
        <item>@string/stats_speed</item>
        <item>@string/stats_average_moving_speed</item>
        <item>@string/stats_average_speed</item>
        <item>@string/stats_max_speed</item>
        <item>@string/stats_pace</item>
        <item>@string/stats_average_moving_pace</item>
        <item>@string/stats_average_pace</item>
        <item>@string/stats_fastest_pace</item>
        <item>@string/stats_altitude</item>
        <item>@string/stats_gain</item>
        <item>@string/stats_loss</item>
        <item>@string/stats_latitude</item>
        <item>@string/stats_longitude</item>
        <item>@string/stats_sensors_heart_rate</item>
        <item>@string/stats_sensors_cadence</item>
        <item>@string/stats_sensors_power</item>
    </string-array>

    <string name="stats_custom_layout_visible_key" translatable="false">statsCustomLayoutVisibleKey</string>
    <string-array name="stats_custom_layout_visible_values">
        <item>@string/stats_moving_time</item>
        <item>@string/stats_distance</item>
        <item>@string/stats_speed</item>
        <item>@string/stats_average_moving_speed</item>
        <item>@string/stats_max_speed</item>
        <item>@string/stats_pace</item>
        <item>@string/stats_average_moving_pace</item>
        <item>@string/stats_fastest_pace</item>
        <item>@string/stats_altitude</item>
        <item>@string/stats_gain</item>
        <item>@string/stats_loss</item>
    </string-array>
```

PreferenceUtils transform all this to a LinkedHashMap whose keys are the stats description strings and values are true or false depending on visibility field (see PreferenceUtils code).

**Default Layout**
When user install OT they'll see this layout:

image

**Handicap**
For people who uses OT for running and biking, for example, there is a handicap: they'll have to configure the layout every time they change sport. We'll have to follow this for adding layout profiles.

**For reviewers**
- Pay attention to StatsBuilder class (I don't know if it's the best approach).
- Some new classes names are too long ¿?

**Video that show how it works**
I don't know why but I cannot add here a video. I added it in Mastodon:
https://mastodon.online/web/statuses/106466455091087333

**Link to the the issue**
#659 
#302 
#820


dennisguse:
Love it.

Thoughts:
* I really liked the old UI [design-wise)
   * different font sizes / sub-groups highlight colors
   * I don't see the benefits of the borders
* What about vertical layout? Can we just put three boxes in one row?
   Would be way better than now...
* (whishlist) Can we get a preview screen in the settings?

rgmf:
>     * I really liked the old UI [design-wise)

Okay, easy one ;)

>       
>       * different font sizes / sub-groups highlight colors
>       * I don't see the benefits of the borders

Tow kinds of data: primary and secondary, right? I'll give it a spin...

> 
>     * What about vertical layout? Can we just put three boxes in one row?
>       Would be way better than now...

We can offer to the user the decision of how many boxes per row they want... Three by default, what do you think?

> 
>     * (whishlist) Can we get a preview screen in the settings?

Something like this?

image

dennisguse:
That preview got me thinking....
... can we not just show the UI with the boxes (as it would be while recording) and make those movable?
But then: how would we show things that are hidden?

:shrug:

rgmf:
> That preview got me thinking....
> ... can we not just show the UI with the boxes (as it would be while recording) and make those movable?
> But then: how would we show things that are hidden?
> 
> 🤷

Okay, that sounds good. About the hidden data I'll think something...

vsevjednom-cz:
The preview of recording looks amazing! It's cool!
I hope the same layout will be for viewing the old activities - as I wrote in # 836.

edern-appere:
@rgmf the system you built is really nice and seems to work really well, from what I can see on the screenshot and the video on Mastodon. Nice work, it really goes into an interesting direction !!

Like @dennisguse mentionned, having the possibility to have primary/secondary style for data could be a great option... unless it brings too much customization possibilities and designing a home screen means having a PhD.

[whishlist mode on] It would be great to be able to save a custom screen for each sport, because you don't look for the same kind of data depending on the sport you are practicing. If you have to change the screen settings, each time you change sport, you won't be able to open OT and just push on the red button which is why OT is so simple to use[whishlist mode off]. 

As a side note, @rgmf I don't know which activity you are practicing with OT, but if it's cycling, 137 Km/h is pretty fast :rofl:

rgmf:
> Like @dennisguse mentionned, having the possibility to have primary/secondary style for data could be a great option... unless it brings too much customization possibilities and designing a home screen means having a PhD.

Don't worry, something will occurs to us... It would be nice and OT always have had primary and secondary texts.

> [whishlist mode on] It would be great to be able to save a custom screen for each sport, because you don't look for the same kind of data depending on the sport you are practicing. If you have to change the screen settings, each time you change sport, you won't be able to open OT and just push on the red button which is why OT is so simple to use[whishlist mode off].

Yes, @dennisguse already pointed to that and I think it's a demanding feature. I'm really interesting on that because I run and ride the bike, so I'd need two layouts.

> As a side note, @rgmf I don't know which activity you are practicing with OT, but if it's cycling, 137 Km/h is pretty fast rofl

:rofl: :rofl: :rofl: 

I speed up x5 the emulator speed so I can create long tracks.

PS/ Thank you so much for participate on this.

rgmf:
@dennisguse @vsevjednom-cz new draft design for discussion: https://mastodon.online/web/statuses/106483908320029922

I've added two RecyclerView:
- The first one with preview.
- The second one with stats fields user can select (click) to add to the layout.

When user click on an item of the first layout it's moved to the second and the other way around. On that way user can add and remove stats items.

Also, you can see an Spinner to select the number of columns per row: 1, 2 or 3.

I'm thinking on a long click to select on a contextual menu primary or secondary field. What do you all think?

dennisguse:
Sounds like a good plan!

PS: We could even implement multiple UI fields for the same data like:
* Gain
* Loss
* Gain + Loss (together)
[ there are probably better examples, but it is too warm at the moment >< ]

rgmf:
@dennisguse we need a database table to complete this feature. I cannot figuring out other way to add several layouts profiles and primary/secondary stats features. Because using preferences will end up difficult to manage. How to keep all this information? Are there other ways?

dennisguse:
Requirements:
* we need to be able to handle changes while running
* we need to be able to upgrade the data structures (new fields etc)

Data: 
* elements have an order (or some kind of relationship between each other)
* elements might have additional configuration data

i see the following options:
* sharedPreferences
* database/sqlite
* file in private storage

(gut feeling: sharedPreferences with CSV like data structure e.g., `cycling;speed;\nrunning;distance`)

PS: I am not too sure if we should implement UIs per activity already.
From a users perspective, this might also imply different sensors (like different bicycles with speed sensors each).
All this adds a lot of complexity (also the configuration).
TL;DR: first make UI configurable and add per activity afterwards

rgmf:
@dennisguse okay, all that makes sense. So I'm going to make UI configurable first. Thank you ;)

rgmf:
Almost done. Summing up:
- [x] Custom recording layout setting.
- [x] Number of columns per row setting (1, 2 or 3. By default it uses 3).
- [x] Importance of fields: primary or secondary.
- [x] Add tests (including espresso tests).
- [x] Show altitude and latitude/longitude settings removed because they are not needed anymore.
- [ ] Add profiles so users can set several recording layouts.

If you don't have time to install it you can see how it works in this video (password: **OpenTracks842**): https://ppp.woelkli.com/s/ZC2NgmiJ4pXrt2x

@dennisguse it's okay for me. If your review is okay, then I'll do the last step: add profiles.

dennisguse:
Pretty nice!

Stuff I noticed while testing for 5min (was just too late):
* lat/lng could be one field?
* sensor data fields could show the sensor name (incl. speed)

* (not sure) Settings Customize recording layout and UI theme should be in the same section? I guess, we need to restructure the settings slightly...
* Settings: preferred rate is still used or not? It does not affect the recording layout, right?
* Is there an option to reset the layout (without resetting all preferences)?
* Sensor data fields can be enabled even if no sensor is configured
  (when configuring a sensor, we could add the field at the end)

* Default layout row count is 3; I guess 2 would be better (closer to the original)
* Default layout: can we somehow make it close to the current (aka non-configurable?) one?

rgmf:
>     * lat/lng could be one field?

I have to look for this. I don't know if a RecyclerView with a GridLayoutManager can have views with span cells. I mean, views that expand more than one cell.

Anyway, I'm sure I'll find a solution to this.

> 
>     * sensor data fields could show the sensor name (incl. speed)

I have only a heart rate monitor and I can see the hexadecimal value. I'll check it out.

> 
>     * (not sure) Settings Customize recording layout and UI theme should be in the same section? I guess, we need to restructure the settings slightly...

Now that you mention it, yes, it makes more sense.

> 
>     * Settings: preferred rate is still used or not? It does not affect the recording layout, right?

Yes, it's used in IntervalsFragment and ChartFragment so show speed or pace, depending on the activity type.

> 
>     * Is there an option to reset the layout (without resetting all preferences)?

Not, there isn't. We can add another setting to reset the layout. It can be really useful.

> 
>     * Sensor data fields can be enabled even if no sensor is configured
>       (when configuring a sensor, we could add the field at the end)

Don't you think it's useful for the user see that X sensor has an unknown value (aka "-")? That way s-he know that sensor is not connected (maybe because s-he forgot to link it).

I have not any sensor except a heart rate monitor. So, let me know what you prefer and I'll do it like that. It doesn't matter to me. What you decide will be fine.

> 
>     * Default layout row count is 3; I guess 2 would be better (closer to the original)

Okay ;)

> 
>     * Default layout: can we somehow make it close to the current (aka non-configurable?) one?

Okay ;)

dennisguse:
@rgmf Regarding the "show sensor UI field, when sensor is added in settings": not sure whatever you prefer.
I was thinking that some users might not be aware that you have to add this UI field if you want to see it.
I don't mind either as I know :)

PS: For distance and speed/pace - it would be awesome if the information could be shown were this data came from.
Especially using wheel-based cycling sensor (which I can recommend as then GPS may be turned off - saves a lot of battery) sometimes the phone does not connect to the sensor and so far the user has no option to check this easily.
I usually check this at the first traffic light were I am stopping as the speed is reducing faster if you stop (when using a sensor).

PS: Really awesome UI.

dennisguse:
@rgmf We have an issue with the CI - for some reason API29 is running into timeouts.
So, don't be upset if it is reported as failing.
I guess, it is not our fault :sunglasses:

rgmf:
> @rgmf Regarding the "show sensor UI field, when sensor is added in settings": not sure whatever you prefer.
> I was thinking that some users might not be aware that you have to add this UI field if you want to see it.
> I don't mind either as I know :)

We can add cadence/hr/power fields to custom layout automatically when user link their sensors at the end if these sensors fields are not added.

> 
> PS: For distance and speed/pace - it would be awesome if the information could be shown were this data came from.

Really interesting :+1: 

> 
> PS: Really awesome UI.

Thank you so much!!!

rgmf:
Update status:
- [x] Custom recording layout setting.
- [x] Number of columns per row setting (1, 2 or 3. By default it uses 3).
- [x] Importance of fields: primary or secondary.
- [x] Add tests (including espresso tests).
- [x] Show altitude and latitude/longitude settings removed because they are not needed anymore.
- [x] Reset layout option in settings.
- [x] Custom layout by default is similar to the original one.
- [x] Coordinate (latitude/longitude) in an only field. There are short fields and long fields (the latter span one row).
- [x] Sensors connected always are shown (if user didn't select them in custom layout settings then they are shown at the end).
- [x] Speed/Pace with secondary text indicating the data's source: sensor or GPS.
- [ ] Add profiles so users can set several recording layouts.

@dennisguse below all sensor data should be appear the name or address (for instance: `sensorDataSet.getHeartRate().getSensorNameOrAddress()`).

@dennisguse Could you give an spin to this? I think it's okay. But every time I request your review I think it's okay and it seldom is  :roll_eyes: :roll_eyes: :rofl:

PS/ Don't review the code, I'd like to do it before you do. Only the logic and functionality.

dennisguse:
@rgmf Still love it - no complains :sunglasses: 
Only nitpick: I guess, the sensor info should be below the value as otherwise the two values in one row are not aligned (for HR and speed).

And that's my layout for tomorrow :partying_face: 
Screenshot_20210717-185116_OpenTracks__Debug_

rgmf:
@dennisguse in my last push, info labels (primary and secondary if any) are below the values. I think it's the best option for two reasons:
- Values are aligned.
- Values are the more interesting info when user is recording, right?

PS1/ Your "complains" makes OT better :wink:

PS2/ Your layout is similar to my layout :D

dennisguse:
One tiny bug: after the connection to a sensor is established, the name is resolved.
And then the name should be shown instead of it's address.
This works in v3.18.6 - but not in the screenshot I pasted :paintbrush:

rgmf:
> One tiny bug: after the connection to a sensor is established, the name is resolved.
> And then the name should be shown instead of it's address.
> This works in v3.18.6 - but not in the screenshot I pasted paintbrush

Heart rate sensor is working... I cannot understand :(

imagen

Is your heart rate sensor working?

dennisguse:
@rgmf Regarding HR: did not test again, but your screenshot looks good.
Might have been an old build - sorry for the confusion.

dennisguse:
@rgmf i did reading (although you said I should not)....

Some things:
* stats_custom_layout_all_fields_value/stats_custom_layout_fields_default_value should use non-localized strings
  For example, I change the language of my phone and the layout does not work anymore.
  Or somebody changes the translation (even the English one).
* is there a concept, how we update the layout preference?
  (like adding profiles, removing fields etc)
* Layout.profile ==Track.category? 
* we should check the license before integrating external code: `RecyclerViewAssertions` 

Just nitpiicking:
* for usage of `Layout` in testing: it is easier to implement `equals()` than test every property individually
* layout.fields should be an orderedSet, right? (aka unique)
* StatsData is not (yet) that easy to understand (like`getDescMain()`).
* StatsData.isLong needs another name: I thought it was checking the data type aka`instanceof Long` => completeRow?
* StatsData has too many constructors
* StatsData would be nicer as a value object (aka no setter, properties final)
* `switch` should work for String (e.g., for StatsDataBuilder): https://docs.oracle.com/javase/8/docs/technotes/guides/language/strings-switch.html
* move now unused strings for preference (like `stats_show_coordinate_key` into settings_deprecated.xml)
  We have no migration concept and those might still be present on some installations; the right approach would be to delete this data on the phones (don't know how to do this properly)

rgmf:
> @rgmf i did reading (although you said I should not)....

It was for your mental health :rofl: but thank you anyway because your comments are very useful.

>     * stats_custom_layout_all_fields_value/stats_custom_layout_fields_default_value should use non-localized strings

Yes, you are right (BTW: something similar occur with track's categories).

> 
>     * is there a concept, how we update the layout preference?

Every time a developer want to change something s-he has to change stats_custom_layout_all_fields_values/stats_custom_alyout_fields_default_value and add an `case/if` in `StatsDataBuilder`. That was the idea.

> 
>     * Layout.profile ==Track.category?

No, it isn't. The profile will be the name user will decide for the profile when profiles will be developed.

> 
>     * we should check the license before integrating external code: `RecyclerViewAssertions`

Okay, I'm going to look into this.

> 
> 
> Just nitpiicking:
> 
>     * for usage of `Layout` in testing: it is easier to implement `equals()` than test every property individually
> 
>     * layout.fields should be an orderedSet, right? (aka unique)
> 
>     * StatsData is not (yet) that easy to understand (like`getDescMain()`).
> 
>     * StatsData.isLong needs another name: I thought it was checking the data type aka`instanceof Long` => completeRow?
> 
>     * StatsData has too many constructors
> 
>     * StatsData would be nicer as a value object (aka no setter, properties final)
> 
>     * `switch` should work for String (e.g., for StatsDataBuilder): https://docs.oracle.com/javase/8/docs/technotes/guides/language/strings-switch.html
> 
>     * move now unused strings for preference (like `stats_show_coordinate_key` into settings_deprecated.xml)
>       We have no migration concept and those might still be present on some installations; the right approach would be to delete this data on the phones (don't know how to do this properly)

I'll check them out. Thank you ;)

dennisguse:
We store the track category as localized string in the database? :exploding_head:   
But it works, right :) might affect the lookup of icons...

Regarding `* is there a concept, how we update the layout preference?`:
I meant the actual upgrade procedure for the devices - something like a database upgrade.
For example, we decide that a new data model is needed (something else than CSV) as it is more flexible/easier/whatever.
We then need to have some data schema upgrade operation for this preference.
How would we do this?

rgmf:
> Regarding `* is there a concept, how we update the layout preference?`:
> I meant the actual upgrade procedure for the devices - something like a database upgrade.
> For example, we decide that a new data model is needed (something else than CSV) as it is more flexible/easier/whatever.
> We then need to have some data schema upgrade operation for this preference.
> How would we do this?

Aaahhhh okay, since I'm rewritten all model I'll do it thinking on this. Let me do it and I'll describe to you the upgrade operation.

PS/ Anyway, according to your experience, do you have any advice for me? :blush: If don't so I'll set up one.

dennisguse:
From the top of my head (order is meaningless)
* upgrade on read
* upgrade on application upgrade (something like [or even in] CustomSQLiteOpenHelper?)
* just set it to default if it cannot be decoded (not really user friendly; but simple)

Question is: where to store the "preference structure version"?
* first line of structure
* separate preference
* just use the database version (and do the upgrade via CustomSQLiteOpenHelper)?

Real upgrade requires that OpenTracks (still) can decode the data structure; so if we change it too often, it get's a maintenance burden.

Why I stumbled upon this: per profile BLE sensors - like different bicycles or running shoes etc.
And then we need to put this info into a very similar (or the same) data structure.
Not sure ;)

rgmf:
> From the top of my head (order is meaningless)
> 
>     * upgrade on read
> 
>     * upgrade on application upgrade (something like [or even in] CustomSQLiteOpenHelper?)
> 
>     * just set it to default if it cannot be decoded (not really user friendly; but simple)
> 
> 
> Question is: where to store the "preference structure version"?
> 
>     * first line of structure
> 
>     * separate preference
> 
>     * just use the database version (and do the upgrade via CustomSQLiteOpenHelper)?
> 
> 
> Real upgrade requires that OpenTracks (still) can decode the data structure; so if we change it too often, it get's a maintenance burden.
> 
> Why I stumbled upon this: per profile BLE sensors - like different bicycles or running shoes etc.
> And then we need to put this info into a very similar (or the same) data structure.
> Not sure ;)

In my last commit you can see my proposal. Now, we can upgrade preferences in general and layout preference in particular.

dennisguse:
@rgmf Due to JDK11 upgrade and the necessary changes in the F-Droid configuration, I would like to release v3.19.0 today/ tomorrow.
We need to wait for a failed build, then adjust the F-Droid config and then the actual build. :man_shrugging: 
More details: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/9462#note_639333417

And then I propose, your awesome work (together with #888) as v3.20.0 in the upcoming weeks?
PS: Would be cool to rebase this PR to main as there are LOTS of changes and it makes testing easier.

dennisguse:
@rgmf How can I help you with this PR?

rgmf:
> @rgmf How can I help you with this PR?

I've been traveling for two weeks and I'll get home tomorrow. Next week I'll complete it.

Anyway, it would be great if you check StatsDataBuilder out. I'm not sure if it can be get simpler.

PS/ I took the laptop with me because my plan was working on it these weeks but my plan failed 🤣

rgmf:
@dennisguse it's ready for me.

Please, review my last two commits: is valid what I did?

PS/ StatsAdapter has to use a layout for recording (with label below) and another one for recorded (with label above and bolder).

PS2/ No hurry ;)

src/main/java/de/dennisguse/opentracks/fragments/StatisticsRecordingFragment.java
```diff
@@ -109,17 +102,28 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
-        sensorsRecyclerView.setLayoutManager(new GridLayoutManager(getContext(), 2));
-        sensorsRecyclerView.setAdapter(sensorsAdapter);
+        RecyclerView recyclerView = viewBinding.statsRecyclerView;
+        recyclerView.addItemDecoration(new DividerItemDecoration(getContext(), RecyclerView.VERTICAL));
+        final int numColumns = PreferencesUtils.getLayoutColumns(sharedPreferences, getContext());
+        gridLayoutManager = new GridLayoutManager(getContext(), numColumns);
+        gridLayoutManager.setSpanSizeLookup(new GridLayoutManager.SpanSizeLookup() {
+            @Override
+            public int getSpanSize(int position) {
+                if (statsAdapter.getItemViewType(position) == StatsAdapter.VIEW_TYPE_LONG) {
```
dennisguse, line 119: Can we define a method for this check on statsAdapter?

src/main/java/de/dennisguse/opentracks/settings/SettingsCustomLayoutActivity.java
```diff
@@ -0,0 +1,168 @@
...
+            @Override
+            public void onSwiped(@NonNull RecyclerView.ViewHolder viewHolder, int direction) {
+            }
+        };
+
+        ItemTouchHelper itemTouchHelper = new ItemTouchHelper(simpleCallback);
+        itemTouchHelper.attachToRecyclerView(recyclerViewVisible);
+
+        // Spinner with items per row.
+        ArrayAdapter<Integer> spinnerAdapter = new ArrayAdapter<>(this, android.R.layout.simple_spinner_dropdown_item, new Integer[]{1, 2, 3});
```
dennisguse, line 82: values should be defined in XML

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -39,12 +48,34 @@
 
     private final static String TAG = PreferencesUtils.class.getSimpleName();
 
+    private static final int PREFERENCES_VERSION = 1;
```
dennisguse, line 51: interesting :)

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -39,12 +48,34 @@
...
+        SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+        int lastVersion = getInt(sharedPreferences, context, R.string.prefs_last_version_key, 0);
+        if (PREFERENCES_VERSION > lastVersion) {
+            onUpgrade(sharedPreferences, context, PREFERENCES_VERSION);
+            setInt(sharedPreferences, context, R.string.prefs_last_version_key, PREFERENCES_VERSION);
+        }
+        return sharedPreferences;
+    }
+
+    private static void onUpgrade(SharedPreferences sharedPreferences, Context context, int toVersion) {
```
dennisguse, line 67: can we extract this into a separate file?
e.g., PreferenceOpenHelper (similar to CustomSQLiteOpenHelper)?

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +345,76 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+            Field field;
+            int counter = 0;
+
+            do {
+                field = resource.getField(key + "_" + counter);
+                typedArrays.add(context.getResources().obtainTypedArray(field.getInt(null)));
+                counter++;
+            } while (field != null);
+        } catch (Exception e) {
+            e.printStackTrace();
```
dennisguse, line 371: use log.

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +345,76 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+        try {
+            Class<R.array> resource = R.array.class;
+            Field field;
+            int counter = 0;
+
+            do {
+                field = resource.getField(key + "_" + counter);
+                typedArrays.add(context.getResources().obtainTypedArray(field.getInt(null)));
+                counter++;
+            } while (field != null);
```
dennisguse, line 369: always true according to AndroidStudio

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +345,76 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+
+    public static void setLayoutColumns(SharedPreferences sharedPreferences, Context context, int columns) {
+        setInt(sharedPreferences, context, R.string.stats_custom_layout_columns_key, columns);
+    }
+
+    private static List<TypedArray> getMultiTypedArray(Context context, String key) {
+        List<TypedArray> typedArrays = new ArrayList<>();
+
+        try {
+            Class<R.array> resource = R.array.class;
```
dennisguse, line 361: This looks complicated; what is happening here?
This feels like reflection...

Is there some nicer approach?
Using switch statement?

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +345,76 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+            Field field;
+            int counter = 0;
+
+            do {
+                field = resource.getField(key + "_" + counter);
```
dennisguse, line 366: Not a fan of manually referencing XML entries: this will prevent us using static code analyzer to find unused resources.

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +345,76 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+        return layout;
+    }
+
+    public static void setCustomLayout(SharedPreferences sharedPreferences, Context context, Layout layout) {
+        List<Layout.Field> fields = layout.getFields();
+        if (fields.size() == 0) {
+            return;
+        }
+
+        String csv = layout.getProfile() + ";" + fields.stream().map(field -> field.getKey() + "," + (field.isVisible() ? "1" : "0") + "," + (field.isPrimary() ? "1" : "0")).collect(Collectors.joining(";")) + ";";
```
dennisguse, line 403: something like: `profile;fieldName,visible,primary;` might be more readable/ maintainable

may we use `public static String CSV_SEPARATOR = ";"`?

dennisguse: and can use create a method in Field to parse/ write it's data?

src/main/java/de/dennisguse/opentracks/util/StatsUtils.java
```diff
@@ -0,0 +1,31 @@
...
+        } else if (statTitle.equals(context.getString(R.string.stats_distance)) || statTitle.equals(context.getString(R.string.stats_speed)) || statTitle.equals(context.getString(R.string.stats_average_moving_speed)) || statTitle.equals(context.getString(R.string.stats_average_speed)) || statTitle.equals(context.getString(R.string.stats_max_speed))) {
+            return context.getString(R.string.stats_empty_value_float);
+        } else {
+            return context.getString(R.string.stats_empty_value_integer);
+        }
+    }
+
+    public static Layout filterVisible(Layout layout, boolean visible) {
+        Layout result = new Layout(layout.getProfile());
+        for (Layout.Field field : layout.getFields()) {
```
dennisguse, line 24: stream/ filter?

src/main/java/de/dennisguse/opentracks/util/StringUtils.java
```diff
@@ -130,6 +130,16 @@ public static String formatCoordinate(double coordinate) {
...
     }
 
+    /**
+     * Formats a complete coordinate (latitude, longitude)
+     *
+     * @param latitude
+     * @param longitude
+     */
+    public static String formatCoordinate(double latitude, double longitude) {
+        return Location.convert(latitude, Location.FORMAT_DEGREES) + COORDINATE_DEGREE + ", " + Location.convert(longitude, Location.FORMAT_DEGREES) + COORDINATE_DEGREE;
```
dennisguse, line 140: We may need a string.xml entry for this as the order might be different per lanuage.

dennisguse: (incl. the degree sign)

src/main/java/de/dennisguse/opentracks/viewmodels/StatsData.java
```diff
@@ -0,0 +1,56 @@
+package de.dennisguse.opentracks.viewmodels;
+
+
+import android.util.Pair;
+
+import androidx.annotation.NonNull;
+
+public class StatsData {
```
dennisguse, line 8: feels like overlapping with Field

src/main/res/layout/stats_recorded_item.xml
```diff
@@ -1,42 +1,47 @@
...
     android:layout_height="wrap_content"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
-    xmlns:tools="http://schemas.android.com/tools">
+    android:paddingBottom="16dp"
+    android:paddingTop="16dp">
 
     <TextView
-        android:id="@+id/stats_sensor_label"
-        style="@style/StatsLargeLabel"
+        android:id="@+id/stats_desc_main"
```
dennisguse, line 12: not a fan of abbrevations: `desc_main` could be `descriptionMain`

dennisguse (approved):
@rgmf If you think it is ready, then it is ready.
Looks good to me, but did not do too many testing.

My proposal: wait until v3.19.1 is released (incl. the two bugs) and then merge and release v3.20.0
And then we will see what we can improve :sunglasses: 

PS: Great job!

dennisguse:
@vsevjednom-cz If you have some time, it would be great if you could do some testing on this PR.
It is a really cool feature!

vsevjednom-cz:
@dennisguse Of course I could. Can I clone complete code somewhere? I don't want to edit real working code 🤪.

rgmf:
> @dennisguse Of course I could. Can I clone complete code somewhere? I don't want to edit real working code 🤪.

If you have a local copy then you can follow the instruction you can see clicking the link below this page that says "command line instruction".

vsevjednom-cz:
> If you have a local copy then you can follow the instruction you can see clicking the link below this page that says "command line instruction".

I don't see the link named "command line instruction" 😥. I tried search it in source code of this page too - without success. Do I need some permissions for it? I would like to do some testing tomorrow.

rgmf:
> I don't see the link named "command line instruction" disappointed_relieved. I tried search it in source code of this page too - without success. Do I need some permissions for it? I would like to do some testing tomorrow.

Sorry, you are right. Follow these steps to get this branch:
```
git fetch origin
git checkout -b customRecordingLayout#820 origin/customRecordingLayout#820
```

After that you'll move to this branch and then you'll be able to test it from Android Studio.

Don't hesitate to ask anything else ;)

src/main/java/de/dennisguse/opentracks/adapters/StatsAdapter.java
```diff
@@ -0,0 +1,139 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import de.dennisguse.opentracks.R;
+import de.dennisguse.opentracks.viewmodels.StatsData;
+
+public class StatsAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder> {
+
+    public static final int VIEW_TYPE_SHORT = 1;
```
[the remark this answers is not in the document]
rgmf: :+1: Created an enum.

src/main/java/de/dennisguse/opentracks/content/data/Layout.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public List<Field> getFields() {
+         return fields;
+    }
+
+    public String getProfile() {
+        return profile;
+    }
+
+    public static class Field {
```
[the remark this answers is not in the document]
rgmf: I think we can't because user can modify the visibility/primary from Settings.

vsevjednom-cz:
Wow, it works, thanks.

I'm testing. I've got two questions:
1. How can I save a customized layout? It saves only a number fields per row.
2. Change of fields per row looks good in tracking. But I don't see any change in older activities.
🙃

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
rgmf: when getField will fail because it doesn't exists then field will be null.

rgmf:
>     1. How can I save a customized layout? It saves only a number fields per row.

You have to go to `Settings -> Customize your recording layout` and make changes in the preview layout: clicking on the fields and moving it where you want.


>     2. Change of fields per row looks good in tracking. But I don't see any change in older activities.
>        upside_down_face

You only can customize the recording layout... When you finish your recording/tracking then you simply see all the fields in the recorded layout.

vsevjednom-cz:
> You have to go to `Settings -> Customize your recording layout` and make changes in the preview layout: clicking on the fields and moving it where you want.

I did it like you write but I don't see any changes when I open `Customize your recording layout` again.

rgmf:
> > You have to go to `Settings -> Customize your recording layout` and make changes in the preview layout: clicking on the fields and moving it where you want.
> 
> I did it like you write but I don't see any changes when I open `Customize your recording layout` again.

Could you make and share an screenshot what you see in Customize your recording layout screen?

vsevjednom-cz:
> Could you make and share an screenshot what you see in Customize your recording layout screen?

First open of `Customize your recording layout`:
Screenshot_20210808_165923_de dennisguse opentracks debug
I did change (move `moving time`):
Screenshot_20210808_165952_de dennisguse opentracks debug
But second open of `Customize your recording layout` looks like first:
Screenshot_20210808_170002_de dennisguse opentracks debug

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
rgmf: It gets all array resources. I mean, `resource` has all array resources.

My first approach was using this:
`int resId = context.getResources().getIdentifier(key + "_" + counter, "array", context.getPackageName());`

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
rgmf: I did it like this so developers can add more fields through settings.xml without the needed of update anything else.

So, if you want to add a new file you only have to add this new array to the settings.xml resources:

```xml
<string-array name="stats_custom_layout_fields_default_value_18">
        <item>@string/stats_custom_layout_newfield_key</item>
        <item>0,0,0</item>
    </string-array>
```

Appending the next number.

With this approach we cannot to use a constant resource's name.

rgmf:
@vsevjednom-cz you are right. It's not working now. Looks like the last changes broke something.

I'm going to check it out and fix it.

Thank you!!!!

rgmf:
@vsevjednom-cz I added a bug this morning. It's now fixed.

Please, pull changes and test it again if you can/want:

```
git pull origin customRecordingLayout#820
```

Thank you so much. Very appreciate ;)

vsevjednom-cz:
> test it again

It works amazing right now 🙃.

src/main/java/de/dennisguse/opentracks/viewmodels/StatsData.java
rgmf: Yes, I refactored it.

Now we have:
- StatisticsField that has the following attributes: key, title, isVisible, isPrimary and isWide.
- Layout with several StatisticsFields.
- StatisticsData that has the following attributes: StatisticsField, value, unit, (optional) description.

Basically:

```
class StatisticsField {
    - key
    - title
    - isVisible
    - isPrimary
    - isWide
}

class StatisticsData {
    - value
    - unit
    - description
}
StatisticsData o-- StatisticsField

class Layout {
    - profile
}
Layout "1" o-- "many" StatisticsField
```

PS/ I'm going to update `data-structure.txt` with all this information.

rgmf:
@vsevjednom-cz I think it's ready. I've refactored some parts and fixed some minor bugs. With my last commit I think it's okay. I tell you this because you can pull and test/use it ;)

vsevjednom-cz:
It's more better right now 📱. English names of fields in `Customize your recording layout` are in Czech language now 🇨🇿.

So I can translate these sentences for you:
`Customize your recording layout` = `Přizpůsobit vzhled měření` 🇨🇿
`Data dataFields per row` (text `Data data` is maybe some error?!) = `Počet polí na řádek` 🇨🇿
Hope it helps 🙃.

rgmf:
> `Data dataFields per row` (text `Data data` is maybe some error?!) = `Počet polí na řádek` czech_republic

Yes, it's an error I just fixed. Thank you ;)

dennisguse:
@rgmf Anything I can do to help?

rgmf:
> @rgmf Anything I can do to help?

It's almost finished. I still have to move preferences version to a separated class, aka PreferencesOpenHelper.

So, don't worry, I'll do it tonight.

dennisguse:
Rebased to main for testing.

src/main/java/de/dennisguse/opentracks/content/data/DataField.java
```diff
@@ -0,0 +1,97 @@
...
+    public void togglePrimary() {
+        isPrimary = !isPrimary;
+    }
+
+    public boolean isWide() {
+        return isWide;
+    }
+
+    public String toCsv() {
+        String visible = this.isVisible ? "1" : "0";
```
dennisguse, line 51: Can we move "1" etc into some static String and give it a name?

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -123,6 +137,12 @@ public static void setString(SharedPreferences sharedPreferences, Context contex
         editor.apply();
     }
 
+    public static void setInt(SharedPreferences sharedPreferences, Context context, int keyId, int value) {
```
dennisguse, line 140: Preferred to be non public.
Can we move the using class into this package?

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +328,78 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+        setInt(sharedPreferences, context, R.string.stats_custom_layout_columns_key, columns);
+    }
+
+    private static List<TypedArray> getMultiTypedArray(Context context, String key) {
+        List<TypedArray> typedArrays = new ArrayList<>();
+
+        try {
+            Class<R.array> resource = R.array.class;
+            Field field;
+            int counter = 0;
```
dennisguse, line 346: that is the index? aka i?

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
dennisguse: And replace printStackTrace

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +328,78 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+            Log.e(TAG, e.getMessage());
+        }
+
+        return typedArrays;
+    }
+
+    @SuppressLint("ResourceType")
+    static String buildDefaultLayout(Context context) {
+        List<TypedArray> fieldsArrays = getMultiTypedArray(context, "stats_custom_layout_fields_default_value");
+        return context.getString(R.string.default_activity_default) + ";"
```
dennisguse, line 364: use defined constants for ";" and ","

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
dennisguse: Works for me: still the parsing code and the generating for this string should be rather close (if there is a generating code).
And use a constant for: "_"

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +328,78 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+                + fieldsArrays.stream().map(i -> i.getString(0) + "," + i.getString(1)).collect(Collectors.joining(";"))
+                + ";";
+    }
+
+    public static Layout getCustomLayout(SharedPreferences sharedPreferences, Context context) {
+        String csvCustomLayout = getString(sharedPreferences, context, R.string.stats_custom_layout_fields_key, buildDefaultLayout(context));
+        List<String> csvParts = Arrays.asList(csvCustomLayout.split(";"));
+        Layout layout = new Layout(csvParts.get(0));
+        for (int i = 1; i < csvParts.size(); i++) {
+            List<String> fieldParts = Arrays.asList(csvParts.get(i).split(","));
```
dennisguse, line 374: why asList? Looks like it could also be used as an array

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +328,78 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+            layout.addField(fieldParts.get(0), DataField.getTitleByKey(context, fieldParts.get(0)), fieldParts.get(1).equals("1"), fieldParts.get(2).equals("1"), fieldParts.get(0).equals(context.getString(R.string.stats_custom_layout_coordinates_key)));
```
dennisguse, line 375: use of constant for "1" etc.

src/main/java/de/dennisguse/opentracks/util/PreferencesUtils.java
```diff
@@ -318,4 +328,78 @@ public static void setDefaultExportDirectoryUri(SharedPreferences sharedPreferen
...
+        }
+
+        return layout;
+    }
+
+    public static void setCustomLayout(SharedPreferences sharedPreferences, Context context, Layout layout) {
+        List<DataField> fields = layout.getFields();
+        if (fields.size() == 0) {
```
dennisguse, line 383: isEmpty

src/main/java/de/dennisguse/opentracks/util/StatisticsUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.stream.Collectors;
+
+import de.dennisguse.opentracks.R;
+import de.dennisguse.opentracks.content.data.Layout;
+
+public class StatisticsUtils {
+
+    public static String emptyValue(@NonNull Context context, @NonNull String statTitle) {
+        if (statTitle.equals(context.getString(R.string.stats_custom_layout_total_time_key)) || statTitle.equals(context.getString(R.string.stats_custom_layout_moving_time_key)) || statTitle.equals(context.getString(R.string.stats_custom_layout_pace_key)) || statTitle.equals(context.getString(R.string.stats_custom_layout_average_moving_pace_key)) || statTitle.equals(context.getString(R.string.stats_custom_layout_average_pace_key)) || statTitle.equals(context.getString(R.string.stats_custom_layout_fastest_pace_key))) {
```
dennisguse, line 15: PreferencestUtils.isKey?

dennisguse (approved):
@rgmf Just tiny stuff.

I will do some testing tomorrow (on a tour), but I don't expect any big findings - usually your testing is very precise.

Long story short: Let's merge this on the weekend and create a new release!

PS: Great job!
I really love that it now shows the source for speed/distance - really helpful for me as I now have sensors for cycling and running.

dennisguse:
Did some testing today - looks good :)
@rgmf Have fun shipping it!

Issue #791, opened 2021-05-23
Workaround for #619: no barometric sensor => no gain/loss in IntervalStatistics


dennisguse:
intervalStatistics show 0m for gain/loss if no barometric sensor is present.
Can we replace this by a `-` or hide it even until we get #619?

PR #846, opened 2021-06-25
Workaround for no barometric sensor: intervals shows dashes (-) for gain and loss.

rgmf:
**Describe the pull request**
The simple solution: show dashes.
Hiding data require some communication strategy between Intervals Adapter and Interval Fragment because IntervalFragment have to hide the header and Adapter the values... Does it worth? I don't think so.

**Link to the the issue**
#791

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/adapters/IntervalStatisticsAdapter.java
+++ b/src/main/java/de/dennisguse/opentracks/adapters/IntervalStatisticsAdapter.java
@@ -59,8 +59,9 @@ public class IntervalStatisticsAdapter extends RecyclerView.Adapter<RecyclerView
 
         viewHolder.rate.setText(StringUtils.formatSpeed(context, interval.getSpeed(), metricUnits, isReportSpeed));
 
-        viewHolder.gain.setText(StringUtils.formatDistance(context, Distance.of(interval.getGain_m()), metricUnits));
-        viewHolder.loss.setText(StringUtils.formatDistance(context, Distance.of(interval.getLoss_m()), metricUnits));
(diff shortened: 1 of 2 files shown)


src/main/java/de/dennisguse/opentracks/adapters/IntervalStatisticsAdapter.java
```diff
@@ -59,8 +59,9 @@ public void onBindViewHolder(@NonNull RecyclerView.ViewHolder holder, int positi
 
         viewHolder.rate.setText(StringUtils.formatSpeed(context, interval.getSpeed(), metricUnits, isReportSpeed));
 
-        viewHolder.gain.setText(StringUtils.formatDistance(context, Distance.of(interval.getGain_m()), metricUnits));
-        viewHolder.loss.setText(StringUtils.formatDistance(context, Distance.of(interval.getLoss_m()), metricUnits));
+        viewHolder.gain.setText(interval.hasGain() ? StringUtils.formatDistance(context, Distance.of(interval.getGain_m()), metricUnits) : "-");
```
dennisguse, line 62: Should be `R.strings.value_unknown` aka `<string name="value_unknown" translatable="false">-</string>`

dennisguse (changes requested):
Still looks good

src/main/java/de/dennisguse/opentracks/adapters/IntervalStatisticsAdapter.java
rgmf: Done... I didn't realize of that.

dennisguse (approved):
+1

PR #848, opened 2021-06-26
Remove KML/KMZ lvariants

dennisguse:
Before OSMDashboard existed, I added the functionality to only share a limited set of geo data with other applications (i.e., only coordinates; no timestamps; no markers; no track description).

All of this should not be necessary anymore (due to the existence OSMDashboard) - it is a privacy respecting map application.
No need to protect the user too much.

Thus, I propose to remove this.


rgmf (approved):
This works for my tracks :+1:

pstorch:
Sorry, I had to send my laptop to the service. No I have limited access to the remaining shared family laptop.

dennisguse:
@pstorch Don't worry - it was more the question if we still need this functionality.

And sending in one's computer sucks hard.
Hope it won't take too long.

dennisguse:
I removed the renaming of the TrackFileFormat enum entries as those are directly used in the settings.
So, we have to keep them for now; I also added a comment for this.

Issue #847, opened 2021-06-25
FATAL EXCEPTION: AsyncTask 🤪


vsevjednom-cz:
**Describe the bug**
I found this bug accidentally 🙃. It's not an important error. But it can happen sometimes 🤖.

**To Reproduce**
1. Start tracking.
2. Go to the list.
3. Mark current activity.
4. Tap to the menu and to the `Delete` (`menu_delete`).
5. Confirm it.
6. `FATAL EXCEPTION`

**Technical information**
 - Device: Honor KSA-LX9
 - OS: Android 9.1.0

```
E/pentracks.debu: [qarth_debug:]  get PatchStore::createDisableExceptionQarthFile method fail.
E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #3
    Process: de.dennisguse.opentracks.debug, PID: 27042
    java.lang.RuntimeException: An error occurred while executing doInBackground()
        at android.os.AsyncTask$3.done(AsyncTask.java:355)
        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
        at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
        at java.util.concurrent.FutureTask.run(FutureTask.java:271)
        at java.util.concurrent.ThreadPoolExecutor.processTask(ThreadPoolExecutor.java:1187)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1152)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:784)
     Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.UUID de.dennisguse.opentracks.content.data.Track.getUuid()' on a null object reference
        at de.dennisguse.opentracks.util.ExportUtils.getExportFileNameForTrack(ExportUtils.java:104)
        at de.dennisguse.opentracks.util.ExportUtils.getExportDocumentFileUri(ExportUtils.java:92)
        at de.dennisguse.opentracks.util.ExportUtils.exportTrack(ExportUtils.java:44)
        at de.dennisguse.opentracks.io.file.exporter.ExportService.onHandleWork(ExportService.java:51)
        at androidx.core.app.JobIntentService$CommandProcessor.doInBackground(JobIntentService.java:391)
[...]

PR #849, opened 2021-06-27
Race condition fixed: deleting a recording track with post-execution on

rgmf:
**The problem**
#847 is a race condition because while is stopping the recording track is deleting it. Meanwhile post-export track is launching too.

I could reproduce it with some smartphones/emulators but not in others.

**My proposal**
I think the best option here is not delete the recording track but shows a Toast: "The recording track cannot be deleted. Stop it first."

**Link to the the issue**
#847

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -478,6 +478,7 @@ limitations under the License.
     <string name="track_delete_multiple_confirm_message">The selected tracks and their markers will be permanently deleted from the device.</string>
     <string name="track_delete_progress_message">Deleting&#8230;</string>
     <string name="track_delete_progress">%1$d/%2$d</string>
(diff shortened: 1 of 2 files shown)


src/main/res/values/strings.xml
```diff
@@ -478,6 +478,7 @@ limitations under the License.
     <string name="track_delete_multiple_confirm_message">The selected tracks and their markers will be permanently deleted from the device.</string>
     <string name="track_delete_progress_message">Deleting&#8230;</string>
     <string name="track_delete_progress">%1$d/%2$d</string>
+    <string name="track_delete_not_recording">The recording track cannot be deleted. Stop it first.</string>
```
dennisguse, line 481: `This track cannot be deleted as it is currently recorded.`

dennisguse (approved):

PR #851, opened 2021-06-27
Refactoring tests: unify assertion of TrackPoints

dennisguse:
The assertions of TrackPoints was implemented several times and done differently.
This makes testing complicated and thus should be improved.

Wait for #848 and then rebase.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
+++ b/src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
@@ -164,9 +164,11 @@ public class IntervalStatistics {
             time = time.plus(trackStatistics.getTotalTime());
             gain_m = trackStatistics.hasTotalAltitudeGain() ? trackStatistics.getTotalAltitudeGain() : gain_m;
             loss_m = trackStatistics.hasTotalAltitudeLoss() ? trackStatistics.getTotalAltitudeLoss() : loss_m;
-            if (lastTrackPoint != null) {
-                gain_m = lastTrackPoint.hasAltitudeGain() ? gain_m - lastTrackPoint.getAltitudeGain() : gain_m;
-                loss_m = lastTrackPoint.hasAltitudeLoss() ? loss_m - lastTrackPoint.getAltitudeLoss() : loss_m;
+            if (gain_m != null && lastTrackPoint != null && lastTrackPoint.hasAltitudeGain()) {
+                gain_m = gain_m - lastTrackPoint.getAltitudeGain();
+            }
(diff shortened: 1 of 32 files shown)


dennisguse:
After merging #848; rebased onto main.
* [ ] Speed is not yet verified (expectation needs to be set)
* [ ] Refactor `TrackPointAssert.expect()` to use TrackPoint constructor directly
* [ ] Check GPX test failure

src/main/java/de/dennisguse/opentracks/viewmodels/IntervalStatistics.java
```diff
@@ -164,9 +164,11 @@ private void add(TrackStatistics trackStatistics, @Nullable TrackPoint lastTrack
             time = time.plus(trackStatistics.getTotalTime());
             gain_m = trackStatistics.hasTotalAltitudeGain() ? trackStatistics.getTotalAltitudeGain() : gain_m;
             loss_m = trackStatistics.hasTotalAltitudeLoss() ? trackStatistics.getTotalAltitudeLoss() : loss_m;
-            if (lastTrackPoint != null) {
-                gain_m = lastTrackPoint.hasAltitudeGain() ? gain_m - lastTrackPoint.getAltitudeGain() : gain_m;
-                loss_m = lastTrackPoint.hasAltitudeLoss() ? loss_m - lastTrackPoint.getAltitudeLoss() : loss_m;
+            if (gain_m != null && lastTrackPoint != null && lastTrackPoint.hasAltitudeGain()) {
```
rgmf, line 167: We have `hasGain` method, why not use it?

```if (hasGain() && lastTrackPoint != null && lastTrackPoint.hasAltitudeGain())```?

rgmf (commented):
I think there's a little bug: gain and loss are always unknown "-" (in both track recording layout and intervals fragment).

dennisguse:
@rgmf Thanks - I will check this today!
And I am so happy that the tests are finally working again!

dennisguse:
@rgmf I even did a `=` instead of  `+=`....
Now it's working: after first pressure measurement 0m is shown and gain/loss is updated via temporary TrackStatistics every second.
Now, we can test in an elevator (what I did)...

dennisguse:
@rgmf I should stop working on this PR - it just gets really big.
I just wanted to refactor the tests... :sun_with_face:

rgmf (changes requested):
👍

dennisguse:
@rgmf I am slightly confused - what needs to be changed?

rgmf:
> @rgmf I am slightly confused - what needs to be changed?

If gain/loss unknown issue is fixed then it's ready for me ;)

rgmf (approved):

Issue #850, opened 2021-06-27
[Android/GPS driver] Orthometric heights (H) being reported as geoid EGM2008 appear skewed


vkbellis:
**Describe the bug**
Orthometric heights (H) being reported as geoid EGM2008 appear to actually be ellipsoid heights (h).

**To Reproduce**
Examine the displayed height on Stats_screen under the heading 'ELEVATION', or any of the vertex elevations in the exported GPX or KML files, and then compare with actual surveyed values.

**Technical information**
 - Device: Samsung Galaxy SM-A716U
 - OS: Android 11
 - Platform 30

The field test was conducted holding my smartphone face up, approximately 1.1 meter above and parallel with the ground during the entirety of the 4.83 km walk. The exported line feature (GPX file) was then brought into Global Mapper in a new workspace, geographic projection, WGS84, and then individual vertices were extracted on their own layer. Next, a new layer with the copied point features was created during a performed calculation to adjust the GPX heights by 1.1 meters to more closely align with the Earth surface. This adjusted layer was exported in LAS format so as to enable Global Mapper to then be able to profile these points.

For context, a digital elevation model (DEM) predicated upon LiDAR flown 2010-2011 was translated from NAD83 into the WGS84(G1762) frame and orthometric heights (H) were translated from Geoid09 into EGM2008 using Geographic Calculator 2020 SP1.

The known benchmark (PID: [BBBF03](https://www.ngs.noaa.gov/OPUS/getDatasheet.jsp?PID=BBBF03&ts=16006162416&style=modern)) was along the chosen path of my walk and when abreast of it, a screenshot of OpenTracks was captured while walking; I didn't stop walking for any of the screen captures.

The published orthometric (H) and ellipsoid (h) heights for BBBF03 were transformed into EGM2008 ortho height (H) and WGS84(G1762) ellipsoid height (h) for comparison with its nearest vertex from the adjusted GPX point feature height, i.e., vertex number 399. Its value is within reasonable range of being consistent with the WGS84(G1762) ellipsoid height (h).

![Horizon view BBBF03](https://user-images.githubusercontent.com/82275968/123554976-d1137180-d750-11eb-8d9f-b4064058314f.jpg)

![Ellipsoid-(GPX)-heights-(h)-profiled](https://user-images.githubusercontent.com/82275968/123554984-dc669d00-d750-11eb-844f-c21bc8e5ccab.png)

[...]

PR #865, opened 2021-07-02
Add option to set required distance before recording a trackpoint to 0m.

dennisguse:
Useful for testing like attitude at a static position (for EGM2008 verification).
The same behavior can actually be achieved by using a BLE sensor.

PS: The whole concept of this setting must be reworked; especially regarding altitude one might be interested in the most recent and not only the one that was stored in the db (or just store it). TrackStatistics might be different.

Used for: https://github.com/OpenTracksApp/OpenTracks/issues/850

### DIFF
--- a/src/main/res/values/settings.xml
+++ b/src/main/res/values/settings.xml
@@ -109,6 +109,7 @@
     <string name="recording_distance_interval_key" translatable="false">recordingDistanceInterval</string>
     <string name="recording_distance_interval_default" translatable="false">10</string>
     <string-array name="recording_distance_interval_values">
+        <item>0</item>
         <item>1</item>
         <item>2</item>
         <item>5</item>


vkbellis:
Currently, in OpenTracks v3.18.5-823dc1c12- (4176), it is not possible to set the Recording distance interval to 0.  

Recording a track with a stationary device for diagnostic testing purposes is one approach, but making it a multi-segmented line feature results in unnecessary line segments. What you're really looking for are point features (waypoints) being recorded. Each point, and for that matter, each vertex in the line feature when recording tracks, should have clearly labeled and recorded attributes for:

-  ellipsoid height (h)
-  orthometric height (H)
-  geoidal separation (N)

It would also be quite helpful if some remark/ note/ meta data could be included at the top of the file indicating the reference frame and geoid model, for example, WGS84 and EGM2008

The diagnostic tool ought to allow: 

- static observations for a user defined time interval
- entry of the height of the antenna reference point (ARP)
- accounting for the ARP by subtracting its value
- recording each fixed point
- displaying mean when completed along with
- standard deviations, HRMS and VRMS

dennisguse:
@vkbellis this will not create new segments (this is handled by another property).
OpenTracks completely ignores all locations within this configured distance to previous - setting it to 0 will store all locations (while keeping segments as before).


Adding the other data is kind of complicated - especially considering import/export.
OpenTracks was not built for this (for GPS debugging) and that makes complicated.
And we do not have access NMEA sentences as Android uses different interfaces to get GPS (parsed by Android) vs raw NMEA data.
And using both at the same time is also a nightmare beyond OpenTracks (at least with the current resources).
Doable: yes.
Likely: no.


We even do not export the EGM2008 corrected data as it is only an estimate that can be created via post-processing (and the KML/GPX standards do not even have this).

dennisguse:
Rather implement: https://github.com/OpenTracksApp/OpenTracks/issues/866

Issue #774, opened 2021-05-15
GPX import speed (compatibility)


dennisguse:
FitoTrack and OsmAnd may export speed without a prefix.
OpenTracks will not import these values (requires the actual prefix and it must be `gpxtpx:speed`).

```
      <trkpt lat="0" lon="0">
        <ele>15</ele>
        <extensions>
          <speed>1</speed>
        </extensions>
        <time>XXXX</time>
      </trkpt>
```

PS: OpenTracks' GPX/KML parser are slightly antique...

PR #877, opened 2021-07-07
added "GPX for Strava" export format

gsoros:
This PR adds a GPX track export format which is compatible with Strava, including sensor data.
Not ideal because of the code duplication.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,6 @@ captures/
 Appfile
 releasePlayStore/*
 Gemfile.lock
+
+# VSCode
+*.code-workspace
--- a/src/main/java/de/dennisguse/opentracks/io/file/TrackFileFormat.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/TrackFileFormat.java
@@ -7,6 +7,7 @@ import java.util.Locale;
 import de.dennisguse.opentracks.R;
 import de.dennisguse.opentracks.content.provider.ContentProviderUtils;
 import de.dennisguse.opentracks.io.file.exporter.GPXTrackExporter;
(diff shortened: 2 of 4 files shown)


dennisguse:
**Thanks for your contribution!**

Sadly, I don't see that we will integrate this PR.
It is not our responsibility (Strava should implement the standard), it makes the configuration more complicated, and adds maintenance effort.

Regarding the import, it would be okay from my perspective to be less strict: https://github.com/OpenTracksApp/OpenTracks/issues/774

**TL;DR** 

Strava does not implement the TrackPointExtension v2 standard, which contains speed and which OpenTracks produces.
http://www.garmin.com/xmlschemas/TrackPointExtension/v2
Strava only supports v1, which does not contain speed: https://developers.strava.com/docs/uploads/
And they do no document that they support `<speed>`.

**Recommendation**
I would recommend getting in touch with Strava.
They are a for profit company and should have the resources to implement this.
They should fix it on their side - I don't see a point in creating/maintaining something, because another party is ignoring standards.

**Code review feedback**
- all strings presented in the UI must be localized

PS: Duplicating code is sometimes fine, but it is also a burden too maintain.
And we would need proper test coverage for both classes (which is not yet there).

PS: I don't get why you want to upload data to Strava, but that's a different story.
I also don't want to advertise them in OpenTracks.

gsoros:
Thank you for looking into this.

I completely agree that resolving this issue of interoperability is the responibility of Strava. I have opened a support request with them, not holding my breath for a solution though. 

The subject of why people use Strava would be a long and OT discussion.

dennisguse:
@gsoros Regarding use of Strava: I know.
I have some semi-professional cyclist around and that is a real vendor lock-in.

PS: Really cool that you tried to fix it!

PR #881, opened 2021-07-11
Xioami remove continuous vibration

dennisguse:
TrackRecordingService: notification should not vibrate.


dennisguse:
@StefanKley Can you give this a try?


@rgmf This also removes the notification (after pressing record, the notification was shown as an overlay at the top of the screen). Kind of nice.

StefanKley:
[This](https://github.com/OpenTracksApp/OpenTracks/pull/881/commits/7d15ce63311f1705a5ecd45fb0015d5cb4dbf9ce) solves the problem for me.

dennisguse:
Just tested and according to the documentation priority must be at least LOW (DEFAULT is more important than  low ><).
So, let's fix this problem :)

https://developer.android.com/guide/components/foreground-services

PR #888, opened 2021-07-21
Share picture: if sharing a track, a share picture (PNG) is created in addition

### LINKED ISSUE #363: Sharing tracks as an image (with track statistics)
I would like to share my activities in a photo containing some statistics. Could it be done with a custom dashboard like [OSMDashboard](https://github.com/OpenTracksApp/OSMDashboard)?

**Additional context**
Like this [one](https://support.strava.com/hc/article_attachments/360008726071/socialShare_run.jpg) see [here](https://support.strava.com/hc/en-us/articles/221089587-Sharing-your-Strava-Activities) for more details.

dennisguse:
* Rendering is done via a layout (really easy to change).
* Rendering is done via ShareContentProvider (no tmp file needed)
* Background image is only OpenTracks logo for now

How to use: select a track and select share from the action menu.
Then the GPX/KML and a PNG is created.

For testing, I sent those via e-mail to myself :shrug: 

That's how it looks like:

Fixes #363.


dennisguse:
Something is odd here: it works for some applications (like K9-Mail), but not for others (like Tusky).
It might be that rendering a preview is an issue?

More details here: https://stackoverflow.com/questions/68539083/inflate-layout-from-applicationcontext-i-e-in-a-contentprovider

dennisguse:
Some tiny layer 8 issues; now it is working.
Tested on API30, but may break in the future due to the inflation of layout with a non-visual context.

dennisguse:
@rgmf @pstorch 
One question regarding the backtground image.
I don't have the time to implement an image picker to select a background image from markers the user created.
Do you think it is okay, to "just" ship a predefined image for now (aka all users get the same)?
Otherwise this would take some more weeks...

rgmf:
> @rgmf @pstorch
> One question regarding the backtground image.
> I don't have the time to implement an image picker to select a background image from markers the user created.
> Do you think it is okay, to "just" ship a predefined image for now (aka all users get the same)?
> Otherwise this would take some more weeks...

I think it's enough for now.

rgmf (approved):
👍

pstorch (approved):
I'm fine with it

dennisguse:
Switched to a dark background (as in splashscreen) and use text color white.

PR #890, opened 2021-07-23
Support Android 4.4 / API19.

dennisguse:
We just need to enable multidex and we can support Android 4.X out of the box.
Don't know if this is still out there - but why not :nerd_face: 

NOTE: not actually tested as I don't have such a device :man_shrugging:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,9 +55,9 @@ android {
 
         buildConfigField "String", "VERSION_NAME_FULL", "\"${getVersionName()}\""
 
-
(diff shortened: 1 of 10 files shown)


dennisguse:
What would not work:
* ImportService and ExportService (Export all, post-workout export, import all)
* BluetoothLE scanning does not work.
* Some animations

dennisguse:
Will keep the branch around, but won't invest any time on this endeavor ;)

PR #906, opened 2021-08-06
Fix: README

Chandler-Lu:


dennisguse:
@Chandler-Lu Thanks for fixing it!
That was my fault while working with fastlane to upload PlayStore metadata.

PR #913, opened 2021-08-16
Removed AdaptiveLocationListenerPolicy.

### LINKED ISSUE #886: Removal of AdaptiveLocationListenerPolicy?
OpenTracks has the functionality to reduce the sampling rate of GPS locations if the device is stationary.
This can be activated in the `Settings -> Recording time interval` and provides the two flavors `Adapt for battery live` and `Adapt for accuracy`.
The alternative is to use a fixed time interval (actual default is `0s` aka as fast as Android provides location).

Relevant class: `AdaptiveLocationListenerPolicy`

## Assumption
The adaptive sampling rate is actually not used by our users.
If this is true, we could remove this and avoid some complexity.

:+1: if you use this functionality
:-1: if you **don't** this functionality

dennisguse:
Contains some other tiny refactorings, i.e. cleanup.

Fixes #886.


dennisguse:
This PR contains a functional change as well: we are now registering with (a) desired time interval (as before) and (b) a minimal distance (NEW) for locations from the LocationManager (before it was 0m).
For this, I re-used the minRecordingDistance preference that is used to filter locations before storing them.
My guess is that we can skip this filtering while storing at all.

rgmf (approved):
Less complexity, the better for maintenance. Great for me, I didn't find issues in my tests and nothing strange for me in the code either ;)

dennisguse:
@rgmf thanks :)

Issue #916, opened 2021-08-21
Edit track while recording is not working.


rgmf:
I think this issue was fixed some time ago so we should fix it again and add tests cover.

**To Reproduce**
1. Start a new recording.
2. While recording go to menu (three dots) and select edit.
3. Change name and activity type, for example.
4. Save the changes and back to recording layout.
5. Stop recording.
6. Name and activity is not the edited one.

**Technical information**
 - Device: any
 - OS: any
 - Version: v3.19.0 and v3.19.1

dennisguse:
Interestingly, while recording, the UI is showing the updated info.
And those should be loaded from the db, so we are overwriting it while finishing the recording?

rgmf:
> Interestingly, while recording, the UI is showing the updated info.
> And those should be loaded from the db, so we are overwriting it while finishing the recording?

Yes, it seems like this.

dennisguse:
The bug is in TrackRecordingManager as this one now caches the Track instead of loading it every time from the database.
The track is only loaded when it is resumed or when it is pushed to the UI....
I did this "performance improvement"...

rgmf:
> The bug is in TrackRecordingManager as this one now caches the Track instead of loading it every time from the database.
> The track is only loaded when it is resumed or when it is pushed to the UI....
> I did this "performance improvement"...

I got this for tomorrow but it you can and want go ahead and fix it ;)

PR #917, opened 2021-08-22
TrackRecordingManager only updates track's statistics

rgmf:
**Describe the pull request**
In my opinion TrackRecordingManager should only update track's statistics every time a new track point is received. Only TrackEditActivity can edit a Track data: name, description and activity type.

I've added an Espresso test that checks that track's name is properly edited.

**Link to the the issue**
#916 

**License agreement**

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/EspressoCustomLayoutTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/EspressoCustomLayoutTest.java
@@ -4,8 +4,11 @@ package de.dennisguse.opentracks;
 import androidx.recyclerview.widget.RecyclerView;
 import androidx.test.core.app.ApplicationProvider;
 import androidx.test.espresso.NoMatchingViewException;
+import androidx.test.espresso.UiController;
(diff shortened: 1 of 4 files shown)


dennisguse (approved):
Awesome solution!
My (quick and dirty) approach would have been to just not cache the Track in TrackRecordingManager (aka load it every time before updating it).

PS: The test is somehow flacky :/
Sometimes, it is recording while being on the TrackListActivity and then it does not find the edit as it is not there.
My first guess is that some of the previous UI tests is not cleaning up correctly.

PPS: Seems to be customLayoutTest is not stopping the recording and then the edit test as trouble.

rgmf:
> Awesome solution!

Thank you ;)

> PS: The test is somehow flacky :/
> Sometimes, it is recording while being on the TrackListActivity and then it does not find the edit as it is not there.
> My first guess is that some of the previous UI tests is not cleaning up correctly.
> 
> PPS: Seems to be customLayoutTest is not stopping the recording and then the edit test as trouble.

In the last push it's fixed. The problem was in CustomLayoutTest like you pointed out. It wasn't stopping it.

Issue #862, opened 2021-07-01
Clink announcement 🔉 for phones without TTS


vsevjednom-cz:
**Is your feature request related to a problem? Please describe.**
Phones without TTS can't use voice announcement. #763

**Describe the solution you'd like**
Some sound like sms notification every kilometer for example?

**Additional context**
Something like this: https://www.youtube.com/watch?v=TCMg0J8mfuE

vsevjednom-cz:
I tested it while ride my mountain bike today at `Android 9.1.0`.

Voice announcement plays every 60 seconds when I set 60 minutes.

And what type of sound I must to amplify to hear voice announcement?

dennisguse:
@vsevjednom-cz by time works for me as expected. Tested with 1 and 2min. Regarding channel: the beep is played as MEDIA. Don't know if this is the best choice.

rgmf:
@vsevjednom-cz @dennisguse I think time interval settings are in seconds not in minutes, altough you can see minutes like a unit:

![Screenshot_20210905-181815_OpenTracks_(Debug)](https://user-images.githubusercontent.com/4819612/132134043-4f440cb2-0009-4512-8190-22d0c5a4440a.png)

dennisguse:
That's my fault...

rgmf:
Fixed and push into main. You can see the commit here:

https://github.com/OpenTracksApp/OpenTracks/commit/40aab9444a9410860e3ef647b013bf8b5cf375d4

vsevjednom-cz:
> Regarding channel: the beep is played as MEDIA.

It's too quiet. I have set the maximum volume. I've got my mobile on my back when I ride my bike. Is possible to set more louder announcement? I use it as a check of turned on application.

And another idea: is possible to set announcement of autopause starts and autopause ends? I used this few years ago in Sports Tracker app.

dennisguse:
@vsevjednom-cz No clue. I guess, you need to experiment here yourself. It was literally the first time I used Android's media subsystem. Also the audio file might not be loud enough itself or it might be device dependent...

vsevjednom-cz:
I also think that the low volume of sound is because the media file is not loud. How can I select another sound file?

dennisguse:
> I also think that the low volume of sound is because the media file is not loud. How can I select another media file?

It is the only mp3 that is in OpenTracks.

vsevjednom-cz:
[...]

PR #922, opened 2021-08-24
Cleanup: For voice announcements use a handler

dennisguse:
Our time-based voice announcements use a Timer until now, but everywhere else we are using Handlers.
We now use a Handler for this and this simplifies the code a lot.

PS: Not yet tested on a real device and there is no test coverage for the code :/


dennisguse:
@vsevjednom-cz Github provides these commands for you. Jus use the arrow on the right side of the "Rebase and merge" button.
Just skip the `git merge main`.

image

vsevjednom-cz:
Yesss. I tried `1 min` of `menu_voice_frequency` (I'm in home). It works 🙃. My idea #862 is done 🥳.

PR #932, opened 2021-08-30
Using Zulu builds of OpenJDK in GH actions

carldea:
The AdoptOpenJDK has been discontinued since July 2021. When using Zulu OpenJDK builds gets you all the latest and greatest (TCK Tested) builds for all versions of OpenJDK.

https://adoptopenjdk.net 

"24th July 2021: AdoptOpenJDK is moving to the Eclipse Foundation and rebranding.
Our July 2021 and future releases will come from Adoptium.net"

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,7 +13,7 @@ jobs:
 
       - uses: actions/setup-java@v2
         with:
-          distribution: 'adopt'
+          distribution: 'zulu'
           java-version: '11'
 
       - run: ./gradlew lint
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
 
       - uses: actions/setup-java@v2
         with:
-          distribution: 'adopt'
+          distribution: 'zulu'
           java-version: '11'
 
       - name: run tests


dennisguse:
@carldea can you squash the two commits?

dennisguse:
@carldea And why not use Temurin?
https://github.com/TeamNewPipe/NewPipe/pull/6995

carldea:
Squash completed. 

> @carldea can you squash the two commits?

I completed the squash.

carldea:
> @carldea And why not use Temurin?
> [TeamNewPipe/NewPipe#6995](https://github.com/TeamNewPipe/NewPipe/pull/6995)

No reason. Except Temurin doesn't support every version JDK if you ever decide to have your app run on different versions of Java such as the upcoming LTS version of JDK 17. 

Some customers in production environments will have a better peace of mind when their app is able to run on many (updated) versions of Java. I guess it's really about comfort level, or level of risk if you plan on deploying your app in production or built with a particular distro. At the end of the day, both distros are TCK tested and built with the same OpenJDK source. 

I chose Zulu builds because they are very mature (excellent track record) with security updates and TCK tested since GitHub actions setup-java@v1. 

Zulu builds offer every version of OpenJDK from 6 to 18-ea. For example 8, 11, 17-ea (LTS versions) are all supported (and versions in between) with the latest security updates. 

It's great to have choice. You could alway try it out, and if you're not happy just change it. Or I could add a matrix (array) to build on both distros. Since runners run in parallel anyways. 
:-)

PR #933, opened 2021-08-31
CI: build nightly on push.

dennisguse:
This PR adds a Github Action task that builds a signed APK and attaches it as an artifact to the Github Action result.

### DIFF
--- a/.github/workflows/build_nightly.yml
+++ b/.github/workflows/build_nightly.yml
@@ -1,0 +1,34 @@
+#https://github.com/marketplace/actions/android-emulator-runner
+name: Build nightly
+on: [ push ]
+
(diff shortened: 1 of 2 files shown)


dennisguse:
@pstorch What do you think?
Does this count as "nightly" build?

pstorch:
@dennisguse I think so. It's a usable first step. 
Next step would be to wrap an F-Droid repo around it, so interested users get updates automatically.

dennisguse:
Fdroid builds would be awesome for the main branch!
We only need to update the versionCode using the commit count.

Only trouble for me at the moment: I have no webspace to host the Fdroid files.
On Github Action itself this is not possible.

PR #939, opened 2021-09-07
[PROPOSAL] Gain and loss are integers

rgmf:
**Describe the pull request**
Altitude, gain and loss are showed like floats with 1 decimal.

I think, gain and loss should be integers because they always are X.0

In this PR are the changes for this.

**License agreement**


dennisguse:
Why not create a separate function in StringUtils to handle gain/loss? Technically, they are altitude but also different :)

PS: For rendering gain integer is sufficient; while internally we might encounter floats as we could implement an adjustable gain/loss difference that is different than the current 3m.
And then we could have something like 1.2m gain/loss increases.

PS: Good idea :D

rgmf:
> Why not create a separate function in StringUtils to handle gain/loss? Technically, they are altitude but also different :)

Done. I've created a new function in StringUtils but I'm not sure if the name is so good. If you think in a better name let me know :D

dennisguse:
What about formatAltitudeChange?￼￼
I wanted to introduce a value object for this anyhow 😁

Issue #931, opened 2021-08-30
Cadence sensor: 0rpm instead of None


rgmf:
**Is your feature request related to a problem? Please describe.**
I'd prefer to see 0rpm instead of None when I stop pedaling my bike.

**Describe the solution you'd like**
When cadence sensor is connected never shows "None" like a value but "0 rpm".

dennisguse:
Should it be value_unknown?

rgmf:
> Should it be value_unknown?

IMHO I should be value_unknown when cadence sensor is not connected and 0rpm if it's connected.

But I don't know very well how sensors work. Does cadence sensor send 0 when cyclist is not pedaling? If so, I don't understand where None value comes from.

dennisguse:
You are right that are two cases.

Regarding the data: those sensors only send data when they measured something (like the wheel rotated).
If nothing is happening, then they don't send data.
So, the timeout occurs OpenTrack invalidates the data (5s timeout) and then the data is null resulting and `StringUtils.formatCadence()` returns None.

So, the invalidation should actually set them 0 and on disconnect to null.

Issue #930, opened 2021-08-30
Speed sensor: speed is never 0km/h


rgmf:
**Describe the bug**
When you are recording a track with a cycling sensor speed and you stop at a semaphore, for example, the speed data never is 0km/h so moving time go on.

I guess the speed I see in this field is the last the sensor send to OT.

**To Reproduce**
1. Start a new recording with a speed sensor.
2. Go ahead and stop.
3. You'll see the speed going down until some speed like 4km/h, for instance.
4. When you are completely stopped the speed is not 0km/h but the last sensor sended to OT.

**Technical information**
 - Device: Pixel 2 XL
 - OS: LineageOS 17.1
 - Version v3.19.x and v3.20.0 (I couldn't test it in older versions)

dennisguse:
Nice finding - I never noticed that.
I thought we use SensorData.isRecent() to determine if we got "fresh" data (Default timeout is 5s; don't know why I came up with 5s. It was for HR sensors).
Afterwards we do some computation in TrackStatisticsUpdater to smoothen the speed (only done when new data arrives).
So, this might be the reason (best guess, did not check).

rgmf:
> Nice finding - I never noticed that.
> I thought we use SensorData.isRecent() to determine if we got "fresh" data (Default timeout is 5s; don't know why I came up with 5s. It was for HR sensors).
> Afterwards we do some computation in TrackStatisticsUpdater to smoothen the speed (only done when new data arrives).
> So, this might be the reason (best guess, did not check).

I see.

This will be a great issue for me so I can look into this part of OT that is unknown for me 😄

dennisguse:
It is not the TrackStatisticsUpdater, but I guess StatisticDataBuilder.
It checks the SensorData.isRecent() and if not uses the GPS value (and that is already old likely containing movement).

I guess, SensorData.isRecent() should be removed and should be handled internally by SensorData.getValue() aka return default value if the data is too old.

rgmf:
2021-09-18 Update:

Same issue with Speed GPS... When you are using GPS for speed and distance (not sensors connected) and you stop pedaling GPS speed is the last GPS valid speed but it should be 0.

Issue #833, opened 2021-06-21
Cycling speed and distance sensor: returned weird value (way too large)


dennisguse:
Looks like an overflow issue.
Problematic data: 
```
<trkpt lat="XX" lon="XX">
<ele>72</ele>
<time>2021-06-21T17:21:23.966Z</time>
<extensions><gpxtpx:TrackPointExtension>
<gpxtpx:speed>8.3</gpxtpx:speed>
<opentracks:gain>0</opentracks:gain>
<opentracks:loss>0</opentracks:loss>
<opentracks:distance>0</opentracks:distance>
</gpxtpx:TrackPointExtension></extensions>
</trkpt>
<trkpt lat="XX" lon="XX">
<ele>67</ele>
<time>2021-06-21T17:21:33.954Z</time>
<extensions><gpxtpx:TrackPointExtension>
<gpxtpx:speed>577947520</gpxtpx:speed>
<opentracks:gain>0</opentracks:gain>
<opentracks:loss>0</opentracks:loss>
<opentracks:distance>9,169,715,200</opentracks:distance>
</gpxtpx:TrackPointExtension></extensions>
</trkpt>
```

Device: Garmin Sensor Speed 2 (software revision unknown).
Wheel circumference: 2135mm

dennisguse:
Hypotheses:
* (A) Overflow issue (distance from sensor).
* (B) Sensor provided broken data (distance from sensor)
* (C) Distance was computed by GPS
  => previous TrackPoints were recorded every 1s; but this one was took 10s

For (A)
```
                long wheelDiff = UintUtils.diff(wheelRevolutionsCount, previous.wheelRevolutionsCount, UintUtils.UINT32_MAX);
```

Just checked the sensor data (raw rotation count):
2021-06-13: 381,616
2021-06-21: 457,371

For the time interval the following would be expected: 10s at ca. 30km/h => ca. 8m/s => 80m-100m => ca. 50 wheel rotations

dennisguse:
Did not happen again.
Closing for now.

rgmf:
I get 4.5374991E9 value as a cadence while working on #931.

I can reproduce it and I did it with main branch.

To reproduce (I did it several times):
- Start a recording with a cadence sensor.
- Stop pedaling for more than 20 minutes (I don't know where threshold is).
- Restart pedaling and you'll see a big value for cadence.

I can see the following values for `crankRevolutionsCount` from the beginning of the recording:
```
[...]

PR #953, opened 2021-09-18
Sensors set to 0 instead of None and bugfixes

rgmf:
**Describe the pull request**

- Cadence set to 0 rpm when connected but user is not pedaling.
- Speed set to 0 km/h when connected but last value is outdated.
- HR sensor set to 0 bpm when connected but last value is outdated.
- The same thing should be happen to Power sensor (but it does not tested).
- Bug: overflow when reset sensors (cadence and speed) are fixed (see #833).
- Moving time stops when is not moving (the last track point is not moving) and restarts when last track point is moving. So far now, moving time always updated its value.

**Test it**
So for now, I tested it at home but not in real routes. I pushed it for discussion but I'll test it in real routes.

**Link to the the issue**
#931
#930
#833

**License agreement**


src/main/java/de/dennisguse/opentracks/content/sensor/SensorDataCycling.java
```diff
@@ -59,17 +59,29 @@ public int getCrankRevolutionsTime() {
...
+            return 0f;
+        }
+
         public void compute(Cadence previous) {
             if (hasData() && previous != null && previous.hasData()) {
                 float timeDiff_ms = UintUtils.diff(crankRevolutionsTime, previous.crankRevolutionsTime, UintUtils.UINT16_MAX) / 1024f * UnitConversions.S_TO_MS;
                 if (timeDiff_ms <= 0) {
                     Log.e(TAG, "Timestamps difference is invalid: cannot compute cadence.");
                     value = null;
-                } else {
```
dennisguse, line 68: I am not sure about this.
On the one hand, the documentation says that the value can overflow and should be treated like this.
On the other hand, we would only miss one rotation if this happens - so the impact is not that big.

The question in my head actually is more: why does the sensor behaves like it does?
Could it be that the sensor turns off (e.g., power saving) and then resets the counter?
And could we detect this in the BluetoothManager (e.g., does a disconnect happen)?

dennisguse (commented):
Cool!

src/main/java/de/dennisguse/opentracks/content/sensor/SensorDataCycling.java
rgmf: Really good questions.

>  Could it be that the sensor turns off (e.g., power saving) and then resets the counter?

I'm not sure but according the tests it looks that sensor turns off and it restarts counter after a long period without activity. But we should look into this.

>  And could we detect this in the BluetoothManager (e.g., does a disconnect happen)?

That will be awesome and we can try to accomplish it.

For now, I think is better missing a rotation that saving a huge value for cadence or speed that would throws off statistics. What do you think? Should we accept this workaround for now?

rgmf:
@dennisguse It'd worked like a charm in today's route with several stops at traffic light and a long stop.

So, for me the only thing to do is decide what to do with the overflow.

dennisguse:
Sounds good!

Only suggestion: put a big todo in the code and let's add a test for this specific case (and deprecate it, so the IDE will provide warnings ;) ).

rgmf:
> Sounds good!
> 
> Only suggestion: put a big todo in the code and let's add a test for this specific case (and deprecate it, so the IDE will provide warnings ;) ).

Sorry @dennisguse, I don't know if I understood what you pointed out to me. Is my last commit what you expected?

dennisguse:
Exactly. Only thing: we can keep @Test and add @Ignore for those tests that should be ignored. Then those are highlighted when executing ;)

PR #964, opened 2021-09-28
Cleanup: SharedPreference as static variable.

### LINKED ISSUE #961: Refactor use of SharedPreferences
SharedPreferences can be referenced as a static field.
This would simplify the handling them, as we don't need to pass the sharedpreferences that we got from each individual context.

The solution is to just keep the sharedpreferences as a static field in PreferencesUtils and init it if it is null.
Sharedpreferences can then be removed everywhere.

For OSMDashboard it was done a while ago: https://github.com/OpenTracksApp/OSMDashboard/pull/102

dennisguse:
Fixes #961


dennisguse:
@pstorch @TacoTheDank does this make sense?

TacoTheDank:
@dennisguse Brings a tear to my eye...

beautiful-amazing

dennisguse:
@TacoTheDank That meme made my day!!! :sunglasses:

rgmf:
@dennisguse is it ready for testing? Can I install it for testing?

PR #971, opened 2021-10-07
Bugfix: reset on SensorDataSet should only take place if the TrackPoint was stored

### LINKED ISSUE #969: Sensor data not sometimes ignored (when using GPS in addition)
Yesterday, I participated in a cycling event with known elevation gain and distance.
As I had the impression that the sensor data handling is dropping data (initial assumption distance only), I started OpenTracks on the same smartphone twice (Fdroid version + most recent build from main; and Komoot for verification).

Used sensor: Garmin Bike Sensor 2

Observations:
* (only GPS)  altitude gain/loss only 80% of the one reported by Komoot; distance identical
* (cycling sensor + GPS) distance only 60%, but GPS track looks okay
* (cycling sensor + GPS)  gain only 20% while loss 75%

Best guess: the problem is that we reset the sensor data every time a new GPS location is provided, but we might not store it as the distance to the previous location is not large enough leading to dropped data. Reset should only take  place if we actually stored the data.

Code: 
```
    public void onNewTrackPoint(TrackPoint trackPoint, Distance thresholdHorizontalAccuracy) {
        fillAndReset(trackPoint);

        service.newTrackPoint(trackPoint, thresholdHorizontalAccuracy);
    }
```

I need to create a test case for this.
This issue is presented since we introduced altitude gain/loss and distance sensors.

dennisguse:
This PR changes the behavior that TrackRecordingManager reports if the current TrackPoint was stored and only this happened, TrackPointCreator resets the sensor data.

This solves the problem that the drop sensor data.
BUT we still may DUPLICATE some sensor data as in some situations, we store a TrackPoint LATER (i.e., the SensorData was not reset) and then the data from the first TrackPoint is also included into the second one.
Only a problem for cumulative sensor data aka distance.

Includes test showing that #969 is fixed, but DUPLICATE is an issue.
The sensor data (distance) is stored twice in SEGMENT_END_MANUAL.
It is currently 4m, but should be 2m (as 2m are already included in the previous TrackPoint that is also stored as it is referenced via lastTrackPoint).

Fixes #969.

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTestLocation.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTestLocation.java
@@ -39,6 +39,7 @@ import de.dennisguse.opentracks.io.file.importer.TrackPointAssert;
 import de.dennisguse.opentracks.services.sensors.AltitudeSumManager;
 import de.dennisguse.opentracks.services.sensors.BluetoothRemoteSensorManager;
 import de.dennisguse.opentracks.settings.PreferencesUtils;
+import de.dennisguse.opentracks.util.PressureSensorUtils;
 
 /**
  * Tests insert location.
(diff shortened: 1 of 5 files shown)


dennisguse:
@rgmf Thanks for testing!
I agree - somehow this recording code is still mindblowing (depending on the time of the day / the problem).

Issue #500, opened 2020-11-07
Indoor training


ChristophWurst:
**Is your feature request related to a problem? Please describe.**

As far as I can tell the app is currently focused on *outdoor training*. That is, GPS signal is assumed to be available and the app most likely also uses it to determine activity and pause sections of a recording.

**Describe the solution you'd like**

For cycling, running and possibly other sports it would be neat if the app could also support training indoors. This means there should be an option to disable GPS. The active/pause is then either fully manual or based on sensors (e.g. steps for running, speed for cycling).

**Describe alternatives you've considered**

n/a

**Additional context**

Cycling power meter sensor PR: #444
My feature request for power meter readings of smart trainers: #498, specifically

> while the sensor data will be shown, we only record (store it in the database) when a GPS coordinate is stored.
So, no GPS signal, no diagrams, statistics, export etc.

dennisguse:
That's a kind of complicated as we only store data based upon GPS and indoors there will be no GPS data.
So, we need to store the data using some timer?

And are not able to provide a distance as we use the GPS for this.
One option would be to use a wheel rotation sensor (BLE: Cycling Speed) to estimate the traveled distance.

ChristophWurst:
> So, we need to store the data using some timer?

I recently inspected some .fit files from wearables and they use timestamps for their data points. So each timestamp has a value for lat, long, speed and so on.

> One option would be to use a wheel rotation sensor (BLE: Cycling Speed) to estimate the traveled distance.

Indeed, that is how other trackers do it as well. Either you have GPS or a speed sensor, otherwise distance can't be determined.

pstorch:
Do we need to revise our backup (import/export) strategy? As KML and GPX are for Tracks, don't know if they are suitable for indoor activities.

dennisguse:
Following the implementation of #316, we could implement this now.
Assumption: there is no GPS signal
For pure indoor training, we would then need to store the generated data (in the STOP trackpoint).

Following tasks:
* [ ] get distance from a sensor (like Cycling Cadence And Speed Sensor)
* [ ] store this distance in the trackpoint
* [ ] use the sensor derived distance (if present) to compute the TrackStatistics 
* [ ] import/export distance via GPX/KML
* [ ] (nice to have) present data source in the UI (also do this for speed)
* [ ] (nice to have) implement distance and speed parsing for wheel-based powermeters like Wahoo Kickr


[...]

PR #984, opened 2021-10-14
Cadence from running sensor not shown while recording

dennisguse:
* [x] show cadence from running sensor
* [x] cadence: show sensor name from running sensor
* [x] speed: show sensor name from running senor


dennisguse:
This is ready for review - although I am not yet that happy with the code.
There is some refactoring required, but it works for now ;)

rgmf:
> Regarding the footpod sensors: they are quite precise (at least the very expansive Stryd). Did not yet test in detail with my Polar Stride. In any case, I want to go running without GPS by only using the sensor. I am usually running the same route and I don't need the coordinates (i.e., special case of #500). Saves also some energy in the phone :D

For indoor it's great, but I didn't realize the possibility of save battery turning GPS off.

PR #1008, opened 2021-11-02
README: add link to Amazon App Store

dennisguse:
I had some time and published OpenTracks to the Amazon App Store as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,15 @@ _OpenTracks_ is a sport tracking application that completely respects your priva
         </td>
         <td></td>
     </tr>
+    <tr>
+        <td></td>
+        <td align="center">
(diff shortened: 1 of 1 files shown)


pstorch:
Cool that it is possible. But is this Store used by usable devices? I only associate Amazon Store with Fire TV and Kindle. Are there devices which can be used for workout?
Anyway, give it a go. If it's useful for someone, why not?

dennisguse:
@pstorch Windows 11 devices - :sunglasses: 
Otherwise, no clue - some years ago I used the Amazon Store every now and then, but yeah.

rgmf (approved):

rgmf:
I you all think this can be useful then welcome ☺️

Issue #1007, opened 2021-11-02
Recording of points without meeting the conditions of the distance and speed filters


vlmendz:
**Describe the bug**
These unnecessary points are recorded following a pattern:

Before each point that meets the conditions of the filters, the previous point received is recorded (see images):

![Captura_cs_ot_5m_0s](https://user-images.githubusercontent.com/91810638/139786965-5d261485-57cd-479f-974f-879bdd20d761.PNG)
![Captura_qgis_ot_5m_0s](https://user-images.githubusercontent.com/91810638/139786985-b865db5a-c80c-4145-bb77-bab8e39484e2.PNG)
☝️Points of a fragment of a recorded track (distance interval=5m)

![Captura_cs_ot_5m_0s_dist](https://user-images.githubusercontent.com/91810638/139787225-72ecd191-d3ad-44b7-87bb-09870c09d01c.PNG)
☝️Distances between the points of a fragment of a recorded track (distance interval=5m)

Examining the TrackRecordingManager.onNewTrackPoint() function, I see a possible cause of this behavior:

	void onNewTrackPoint(TrackPoint trackPoint, Distance thresholdHorizontalAccuracy) {

        //TODO Figure out how to avoid loading the lastValidTrackPoint from the database

        TrackPoint lastValidTrackPoint = getLastValidTrackPointInCurrentSegment(trackId);

        //Storing trackPoint

        // Always insert the first segment location
        if (!currentSegmentHasTrackPoint()) {
            insertTrackPoint(trackId, trackPoint);
            lastTrackPoint = trackPoint;
            
            // Posible 2nd modification:   ******************************
            // Posible solution for TODO above?
            lastValidTrackPoint = trackPoint; 
            
            return;
        }

        Distance distanceToLastTrackLocation = trackPoint.distanceToPrevious(lastValidTrackPoint);
        if (distanceToLastTrackLocation != null) {
            if (distanceToLastTrackLocation.greaterThan(maxRecordingDistance)) {
                
                // Remove also this?  ******************************              
                insertTrackPointIfNewer(trackId, lastTrackPoint);

                trackPoint.setType(TrackPoint.Type.SEGMENT_START_AUTOMATIC);
                insertTrackPoint(trackId, trackPoint);

                lastTrackPoint = trackPoint;
                
                // Posible 2nd modification:   ******************************
                // Posible solution for TODO above?
                lastValidTrackPoint = trackPoint; 
                
                return;
            }
            
            
[...]

PR #1010, opened 2021-11-03
Proposed solution for recording points that do not meet filter conditions

vlmendz:

**Describe the pull request**
Proposed solution to #1007 

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/1007#issue-1041879140

**License agreement**

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTestLocation.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTestLocation.java
@@ -20,11 +20,14 @@ import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.internal.debugging.LocationImpl;
 
+import java.time.Instant;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import de.dennisguse.opentracks.content.data.Altitude;
 import de.dennisguse.opentracks.content.data.Distance;
 import de.dennisguse.opentracks.content.data.Speed;
 import de.dennisguse.opentracks.content.data.TestDataUtil;
@@ -187,7 +190,7 @@ public class TrackRecordingServiceTestLocation {
         TrackRecordingServiceTest.newTrackPoint(service, 45.000002, 35.0, 3, 15);
         TrackRecordingServiceTest.newTrackPoint(service, 45.000003, 35.0, 4, 15);
         TrackRecordingServiceTest.newTrackPoint(service, 45.000004, 35.0, 5, 15);
-        TrackRecordingServiceTest.newTrackPoint(service, 45.000005, 35.0, 6, 15);
+        TrackRecordingServiceTest.newTrackPoint(service, 45.0002, 35.0, 6, 15);
 
         service.endCurrentTrack();
 
@@ -207,7 +210,7 @@ public class TrackRecordingServiceTestLocation {
                         .setAltitudeGain(0f)
                         .setAltitudeLoss(0f),
                 new TrackPoint(TrackPoint.Type.TRACKPOINT, null)
-                        .setLatitude(45.0005)
+                        .setLatitude(45.0002)
                         .setLongitude(35)
                         .setHorizontalAccuracy(Distance.of(6))
                         .setSpeed(Speed.of(15))
@@ -217,6 +220,9 @@ public class TrackRecordingServiceTestLocation {
                         .setAltitudeGain(0f)
                         .setAltitudeLoss(0f)
         ), trackPoints);
+
+
+
     }
 
     @MediumTest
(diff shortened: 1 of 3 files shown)


dennisguse:
Yes, you are right that is the code that creates these additional TrackPoints.
Storing these additional TrackPoints is not required for the sensor data.

PS: If you now setup AndroidStudio, you can execute the tests (using the emulator) and see the difference (some tests are failing with those changes). Failing tests are in TrackRecordingServiceTestLocation.
Because now we need to start discussing why/ if we need to store these TrackPoints or not or just in some cases.

One test case result: 
```
java.lang.AssertionError: Expected: [time=null (type=SEGMENT_START_MANUAL(-2)), time=null (type=TRACKPOINT(0)): lat=45.0 lng=35.0 acc=Distance{distance_m=1.0}, time=null (type=TRACKPOINT(0)): lat=45.0 lng=35.0 acc=Distance{distance_m=6.0}, time=null (type=SEGMENT_END_MANUAL(1))]
actual: [time=2021-11-03T21:44:16.716Z (type=SEGMENT_START_MANUAL(-2)), time=2021-11-03T21:44:16.766Z (type=TRACKPOINT(0)): lat=45.0 lng=35.0 acc=Distance{distance_m=1.0}, time=2021-11-03T21:44:16.806Z (type=SEGMENT_END_MANUAL(1))]
at de.dennisguse.opentracks.io.file.importer.TrackPointAssert.assertEquals(TrackPointAssert.java:86)
at de.dennisguse.opentracks.services.TrackRecordingServiceTestLocation.testOnLocationChangedAsync_idle_withMovement(TrackRecordingServiceTestLocation.java:292)
```

vlmendz:
As I see it, storing these GPS TrackPoints breaks the operation of the filters, and as a consequence produces unwanted side effects in track recording.

@dennisguse for now I can't install Android Studio. Wouldn't it be better to assign this PR to someone with conditions to perform these tests?

dennisguse:
Failing tests are expected (actually a good thing as we are testing this functionality). One approach would be to check the test (especially the one that fail) and check if the difference makes sense (i.e., is desired).

For me, this has a low priority.
So, it might take until next year.
Other (also non-OpenTracks) things are at the moment with a higher priority :man_shrugging: 

PS/ does not seem to be urgent as we have this functionality in the code since years.

PS/ actually Github should execute the tests, but it isn't starting them (and I cannot trigger it manually) for Pull Requests from forked repositories.
I will look into this.

vlmendz:
@dennisguse Good day, today I reread the code, taking into account what you wrote above ("Storing these additional TrackPoints is not required for the sensor data."). It seems to me that this related lastTrackPoint code is basically reminiscent of the old idle/non-idle code.

Assuming this, I modified the code again, removing everything related to the storage of lastTrackPoint. Now the code is somewhat simpler, and some energy could be saved (at least theoretically).

P.s. Possibly now only one of the functions (insertTrackPoint or insertTrackPointHelper) would be necessary. I kept both because I don't know if they're used anywhere else.

dennisguse:
After some heavy thinking: I agree that code is doing something but is not really useful (anymore).
Removing this, simplifies the recording code a lot (even getting rid of the trouble with cumulative sensor data that was fixed with v3.21.0).

Their is one edge case though: if a user pauses/stops a track it would be slightly more correct to store lastTrackPoint (if it wasn't stored so far; and ignore if the recording distance intervals is larger than the threshold).
In this case, we have received a location, but ignored it (so far) and might store as it is slightly closer to the user's final position than the one we stored beforehand.
With relatively low recording distance interval this is not really an issue: 10m (100m) distance threshold means, we would lose max 9.9m (99.9m).
IMHO: this is not worth the effort to keep this complexity around.

@pstorch @rgmf Usual question: did I miss something?
Could this code be useful for anything else?
PS/ test cases were initially created using the current implementation - without understanding it by me some time ago.
Just to prevent regressions due to refactoring.

@vlmendz I like your style on how you dig in the code base - it really triggers some thoughts :sunglasses: 
I adjusted the tests, removed other code that is then not used anymore.
As you still cannot compile yourself(?) and the nightly build is not triggered for you PRs (don't know why): APK is attached; signed with the PlayStore key.
[OpenTracksUpstream-releasePlayStore.zip](https://github.com/OpenTracksApp/OpenTracks/files/7492471/OpenTracksUpstream-releasePlayStore.zip)

rgmf:
> With relatively low recording distance interval this is not really an issue: 10m (100m) distance threshold means, we would lose max 9.9m (99.9m). IMHO: this is not worth the effort to keep this complexity around.


I agree with yout. If user select a distance interval of 100m then user doesn't care about the precision, so 99.9m is not too much.

The code now seems simpler.

Great job @vlmendz

vlmendz:
Their is one edge case though: if a user pauses/stops a track it would 
    be slightly more correct to store lastTrackPoint ...

As I see it, the key aspect here is to use appropriate values for filters. For example, with the current recording logic I can use a distance interval of 5m and a minimum speed of 2 km/h. These values allow to record a track in great detail and at the same time with practically no unnecessary locations/noise (even making multiple and/or long stops).

Therefore, the pause would only be necessary to continue a track in a different place from where it stopped (I like this feature of OpenTracks that creates a new segment when using the pause and when moving without a GPS fixation for a distance > = than a customizable maximum distance). 

    As you still cannot compile yourself(?) 

I can't compile yet 😟

    and the nightly build is not triggered for you PRs (don't know why)

This is the first PR I have created, so if you don't know why, what could I know 🤷🏻😀

    APK is attached; signed with the PlayStore key.

Downloaded and installed 👍

    I agree with yout. If user select a distance interval of 100m then user 
    doesn't care about the precision, so 99.9m is not too much.

Also agree 👌

P.s. Thanks :)

dennisguse:
@vlmendz FYI: I will squash the commits (so only one commit will be merged.

And what can we do to enable you to compile yourself?
It is basically installing AndroidStudio and cloning the code :)

dennisguse:
Damn: forgot to update the message.
Will create another commit locally.

dennisguse:
Actual commit on main: https://github.com/OpenTracksApp/OpenTracks/commit/08ba7d50c73197a1230eca8cc72dc1d501d7a6af

PR #1011, opened 2021-11-07
UI: Remove "Sensor:" prefix from speed description.

dennisguse:
While working on the sensor stuff, I noticed that for speed, we put "Sensor: " in front of the name while for the other sensors, we don't do this.


rgmf (approved):
I think it's clear it's a sensor so "sensor" prefix is not needed (the cleaner UI, the better).

PR #1012, opened 2021-11-07
Added profiles for custom layouts

### LINKED ISSUE #938: Include option for changing workout that is being recorded.

I like the new customisable layout when recording tracks, however I could not see the ability to add a button/cell that allows you to choose what activity you are recording. That is now only accessible via the menus.

Restore the button that allows the user to choose what activity they are recording, but as a configurable option in-keeping with the new layout.

rgmf:
Users can add severals custom layouts and select one of them in recording fragment. The last selected layout is the one that will be used in the next recording.
Fixes #938.

~~Some videos to see quickly how it works:~~

PD/ Only tested on AVD. I have to test it in real situations.

PD2/ I can't upload videos here to show the new feature. I've uploaded them to mastodon: https://mastodon.online/web/@rgmf/107236937885092968


rgmf:
@slackline FYI, as you opened #938, here you have a PR that you can test. Otherwise, I hope there will be a future version with this PR.

src/androidTest/java/de/dennisguse/opentracks/settings/PreferencesOpenHelperTest.java
```diff
@@ -0,0 +1,54 @@
...
+    private final Context context = ApplicationProvider.getApplicationContext();
+
+    @Test
+    public void test_upgradeFrom1To2() {
+        // given the version 1
+        SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+        SharedPreferences.Editor editor = sharedPreferences.edit();
+
+        String oldCustomLayoutCsv = context.getString(R.string.activity_type_unknown) + CsvConstants.ITEM_SEPARATOR
+                + "distance,1,1,0" + CsvConstants.ITEM_SEPARATOR + "speed,1,1,0" + CsvConstants.ITEM_SEPARATOR;
```
dennisguse, line 35: In the tests, ITEM_SEPARATOR must not be referenced.
It is easier to read if not :)

src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
```diff
@@ -256,11 +261,11 @@ public boolean onPrepareOptionsMenu(Menu menu) {
...
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
-        if (item.getItemId() == R.id.track_detail_insert_marker) {
-            Intent intent = IntentUtils
-                    .newIntent(this, MarkerEditActivity.class)
-                    .putExtra(MarkerEditActivity.EXTRA_TRACK_ID, trackId);
-            startActivity(intent);
+        if (item.getItemId() == R.id.track_detail_menu_select_layout) {
+            AlertDialog.Builder builder = new AlertDialog.Builder(this);
+            List<String> profiles = PreferencesUtils.getAllCustomLayouts().stream().map(Layout::getProfile).collect(Collectors.toList());
```
dennisguse, line 266: Could this be moved to PreferenceUtils? (i.e., create the list there?)

src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
```diff
@@ -269,6 +274,14 @@ public boolean onOptionsItemSelected(MenuItem item) {
             return true;
         }
 
+        if (item.getItemId() == R.id.track_detail_insert_marker) {
```
dennisguse, line 277: should still be the first entry in the menu

src/main/java/de/dennisguse/opentracks/services/handlers/LocationHandler.java
```diff
@@ -106,6 +109,14 @@ public void onLocationChanged(@NonNull Location location) {
         boolean isAccurate = trackPoint.fulfillsAccuracy(thresholdHorizontalAccuracy);
         boolean isValid = LocationUtils.isValidLocation(location);
 
+        if (lastTrackPoint != null) {
+            double elapsedTime = trackPoint.getTime().minus(lastTrackPoint.getTime().toEpochMilli(), ChronoUnit.MILLIS).toEpochMilli() / 1000d;
+            double elapsedDistance = trackPoint.distanceToPrevious(lastTrackPoint).toM();
+            //Log.e("probando", "Speed (km/h): " + ((elapsedDistance / elapsedTime) * UnitConversions.MPS_TO_KMH));
```
dennisguse, line 115: :)

src/main/java/de/dennisguse/opentracks/settings/PreferencesOpenHelper.java
```diff
@@ -1,34 +1,58 @@
...
+        if (version > lastVersion) {
             onUpgrade();
         }
     }
 
     private void onUpgrade() {
-        PreferencesUtils.setInt(R.string.prefs_last_version_key, PREFERENCES_VERSION);
-        switch (PREFERENCES_VERSION) {
+        PreferencesUtils.setInt(R.string.prefs_last_version_key, version);
+        switch (version) {
```
dennisguse, line 31: should be a loop, so 0->1 and 1->2 are executed.

src/main/java/de/dennisguse/opentracks/util/RecyclerViewSwipeDeleteCallback.java
```diff
@@ -0,0 +1,91 @@
...
+import android.view.View;
+
+import androidx.annotation.NonNull;
+import androidx.core.content.ContextCompat;
+import androidx.recyclerview.widget.ItemTouchHelper;
+import androidx.recyclerview.widget.RecyclerView;
+
+import de.dennisguse.opentracks.R;
+
+abstract public class RecyclerViewSwipeDeleteCallback extends ItemTouchHelper.Callback {
```
dennisguse, line 20: Fancy :)

src/main/res/values/colors.xml
```diff
@@ -30,6 +30,9 @@ limitations under the License.
     <color name="colorAccent">@android:color/holo_orange_dark</color>
     <color name="colorAccentTransparent">#55FF8800</color>
 
+    <color name="colorBgDelete">#F44336</color>
```
dennisguse, line 33: color names are not obivous

dennisguse (commented):
Just did some scrolling.
Looks kind of nice!

PS/ What happens if I delete all profiles?

src/androidTest/java/de/dennisguse/opentracks/settings/PreferencesOpenHelperTest.java
rgmf: Okay, I did it, but I don't understand why must not reference to ITEM_SEPARATOR :thinking: because if we change it then we have to adjust all tests.

src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
rgmf: The first one was Map.
What would be your favorite order?

I've decided this:

1. Profiles
2. Map
3. Insert Marker

because I use the more profiles to change between running and cycling, then I usually use the map and I seldom add markers.

Anyway, I don't care about this. Because of this, let me know what the order should be and I'll do it ;)

rgmf:
> PS/ What happens if I delete all profiles?

That's not possible because swipe to delete is disabled when the recycler view only contains one item.

```java
RecyclerViewSwipeDeleteCallback recyclerViewSwipeDeleteCallback = new RecyclerViewSwipeDeleteCallback(this) {
            @Override
            public int getMovementFlags(@NonNull RecyclerView recyclerView, @NonNull RecyclerView.ViewHolder viewHolder) {
                // When there's only one profile it cannot be deleted (so, "disable" movements: drag flags and swipe flags).
                return adapter.getItemCount() > 1 ? makeMovementFlags(0, ItemTouchHelper.LEFT) : makeMovementFlags(0, 0);
            }
....
```

Anyway, now that you mention it, we can add some logic in `PreferencesUtils.getCustomLayout` to ensure that there's always a layout to return. What do you think?

src/androidTest/java/de/dennisguse/opentracks/settings/PreferencesOpenHelperTest.java
dennisguse: More a philosophical thing: tests should not rely on implementation details (in this case CsvConstants).
We do this in some place, but it is nicer to not do it, i.e., if somebody changes ITEM_SEPARATOR the tests should fail (makes the tests also easier to read).

There is a tradeoff writing tests and not writing them, as they require maintenance :man_shrugging:

rgmf: Okay, I understand. Thank you so much for the clarification ;)

src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
dennisguse: IMHO: map, insert marker, profiles.

(I often open the map and then close it again)

dennisguse:
Regarding swipe to delete: can we get this for TrackListActivity as well? :sun_with_face:

rgmf:
> Anyway, now that you mention it, we can add some logic in `PreferencesUtils.getCustomLayout` to ensure that there's always a layout to return. What do you think?

@dennisguse `PreferencesUtils.getCustomLayout` always return a Layout:

```java
String csvCustomLayouts = getString(R.string.stats_custom_layouts_key, buildDefaultLayout());
```

If nothing in `R.string.stats_custom_layouts_key` (all deleted) then `buildDefaultLayout` is executed.

rgmf:
> Regarding swipe to delete: can we get this for TrackListActivity as well? sun_with_face

Yes, that was the idea, but for other PR. I'll open a new Issue for that.

rgmf:
If any complaints then I'll merge it this weekend. I've been using it for a couple of weeks and it works fine (I hope).

dennisguse:
@rgmf Will test tomorrow :D

rgmf:
> @rgmf Will test tomorrow :D

Then, no hurry. If you want to test it or check it out, it can wait ;)

src/main/java/de/dennisguse/opentracks/settings/PreferencesOpenHelper.java
```diff
@@ -1,34 +1,65 @@
...
+        PreferencesUtils.setInt(R.string.prefs_last_version_key, version);
+        for (int i = 1; i <= version; i++) {
+            switch (i) {
+                case 1:
+                    upgradeFrom0to1();
+                    break;
+                case 2:
+                    upgradeFrom1to2();
+                    break;
+            }
```
dennisguse, line 41: add default case with throw new RuntimeException: somebody like me might forget to add the new cases here :)

src/main/java/de/dennisguse/opentracks/util/RecyclerViewSwipeDeleteCallback.java
```diff
@@ -0,0 +1,90 @@
...
+
+    Context context;
```
dennisguse, line 21: protected?

src/androidTest/java/de/dennisguse/opentracks/settings/PreferencesOpenHelperTest.java
```diff
@@ -0,0 +1,118 @@
...
+        assertEquals(layouts.get(0).toCsv(), PreferencesUtils.getCustomLayout().toCsv());
+    }
+
+    @Test
+    public void test_upgradeFrom1To2_withOldVersion() {
+        // given the version 1
+        SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+        SharedPreferences.Editor editor = sharedPreferences.edit();
+
+        String oldCustomLayoutCsv = context.getString(R.string.activity_type_unknown) + ";distance,1,1,0;speed,1,1,0;";
```
dennisguse, line 53: Best guess: we should use a separate string for this; and something that is not localized.
Otherwise, we will get in trouble if a user switches the phone's language.

And for showing it in the UI, we need to actually translate this value.
And for this: what about "Default" rather than "Unknown"?

dennisguse (approved):
Looks awesome!
And works like a charm!

Just the complaint that we use localized strings for storing settings.
We already do this for the default activity and that is not good either :man_shrugging:

src/main/java/de/dennisguse/opentracks/util/RecyclerViewSwipeDeleteCallback.java
rgmf: In fact, we don't need the context like a class member (only a local variable).

src/androidTest/java/de/dennisguse/opentracks/settings/PreferencesOpenHelperTest.java
rgmf: I've only used literal strings on tests (for simplicity). In the rest of the code we are using localized strings.

src/main/java/de/dennisguse/opentracks/settings/PreferencesUtils.java
```diff
@@ -576,46 +574,80 @@ public static void setLayoutColumns(int columns) {
...
-        List<String> csvParts = Arrays.asList(csvCustomLayout.split(CsvConstants.LINE_SEPARATOR));
-        Layout layout = new Layout(csvParts.get(0));
-        for (int i = 1; i < csvParts.size(); i++) {
-            String[] fieldParts = csvParts.get(i).split(CsvConstants.ITEM_SEPARATOR);
-            layout.addField(fieldParts[0], DataField.getTitleByKey(resources, fieldParts[0]), fieldParts[1].equals(DataField.YES_VALUE), fieldParts[2].equals(DataField.YES_VALUE), fieldParts[0].equals(resources.getString(R.string.stats_custom_layout_coordinates_key)));
+        String csvCustomLayouts = getString(R.string.stats_custom_layouts_key, buildDefaultLayout());
+        String[] csvLines = csvCustomLayouts.split(CsvConstants.LINE_SEPARATOR);
+        String profileSelected = getString(R.string.stats_custom_layout_selected_profile_key, null);
+        if (profileSelected == null) {
+            return Layout.fromCsv(csvLines[0], resources);
```
dennisguse, line 595: csvLines[0] might not exist; why not just create a Layout instead of parse the CSV?

src/main/java/de/dennisguse/opentracks/settings/PreferencesOpenHelper.java
```diff
@@ -1,34 +1,67 @@
...
+        if (preferenceValue.isEmpty()) {
+            PreferencesUtils.setString(R.string.stats_custom_layouts_key, PreferencesUtils.buildDefaultLayout());
+        }
+    }
+
+    private void upgradeFrom1to2() {
+        String csvVersion1CustomLayout = PreferencesUtils.getString(R.string.stats_custom_layouts_key, PreferencesUtils.buildDefaultLayout());
+        ArrayList<String> parts = new ArrayList<>();
+        Collections.addAll(parts, csvVersion1CustomLayout.split(CsvConstants.ITEM_SEPARATOR));
+        if (!parts.get(1).matches("\\d+")) {
```
dennisguse, line 58: Index out of bounds check.

src/main/java/de/dennisguse/opentracks/fragments/StatisticsRecordingFragment.java
```diff
@@ -65,13 +65,10 @@ public static Fragment newInstance() {
             preferenceMetricUnits = PreferencesUtils.isMetricUnits();
         }
 
-        if (PreferencesUtils.isKey(R.string.stats_custom_layout_fields_key, key)) {
+        if (PreferencesUtils.isKey(R.string.stats_custom_layouts_key, key) || PreferencesUtils.isKey(R.string.stats_custom_layout_selected_profile_key, key)) {
```
dennisguse, line 68: What is the difference between those two keys?

src/main/java/de/dennisguse/opentracks/content/data/Layout.java
```diff
@@ -1,16 +1,61 @@
...
 import java.util.List;
+import java.util.stream.Collectors;
+
+import de.dennisguse.opentracks.R;
+import de.dennisguse.opentracks.settings.PreferencesUtils;
+import de.dennisguse.opentracks.util.CsvConstants;
 
-public class Layout {
+public class Layout implements Parcelable {
     private final String profile;
```
dennisguse, line 17: What is the profile? It is the name for the layout?
If yes, then I would rather use name.

src/main/java/de/dennisguse/opentracks/content/data/Layout.java
```diff
@@ -1,16 +1,61 @@
...
+    private int columnsPerRow;
     private final List<DataField> dataFields = new ArrayList<>();
 
+    public static Layout fromCsv(String csvLine, Resources resources) {
+        List<String> csvParts = Arrays.asList(csvLine.split(CsvConstants.ITEM_SEPARATOR));
+        Layout layout = new Layout(csvParts.get(0), Integer.parseInt(csvParts.get(1)));
```
dennisguse, line 23: Index out of bounds check.

src/main/java/de/dennisguse/opentracks/content/data/Layout.java
```diff
@@ -1,16 +1,61 @@
...
+        for (int i = 2; i < csvParts.size(); i++) {
+            String[] fieldParts = csvParts.get(i).split(CsvConstants.PROPERTY_SEPARATOR);
```
dennisguse, line 25: Index out of bounds check.

dennisguse (approved):
I am slightly confused about some things:
* What does Layout.profile contain? Is it just a user-chosen name or is this somehow related with the default activity type?
* How is the default layout set?
* How does OpenTracks decide which layout to pick if I start a recording?
  Is it just the first one? Or is the activity type somehow used?

rgmf:
> I am slightly confused about some things:
> 
> * What does Layout.profile contain? Is it just a user-chosen name or is this somehow related with the default activity type?

The name of the layout.

> * How is the default layout set?

In settings.xml (from line 346) I've added a set of arrays with a default layout.

> * How does OpenTracks decide which layout to pick if I start a recording?
>   Is it just the first one? Or is the activity type somehow used?

OpenTracks always uses the last user selection.

dennisguse:
@rgmf Understood :)
Then I am in favor of replacing Layout.profile by Layout.name (or layoutName) + add a comment that is user-generated.
And it is good to go from my side (index checks would be great though).

What about merging and then we can all test it again via the nightly and then release by the end of next week?

src/main/res/values/settings.xml
```diff
@@ -46,6 +46,9 @@
     <string name="settings_import" translatable="false">settingsImport</string>
     <string name="settings_export" translatable="false">settingsExport</string>
 
+    <!-- Stats layout customization default profile -->
+    <string name="stats_custom_layout_default_profile">Default Profile</string>
```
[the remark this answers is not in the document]
rgmf: Yes, you are right, it must be in strings.xml

src/main/res/values/settings.xml
```diff
@@ -310,11 +313,11 @@
     <string name="stats_custom_layout_key" translatable="false">statsCustomLayoutKey</string>
 
     <!-- Stats layout customization: number of columns per row -->
-    <string name="stats_custom_layout_columns_key" translatable="false">statsCustomLayoutColumnsKey</string>
     <integer name="stats_custom_layout_columns_default" translatable="false">2</integer>
 
     <!-- Stats layout customization: key -->
-    <string name="stats_custom_layout_fields_key" translatable="false">statsCustomLayoutFieldsKey</string>
+    <string name="stats_custom_layouts_key" translatable="false">statsCustomLayoutFieldsKey</string>
+    <string name="stats_custom_layout_selected_profile_key" translatable="false">statsCustomSelectedLayoutKey</string>
```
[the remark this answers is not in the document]
rgmf: From TrackRecordingActivity (menu).

src/main/res/values/strings.xml
```diff
@@ -298,11 +299,13 @@ limitations under the License.
     <string name="menu_select_all">Select all</string>
     <string name="menu_settings">Settings</string>
     <string name="menu_share">Share</string>
+    <string name="menu_select_layout">Select Layout</string>
     <string name="menu_show_on_map">Show on map</string>
     <string name="menu_start_gps">Start GPS</string>
     <string name="menu_stop_gps">Stop GPS</string>
     <string name="menu_voice_rate">Voice speed</string>
     <string name="menu_resume_track">Resume track</string>
+    <string name="menu_add_custom_layout_profile">New Profile</string>
```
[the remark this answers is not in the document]
rgmf: We had a custom layout and we now have several profiles that results in several custom layouts. In the end is the same thing, so I'm going to call everything layout.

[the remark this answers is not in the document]
rgmf: The new strings are the following:

```xml
<string name="menu_add_custom_layout">New Layout</string>

<!-- Custom Layout -->
    <string name="custom_layout_list_title">Layouts</string>
    <string name="custom_layout_list_edit_name">Type layout\'s name to add</string>
    <string name="custom_layout_list_edit_already_exists">Layout\'s name already exists</string>
    <string name="custom_layout_list_layout_removed">Layout was removed</string>
    <string name="custom_layout_select_layout">Select Layout</string>
    <string name="stats_custom_layout_default_layout">Default Layout</string>
``

src/androidTest/java/de/dennisguse/opentracks/settings/PreferencesOpenHelperTest.java
rgmf: csvLines always has at least one item because getString(R.string.stats_custom_layouts_key, buildDefaultLayout()) alwasy returns at least one layout.

src/main/java/de/dennisguse/opentracks/settings/PreferencesOpenHelper.java
rgmf: It never should occur but I've added check anyway.

src/main/java/de/dennisguse/opentracks/fragments/StatisticsRecordingFragment.java
rgmf: `stats_custom_layouts_key`: it contains all layouts in a CSV way.
`stats_custom_layout_selected_layout_key`: it contains last selected layout in a CSV way.

Every time the user select a layout then it's copied to `stats_custom_layout_selected_layout_key` that will be use when user starts a new recording.

src/main/java/de/dennisguse/opentracks/content/data/Layout.java
rgmf: Yes, it's already fixed. You were right, it was a messy. Every reference to "profile" in the code has been modified by "name" or "layoutName".

src/main/java/de/dennisguse/opentracks/content/data/Layout.java
rgmf: That would be a developer error. How could we handle this kind of mistakes? The error would come from another part and we should launch an Exception or something like this, right? Malformed CSV Layout or something like this.

rgmf:
> @rgmf Understood :) Then I am in favor of replacing Layout.profile by Layout.name (or layoutName) + add a comment that is user-generated. And it is good to go from my side (index checks would be great though).
> 
> What about merging and then we can all test it again via the nightly and then release by the end of next week?

Okay. But I need your feedback on https://github.com/OpenTracksApp/OpenTracks/pull/1012#discussion_r763252352

src/main/java/de/dennisguse/opentracks/content/data/Layout.java
dennisguse: What about returning the default layout and print a log message with severity error?

src/main/java/de/dennisguse/opentracks/content/data/Layout.java
dennisguse: Same here: https://github.com/OpenTracksApp/OpenTracks/pull/1012#discussion_r763366589

Issue #952, opened 2021-09-17
Nightly improvements


dennisguse:
* [ ] install git
* [x] use black icon background
* [ ] (also for debug) change background Color of splash screen
* [x] append nightly to app name

pstorch:
I've got the build to run on my little vserver and setup a cronjob for 1 a.m. to check for new commits and build a new nightly apk.
Let's see if it is working.
In case someone has already installed the first test nightly from my repo, you need to uninstall it first. I had to change the signing key.

Repo (still beta): https://fdroid.storchp.de/fdroid/repo/

dennisguse:
@pstorch :heart_eyes: 
Just installed the nightly on my phone (in parallel the other two installations ><).
Now I am curiously waiting for the next just released version...

pstorch:
First successful build via cronjob last night 🎉

dennisguse:
@pstorch Got the update in the morning!
Nice work;

dennisguse:
Now we have a black background for the launcher icon, so I distinguish all three variants :D
Staying tuned for the new nightly build :partying_face:

pstorch:
I got a black icon 🎉

dennisguse:
@pstorch PR for the webpage welcome :D

PR #1014, opened 2021-11-09
Set distinct versionCode and name for nightlies

pstorch:
I'm experimenting with my own fdroid repository: https://fdroid.storchp.de/fdroid/repo

Currently it's all manually build and an apk only repository. But one can add it to the F-Droid client and would receive updates for nightlies. This needs to be automated and metadata be added.

What we also need is a nightly distinct versionCode and apk name. That's the reason for this PullRequest. I set the versionCode to the current date in the format YYYYMMDD and the apk filename to applicationId + _ + versionCode.

Probably the GitHub action need to be adjusted as well.

For #952 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,6 +30,12 @@ def getVersionName = { ->
     }
 }
 
+def versionCodeFromDate = { ->
+    def date = new Date()
+    def formattedDate = date.format('yyyyMMdd')
+    return Integer.valueOf(formattedDate)
+}
+
 android {
     compileOptions {
         coreLibraryDesugaringEnabled true
@@ -41,7 +47,7 @@ android {
     compileSdkVersion 31
 
     buildFeatures {
-        viewBinding = true
+        viewBinding true
     }
 
     aaptOptions {
@@ -82,7 +88,6 @@ android {
         nightly {
             signingConfig signingConfigs.nightly
             applicationIdSuffix ".nightly"
-            versionNameSuffix "-nightly"
         }
 
         release {
@@ -97,6 +102,14 @@ android {
 
     applicationVariants.all { variant ->
         variant.resValue "string", "applicationId", variant.applicationId
+
+        if (variant.buildType.name == 'nightly') {
+            variant.outputs.all {
+                setVersionCodeOverride(versionCodeFromDate())
+                setVersionNameOverride(getVersionName())
+                outputFileName = "${applicationId}_${variant.versionCode}.apk"
+            }
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -118,3 +139,25 @@ dependencies {
 
     androidTestUtil 'androidx.test:orchestrator:1.4.0'
 }
+
+
+
+// get the most recent git commit ID in long and short form
+class GitCommit {
```
dennisguse, line 146: Why not use?
Same as the stuff above.

```
def getVersionCode = { ->
    try {
        def stdout = new ByteArrayOutputStream()
        exec {
            commandLine 'git', 'rev-parse', 'HEAD`'
            standardOutput = stdout
        }
        return stdout.toString().trim()
    } catch (ignored) {
        return null
    }
}
```

dennisguse: and `git rev-parse --short HEAD`

we could also generalize the execute and parsing code.

build.gradle
```diff
@@ -97,6 +110,14 @@ android {
 
     applicationVariants.all { variant ->
         variant.resValue "string", "applicationId", variant.applicationId
+
+        if (variant.buildType.name == 'nightly') {
```
dennisguse, line 114: not a fan of this if; is there not a nicer place?
like in the ` nightly {}`-section?

build.gradle
```diff
@@ -82,7 +95,7 @@ android {
         nightly {
             signingConfig signingConfigs.nightly
             applicationIdSuffix ".nightly"
-            versionNameSuffix "-nightly"
+            versionNameSuffix "-nightly-" + GitCommit.currentShort()
```
dennisguse, line 98: is overriden again later?

dennisguse (approved):
Works for me :)
Some nitpicking, but it is just build code :sunglasses: 

PS/ would love to see nightlies :D

build.gradle
pstorch: That can be removed indeed. Leftover from my previous attempts.

build.gradle
pstorch: When I put it into the nightly {} block I get an build failed error:

image

Took me some time to find this solution. I'm also not a fan of it.

pstorch: If someone has a better solution, it is more than welcome.

dennisguse:
@pstorch If it works for you, just merge it :)

PS/ somehow Github actions are not triggered for PRs from forks :man_shrugging:

build.gradle
dennisguse: If it works, it works :)

PR #1016, opened 2021-11-11
Bugfix: running sensor data total is distance is UINT_32.

dennisguse:
End of cycling season and I started running again using a shoe-based running sensor.
... and reported distance got negative happened 2 or 3 times.
And today, I checked the parsing code and is an overflow that happened every 6.5km (UINT16 in decimeter).
But it should be a UINT32 :/

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/util/BluetoothUtilsTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/util/BluetoothUtilsTest.java
@@ -93,8 +93,8 @@ public class BluetoothUtilsTest {
 
     @Test
     public void parseRunningSpeedAndCadence_with_distance() {
-        BluetoothGattCharacteristic characteristic = new BluetoothGattCharacteristic(BluetoothUtils.CYCLING_SPEED_CADENCE_MEASUREMENT_CHAR_UUID, 0, 0);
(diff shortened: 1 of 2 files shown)


dennisguse:
@rgmf @pstorch After merging this, I would create v3.22.0 - any more stuff to merge?

rgmf:
> @rgmf @pstorch After merging this, I would create v3.22.0 - any more stuff to merge?

Go ahead.

dennisguse:
Will do after a test run tomorrow :)

rgmf:
> Will do after a test run tomorrow :)

Have a good run ;)

dennisguse:
PS: I just want to facepalm myself by about this issue :/
And the parsing for the cycling sensor is also broken......

Stay tuned

Note to myself: don't trust your own work.
And

dennisguse:
Post mortem to this issue:
The data via BLE is encoded as 4 byte and so far we only decoded the first 2 bytes (UINT16).
As this contains the LSO part, it seemed correct while testing, but leads to an overflow that is not handled properly (due to ignoring the MSO part).
In addition, we are not interested in the total distance of the sensor, but the total distance of the current track.

From the spec:
```
 The fields in the above table are in the order of LSO to MSO. Where LSO = Least Significant Octet and MSO = Most Significant Octet. 
```

@rgmf @FYI

dennisguse:
Works :D

Issue #1015, opened 2021-11-10
Current time field


vsevjednom-cz:
**Is your feature request related to a problem? Please describe.**
Sometimes I want to see the current time when I'm going somewhere. It's annoying to look on the small clock at the top right of the phone.

**Describe the solution you'd like**
New field called "Hours". It can only display the current time.

vsevjednom-cz:
It's amazingly amazing. Thank you!!!

Issue #936, opened 2021-09-03
Multi-column stats


JochenLutz:
**Is your feature request related to a problem? Please describe.**
When cycling, the most important stat for me is current speed. While the new customizable layout really helps to quickly capture the current speed, having the speed value span all columns (with increase font size), would further improve cycling safety.

**Describe the solution you'd like**
In the “Customize your recording layout“, in the settings of a stat item, add an option to make that item span all columns. Items spanning all columns should get bigger font — font size increase between new multi-column and primary items should be comparable to increase between primary and secondary items.

**Additional context**
![Quick mockup](https://user-images.githubusercontent.com/8270973/132055418-9c529dfc-20bf-4294-aa2a-0be43cc85829.png)

rgmf:
It looks like a feasible improvement and I think it'd be nice.

Only for clarification: you expect three sizes for data fields, don't you?:
- Small for secondary.
- Big for primary.
- Extra-big for spanned.

JochenLutz:
Exactly.

PR #1019, opened 2021-11-13
New field for recording layout: clock (current time field).

rgmf:
**Describe the pull request**
Added a new field for recording layout: clock.
It uses `TextClock` so layout to be loaded with this field is different.

There are four kind of fields:
- Short (100).
- Wide (200).
- Clock short (1 + 100).
- Clock wide (1 + 200).

We have room for 99 short custom fields and 99 wide custom fields.
For now, only (lat, lon) is wide but the code is ready for #936.

**Link to the the issue**
#1015 

**License agreement**


src/main/java/de/dennisguse/opentracks/adapters/CustomLayoutFieldType.java
```diff
@@ -1,8 +1,9 @@
 package de.dennisguse.opentracks.adapters;
 
 enum CustomLayoutFieldType {
-    SHORT(0),
-    WIDE(1);
+    SHORT(100),
+    WIDE(200),
+    CLOCK(1);
```
dennisguse, line 6: Feels like clock is different than short and wide.
Would a separate Enum be better in the long run?

dennisguse (approved):
Looks alright.
Did not try it...

src/main/java/de/dennisguse/opentracks/adapters/CustomLayoutFieldType.java
rgmf: Okay, let me give it a spin

PR #1021, opened 2021-11-15
Replace tts_fallback.mp3

vsevjednom-cz:
**Describe the pull request**
Original tts_fallback.mp3 is too quiet. So I "normalized" it with Waves L3 Multimaximizer. I worked in radio as presenter/DJ few years ago - so I still use some plugins for nice normalize.

**Link to the the issue**
#862


dennisguse:
@vsevjednom-cz Adding a file via upload is almost cheating ;)
Thanks :partying_face:

Issue #1027, opened 2021-11-18
Ignore locations if distance to previous location is too close (even when using a distance sensor)


dennisguse:
i was running with a shoe-based running sensor that disconnected while running.
And we store way too many points in between (as long as the sensor is connected).
Not useful for anything.

Other aspects:
* [ ] sensor did not reconnect (needed to stop the track and start a new one)
* [ ] after sensor disconnected, distance and speed were NOT taken from GPS although they should have been, right?

![Screenshot_20211118-201936_OSM_Dashboard_for_OpenTracks](https://user-images.githubusercontent.com/2820825/142483333-18f10789-48e7-4140-8ddd-d89719c65610.png)

PR #1028, opened 2021-11-18
Sensor issues while recording #1027

dennisguse:


src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
```diff
@@ -161,7 +161,7 @@ boolean onNewTrackPoint(TrackPoint trackPoint) {
                 return true;
             }
 
-            if (trackPoint.hasSensorData() || (distanceToLastTrackLocation.greaterOrEqualThan(recordingDistanceInterval) && trackPoint.isMoving())) {
+            if (distanceToLastTrackLocation.greaterOrEqualThan(recordingDistanceInterval) && trackPoint.isMoving()) {
```
rgmf, line 164: Could it result in a low heart rate or cadence precision? Shouldn't we compute the average or something like this between last point recorded and this one?

rgmf (changes requested):

src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
dennisguse: You are totally right - we are should do some averaging there I guess.
We were never doing it correctly, as we always stored the most recent reported sensor value when a GPS location was received.
The rest is dropped.

I created a new issue for this: #1029.

@rgmf Would it be okay for you to ignore this for now?

dennisguse:
I need to do some testing with the reconnect as the autoconnect option should already trigger an automatic reconnect.
It just felt as if this was not taking place.

src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
rgmf: It would be okay for me. In fact, separate this, sounds like a good idea.

rgmf (approved):

dennisguse:
Just tested the re-connect (i.e., by pulling the battery from the sensor): autoReconnect works.
So, I removed the change that manually triggered the re-connect after disconnect.

One side effect of this change is that whenever a sensor disconnects, we drop all _unsaved_ data - at the moment this is only relevant for distance.

dennisguse:
@pstorch Happily waiting for the new nightly build :partying_face:

PR #1030, opened 2021-11-20
App strings reworked

comradekingu:
https://weblate.bubu1.eu/ is down for the time being.
settings.xml should be translatable.

### DIFF
--- a/src/main/res/values/settings.xml
+++ b/src/main/res/values/settings.xml
@@ -224,7 +224,7 @@
 
     <string name="stats_rate_speed_or_pace_default" translatable="false">ACTIVITY_TYPE</string>
     <string name="stats_rate_pace" translatable="false">Pace</string>
-    <string name="stats_rate_speed" translatable="false">Spped</string>
+    <string name="stats_rate_speed" translatable="false">Speed</string>
     <string-array name="stats_rate_values">
         <item>@string/stats_rate_speed_or_pace_default</item>
         <item>@string/stats_rate_speed</item>
--- a/src/main/res/values/settings.xml
+++ b/src/main/res/values/settings.xml
@@ -12,39 +12,39 @@
 
     <!-- Main settings -->
     <string name="settings_open_tracks_key" translatable="false">settingsOpenTracks</string>
-    <string name="settings_open_tracks_title" translatable="false">Open Tracks</string>
-    <string name="settings_open_tracks_summary">All About OpenTracks Project</string>
+    <string name="settings_open_tracks_title">About</string>
+    <string name="settings_open_tracks_summary">All about the OpenTracks project</string>
 
     <string name="settings_defaults_key" translatable="false">settingsDefaults</string>
-    <string name="settings_defaults_title">Default Units and Activity</string>
-    <string name="settings_defaults_summary">Preferred Units, Default Activity, Track Default Name</string>
+    <string name="settings_defaults_title">Default units and activity</string>
+    <string name="settings_defaults_summary">Preferred units, default activity, track default name</string>
 
     <string name="settings_ui_key" translatable="false">settingsUI</string>
-    <string name="settings_ui_title">User Interface</string>
-    <string name="settings_ui_summary">Appearance, Recording Layout, Theme</string>
+    <string name="settings_ui_title">Interface</string>
+    <string name="settings_ui_summary">Appearance, recording layout, theme</string>
 
     <string name="settings_gps_key" translatable="false">settingsGps</string>
     <string name="settings_gps_title">GPS</string>
-    <string name="settings_gps_summary">Time interval, Distance interval, Max Distance, Accuracy</string>
+    <string name="settings_gps_summary">Time interval, distance interval, max. distance, accuracy</string>
 
     <string name="settings_sensors_key" translatable="false">settingsSensorsGps</string>
     <string name="settings_sensors_title">Bluetooth Sensors</string>
-    <string name="settings_sensors_summary">Cycling Sensors, Wheel Circumference, Running Sensors</string>
+    <string name="settings_sensors_summary">Cycling sensors, wheel circumference, running sensors</string>
 
     <string name="settings_announcements_key" translatable="false">settingsAnnouncements</string>
     <string name="settings_announcements_title">Voice Announcements</string>
-    <string name="settings_announcements_summary">Time, Distance, Voice Speed</string>
+    <string name="settings_announcements_summary">Time, distance, voice speed</string>
 
     <string name="settings_import_export_key" translatable="false">settingsImportExport</string>
     <string name="settings_import_export_title">Import/Export</string>
-    <string name="settings_import_export_summary">Import, Export, Auto Export</string>
+    <string name="settings_import_export_summary">Import, export, auto-export</string>
 
     <!-- Setting's categories -->
     <string name="settings_units_category">Units</string>
-    <string name="settings_activity_defaults_category">Activity\'s Defaults</string>
+    <string name="settings_activity_defaults_category">Activity defaults</string>
 
     <string name="settings_reset_key" translatable="false">settingsReset</string>
-    <string name="settings_reset_summary">Reset Settings to Default Values</string>
+    <string name="settings_reset_summary">Reset settings to defaults</string>
     <string name="settings_layout_reset_key" translatable="false">settingsLayoutReset</string>
 
     <string name="settings_sensor_bluetooth_heart_rate_key" translatable="false">bluetoothSensor</string>
@@ -52,7 +52,7 @@
     <string name="settings_sensor_bluetooth_cycling_speed_key" translatable="false">bluetoothCyclingSpeedSensor</string>
     <string name="settings_sensor_bluetooth_cycling_power_key" translatable="false">bluetoothCyclingPowerSensor</string>
     <string name="settings_sensor_bluetooth_running_speed_and_cadence_key" translatable="false">bluetoothRunningSpeedAndCadenceSensor</string>
-    <string name="sensor_type_value_none" translatable="false">NONE</string>
+    <string name="sensor_type_value_none">None</string>
 
     <string name="settings_sensor_bluetooth_cycling_speed_wheel_circumference_key" translatable="false">bluetoothCyclingSpeedWheelCircumference</string>
     <string name="settings_sensor_bluetooth_cycling_speed_wheel_circumference_default" translatable="false">2135</string>
@@ -223,8 +223,8 @@
     </string-array>
 
     <string name="stats_rate_speed_or_pace_default" translatable="false">ACTIVITY_TYPE</string>
-    <string name="stats_rate_pace" translatable="false">PACE</string>
-    <string name="stats_rate_speed" translatable="false">SPEED</string>
+    <string name="stats_rate_pace" translatable="false">Pace</string>
+    <string name="stats_rate_speed" translatable="false">Spped</string>
     <string-array name="stats_rate_values">
         <item>@string/stats_rate_speed_or_pace_default</item>
         <item>@string/stats_rate_speed</item>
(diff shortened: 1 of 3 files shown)


src/main/res/values/strings.xml
```diff
@@ -191,18 +189,18 @@ limitations under the License.
...
     <string name="description_total_distance">Total distance: %1$.2f km (%2$.1f mi)</string>
     <string name="description_total_time">Total time: %1$s</string>
     <string name="description_speed_source_gps">GPS</string>
     <!-- Export -->
     <string name="export_with_photos">with photos</string>
     <string name="export_without_photos">without photos</string>
-    <string name="export_error_post_workout">Post-workout export failed, please check the export directory.</string>
-    <string name="export_progress_message">Exporting to %1$s&#8230;</string>
-    <string name="export_track_already_exists_msg">The track %1$s already exists at the destination directory.</string>
+    <string name="export_error_post_workout">Could not export the post-workout. Check the export folder.</string>
```
pstorch, line 201: I think the meaning changed now. It's not a workout after a workout (post-workout), it's an export after the workout (post-workout-export).

comradekingu: OK. What is being exported, and is it important that it happens "post-workout"?

src/main/res/values/strings.xml
```diff
@@ -147,13 +145,13 @@ limitations under the License.
...
     <string name="activity_type_seaplane">seaplane</string>
-    <string name="activity_type_skate_boarding">skate boarding</string>
+    <string name="activity_type_skate_boarding">skateboarding</string>
     <string name="activity_type_skating">skating</string>
     <string name="activity_type_skiing">skiing</string>
     <string name="activity_type_sky_jumping">sky jumping</string>
     <string name="activity_type_sled">sled</string>
     <string name="activity_type_snow_boarding">snow boarding</string>
-    <string name="activity_type_snow_shoeing">snow shoeing</string>
+    <string name="activity_type_snow_shoeing">snowshoeing</string>
```
comradekingu, line 154: Does this mess up the entries in settings.xml?

src/main/res/values/strings.xml
pstorch: Right after the workout is finished some users want to have the recorded track automatically exported. So yes, it is important that the export of the track happens post-workout.

src/main/res/values/settings.xml
```diff
@@ -223,8 +223,8 @@
     </string-array>
 
     <string name="stats_rate_speed_or_pace_default" translatable="false">ACTIVITY_TYPE</string>
-    <string name="stats_rate_pace" translatable="false">PACE</string>
-    <string name="stats_rate_speed" translatable="false">SPEED</string>
+    <string name="stats_rate_pace" translatable="false">Pace</string>
+    <string name="stats_rate_speed" translatable="false">Spped</string>
```
comradekingu, line 227:
```suggestion
    <string name="stats_rate_speed" translatable="false">Speed</string>
```

src/main/res/values/settings.xml
```diff
@@ -223,8 +223,8 @@
     </string-array>
 
     <string name="stats_rate_speed_or_pace_default" translatable="false">ACTIVITY_TYPE</string>
```
comradekingu, line 225: Not sure everything is user-facing here.

src/main/res/values/strings.xml
```diff
@@ -307,47 +305,47 @@ limitations under the License.
...
+    <string name="permission_bluetooth_failed">Grant OpenTracks permission to use Bluetooth first.</string>
+    <string name="permission_bluetooth_failed_rejected">Manually grant OpenTracks access to Nearby Devices in the system settings first.</string>
     <!-- Search -->
     <!-- Sensor State -->
     <string name="sensor_state_battery">Battery level</string>
     <string name="sensor_state_cadence">Cadence</string>
-    <string name="sensor_state_cadence_avg">Avg Cadence</string>
-    <string name="sensor_state_cadence_max">Max Cadence</string>
+    <string name="sensor_state_cadence_avg">Avg. Cadence</string>
+    <string name="sensor_state_cadence_max">Max. Cadence</string>
```
comradekingu, line 316: Went with "max." to have fewer errors in translation.

dennisguse:
@comradekingu Everyhing that was upcase cannot be changed as those are not translations, but we use in the preferences to the user's choice (i.e., marked as translatable false).
Everything else should not be in settings.xml, but be migrated to strings.xml.

pstorch:
I think in general, you shouldn't change texts with "translatable=false"

pstorch:
Sorry, closed accidentially.

src/main/res/values/strings.xml
dennisguse: Yes, it would as we store in the preferences directly (best guess we even store the localized string) as well as per track.
Seemingly, this was not abstracted away aka mixing up localization and data storage.

src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="export_error_post_workout">Could not export track following newly finished workout . Check the export folder.</string>
```

src/main/res/values/strings.xml
comradekingu: Will it work if I change it in both places?

comradekingu:
@pstorch Some of the strings include some language part, like "Google's" and so on.
Will move strings and fix things once I know what to do about the categories.

src/main/res/values/strings.xml
```diff
@@ -191,18 +189,18 @@ limitations under the License.
...
     <string name="description_total_distance">Total distance: %1$.2f km (%2$.1f mi)</string>
     <string name="description_total_time">Total time: %1$s</string>
     <string name="description_speed_source_gps">GPS</string>
     <!-- Export -->
     <string name="export_with_photos">with photos</string>
     <string name="export_without_photos">without photos</string>
-    <string name="export_error_post_workout">Post-workout export failed, please check the export directory.</string>
-    <string name="export_progress_message">Exporting to %1$s&#8230;</string>
-    <string name="export_track_already_exists_msg">The track %1$s already exists at the destination directory.</string>
+    <string name="export_error_post_workout">Could not export track following newly finished workout . Check the export folder.</string>
```
rgmf, line 201: There's an extra space after "workout", right? It should be:

`Could not export track following newly finished workout. Check the export folder.`

comradekingu: Good catch :)

comradekingu:
```suggestion
    <string name="export_error_post_workout">Could not export track following newly finished workout. Check the export folder.</string>
```

src/main/res/values/settings.xml
```diff
@@ -12,47 +12,47 @@
 
     <!-- Main settings -->
     <string name="settings_open_tracks_key" translatable="false">settingsOpenTracks</string>
-    <string name="settings_open_tracks_title" translatable="false">Open Tracks</string>
-    <string name="settings_open_tracks_summary">All About OpenTracks Project</string>
+    <string name="settings_open_tracks_title">About</string>
```
dennisguse, line 15: Let's move them to strings.xml first.

src/main/res/values/strings.xml
dennisguse: Sadly it won't work; if it is wrong, just change it; the problem is not that big.

dennisguse:
I moved the translatable strings into strings.xml.
They are now translatable via Weblate.

src/main/res/values/strings.xml
```diff
@@ -14,23 +14,23 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <resources>
-    <string name="about_preference_title">About OpenTracks</string>
-    <string name="about_description"><i>OpenTracks</i> is a sport tracking application that completely respects your privacy.</string>
-    <string name="about_url">Project web page: <a href="%1$s">%1$s</a></string>
-    <string name="about_version_name">Version name: %1$s</string>
-    <string name="about_version_code">Version code: %1$d</string>
+    <string name="about_preference_title">About</string>
+    <string name="about_description">A sport tracking buddy that respects your privacy.</string>
```
dennisguse, line 18: doesn't sound right; why buddy?

src/main/res/values/strings.xml
```diff
@@ -14,23 +14,23 @@ See the License for the specific language governing permissions and
...
+    <string name="about_url">Project website\n<a href="%1$s">%1$s</a></string>
```
dennisguse, line 19: Why removing the ":"

src/main/res/values/strings.xml
```diff
@@ -14,23 +14,23 @@ See the License for the specific language governing permissions and
...
+    <string name="about_version_name">Version name, %1$s</string>
+    <string name="about_version_code">Version %1$d</string>
```
dennisguse, line 21: Technically, it is the version code :man_shrugging:

src/main/res/values/strings.xml
comradekingu: It is easier to identify with people, and in terms of privacy that connection becomes meaningful.
A personified concern is less esoteric than just the concept.

dennisguse:
@comradekingu Could you split this one up into smaller chunks?
Some of these changes are great, some of them do not fit that well.
There are way too many changes to review to accept them together.

PS/ great that you work that!

src/main/res/values/strings.xml
```diff
@@ -513,18 +511,18 @@ limitations under the License.
...
     </plurals>
     <plurals name="voiceTotalDistanceMiles">
-        <item quantity="one">total distance 1 mile</item>
-        <item quantity="other">total distance %1$.2f miles</item>
+        <item quantity="one">total distance, 1 mile</item>
+        <item quantity="other">total distance, %1$.2f miles</item>
     </plurals>
     <!-- Waypoint Type -->
-    <string name="waypoint_type_atm">ATM</string>
+    <string name="waypoint_type_atm">Cash machine</string>
```
dennisguse, line 525: It's an ATM in most places, right?

src/main/res/values/strings.xml
comradekingu: I don't get the benefit to ~"versionCode" over what "version" is. Also never seen it.
Similarly, "MIME" is changed to "Media type" below, although in Android terms it should be the former.

dennisguse: Practically, the version code is only relevant for Android and for humans the version name - so we could even remove the version code completely from the settings.
The change of MIME is great - I would still put like this "Media type (MIME)".

src/main/res/values/strings.xml
comradekingu: Less dereferencing. Better flow of information. More visually appealing. Consistent with the rest.

src/main/res/values/strings.xml
dennisguse: Moving this change into a separate PR would be great; might be a nice place for others to give there opinion.
Plus: the fastlane files also needs to be changed.

src/main/res/values/strings.xml
comradekingu: It is in the US, whereas more people understand "cash machine" is my thinking.
It is possible to have both en_US and en_UK translations, but alas.

src/main/res/values/strings.xml
comradekingu: https://github.com/OpenTracksApp/OpenTracks/pull/1032

dennisguse:
@comradekingu this PR cannot be merged anymore; would you have the time to do these changes on our Weblate project?

comradekingu:
@dennisguse Or I can try to solve the conflict here?

dennisguse:
Null problemo :sunglasses:

comradekingu:
@dennisguse Try opening it now.

PR #1031, opened 2021-11-20
Change versionCode for nightlies to commit count

pstorch:
As suggested in https://github.com/OpenTracksApp/opentracksapp.github.io/pull/8#issuecomment-974642302

When this change gets merged, user who already use the nightlies from the repo need to uninstall once manually because the old versionCode scheme has a much higher number. Otherwise one would never see updates. But I guess we can communicate that, right?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,10 +30,17 @@ def getVersionName = { ->
     }
 }
 
-def versionCodeFromDate = { ->
-    def date = new Date()
-    def formattedDate = date.format('yyyyMMdd')
-    return Integer.valueOf(formattedDate)
+def getVersionCode = { ->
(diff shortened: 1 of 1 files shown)


dennisguse:
Looks good - better now than later.
Should not affect too many people as we have not yet promoted nightlys too much.

pstorch:
Ok, then let's go

pstorch:
Repo is updated with the latest nightly build. I had to execute it manual now and delete the older versions with the higher versionCode. If you don't uninstall the old version first you don't get the update button, but a "downgrade" possibility to the newer (suggested) version.

Now we also see the changelog of the last 5 commits :wink:

Issue #652, opened 2021-03-14
GPX export TrackStatistics using TrackStatsExtension


dennisguse:
For GPX, there is an extension to export statistics about a track.
We can include those without a problem:
```
<xsd:element name="Distance" type="Meters_t" minOccurs="0"/>
<xsd:element name="TimerTime" type="Seconds_t" minOccurs="0"/>
<xsd:element name="TotalElapsedTime" type="Seconds_t" minOccurs="0"/>
<xsd:element name="MovingTime" type="Seconds_t" minOccurs="0"/>
<xsd:element name="StoppedTime" type="Seconds_t" minOccurs="0"/>
<xsd:element name="MovingSpeed" type="MetersPerSec_t" minOccurs="0"/>
<xsd:element name="MaxSpeed" type="MetersPerSec_t" minOccurs="0"/>
<xsd:element name="MaxElevation" type="Meters_t" minOccurs="0"/>
<xsd:element name="MinElevation" type="Meters_t" minOccurs="0"/>
<xsd:element name="Ascent" type="Meters_t" minOccurs="0"/>
<xsd:element name="Descent" type="Meters_t" minOccurs="0"/>
```

https://www8.garmin.com/xmlschemas/TrackStatsExtension.xsd

dennisguse:
And we could export the activity/category via TripExtension's `TransportationMode`.
https://github.com/cpesch/RouteConverter/blob/796cbc79b08f2edb1694891a6b61417b9cc104fa/gpx/src/main/doc/TripExtensionsv1.xsd#L19

akanakis:
I would like to try to fix this issue, if no one else hasn't yet.

akanakis:
I'm really sorry, but for some personal reasons I can't get involve with the project right now. So I will unassign myself from this issue and let anyone else to fix it.

dennisguse:
@akanakis Don't worry!

rgmf:
@akanakis if you can get involve in the future it would be great... we'd help you on everything you need ;)

Great to know people interesting on OpenTracks :D

akanakis:
OK thank you very much!

PR #1033, opened 2021-11-21
Storing data without GPS

### LINKED ISSUE #500: Indoor training

dennisguse:
OpenTracks now tries to create TrackPoints from BLE sensors (i.e., the ones that provide distance) every second.
It is a really naive implementation and works nicely if GPS is **not** used.

The current implementation has a preference of sensor-based TrackPoints as those will be triggered more often.
This will very likely result in ignoring useful GPS locations (which are expansive to get from an energy perspective).

For this, we need some more magic (aka don't try to store sensor data every second).

Fixed #500.

* [x] decide on how to implement (change type or just use lat/lng null)
* [x] create a new type for BLE sensor-based trackpoints: `SENSORPOINT`
* [x] implement OSMDashboard change: https://github.com/OpenTracksApp/OSMDashboard/pull/125
* [x] Modify ExportImport test (create senor-based Trackpoints in addition)
* [x] Check behavior for GPX exports/ imports (as TrackPoints without location cannot be exported)
* [x] Export/Import KML/KMZ
* [x] ~Upgrade OSMDashboard protocol version OR~ wait for new OSMDashboard release: https://github.com/OpenTracksApp/OSMDashboard/releases/tag/v2.13
* [x] GPX must export cumulative data from SENSORPOINTs (i.e., distance and altitude gain/loss)
* [x] Add this feature to the README
* [x] Rebase onto #1047
* [x] Wait for release of https://github.com/OpenTracksApp/OpenTracks/pull/1012 (at least a 3 days after availability on F-Droid)

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ _OpenTracks_ is a sport tracking application that completely respects your priva
 * __Photos and Markers:__ mark interesting locations while tracking
 * __Export:__
   * export tracks either as KMZ (incl. photos), KML, or GPX
-  * export automatically after each recording (e.g., to sync)
+  * export automatically after each recording (e.g., to sync via [Nextcloud](https://nextcloud.com/))
   * avoid duplication: each exported file contain a random unique identifier (i.e., `opentracks:trackid`)
 * __Altitude:__
   * gain/loss via barometric sensor (if present)
@@ -89,6 +89,8 @@ _OpenTracks_ is a sport tracking application that completely respects your priva
   * cycling: cadence
   * cycling: power meter
   * running: speed and cadence
+  * support for BLE sensor training only (i.e., without GPS) for indoor training
+
   An overview of tested sensors: [README_TESTED_SENSORS.md](README_TESTED_SENSORS.md)
 
 ### Gadgetbridge integration
--- a/src/androidTest/java/de/dennisguse/opentracks/content/provider/CustomContentProviderUtilsTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/content/provider/CustomContentProviderUtilsTest.java
@@ -15,6 +15,13 @@
  */
 package de.dennisguse.opentracks.content.provider;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.when;
+
 import android.content.ContentResolver;
 import android.content.ContentUris;
 import android.content.ContentValues;
@@ -56,13 +63,6 @@ import de.dennisguse.opentracks.stats.SensorStatistics;
 import de.dennisguse.opentracks.stats.TrackStatistics;
 import de.dennisguse.opentracks.util.FileUtils;
 import de.dennisguse.opentracks.util.UUIDUtils;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-import static org.mockito.Mockito.when;
 
 /**
  * A unit test for {@link ContentProviderUtils}.
@@ -220,6 +220,8 @@ public class CustomContentProviderUtilsTest {
         assertEquals(0, tracksPointsCursor.getCount());
         markerCursor = contentResolver.query(MarkerColumns.CONTENT_URI, null, null, null, MarkerColumns._ID);
         assertEquals(0, markerCursor.getCount());
+
+        //TODO Close all cursors
     }
 
     /**
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
@@ -133,6 +133,7 @@ public class ExportImportTest {
 
         // A sensor-only TrackPoint
         mockBLESensorData(trackPointCreator, 15f, sensorDistance, 66f, 3f, 50f);
+        mockAltitudeChange(trackPointCreator, 1);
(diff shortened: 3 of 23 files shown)


dennisguse:
One simple option: 
* store sensor-based trackpoints if recording distance interval is fulfilled (relative to the previous stored [any] TrackPoint aka lastValidTrackPoint).
* store GPS-based trackpoints if recording distance interval is fulfilled (relative to the previously stored GPS-based TrackPoint)

dennisguse:
Alternative with longer storage intervals was mentioned here: https://github.com/OpenTracksApp/OpenTracks/issues/500#issuecomment-907650784

dennisguse:
@pstorch This PR affects OSMDashboard.
At the moment, we are creating TrackPoints of type `TRACKPOINT` that do not contain coordinates.
OSMDashboard handles this case using 0/0.

An alternative would be to introduce a new type like `SENSOR` (i.e., no coordinates), so that `TRACKPOINT` has always a coordinates.
Which option do you like more?

pstorch:
What should the OSMDashboard do with those trackpoints? Ignore? I assume that trackpoints with coordinates 0/0 will be displayed at the favourite NULL position in the ocean. I've not tested it.

dennisguse:
That's what happens and how I discovered that we need to talk about this :)
I guess ignore those TrackPoints as they contain no useful data (for now).
Data might be interesting if we would implement something like line color by speed.

Anyhow, it is a breaking change so we need to upgrade the protocol version.

dennisguse:
@ChristophWurst Did a test run today and it works :)
With some to be expected flaws :sun_with_face:

pstorch:
Is it an option to filter out those beforehand on OT side? I mean trackpoints without coordinates are not that useful on a map. 
If not filter out 0/0 would be easy on the Dashboard side but a little bit a hack. Some apps do it like that, but 0/0 is still a valid coordinate. Not very useful though.
Better to have a flag in the data to know that kind of trackpoints. Make it explicit.

dennisguse:
@pstorch I introduced a new TrackPoint type (SENSORPOINT, 2), but that requires a change in OSMDashboard: 
Current implementation to determine pause would not be working anymore:
```
        this.pause = type != null ? type != 0 : latitude == PAUSE_LATITUDE;
```

I see two options:
* modify the where to ignore Type=2; I propose allow list rather than disallow list 
  `        try (final Cursor cursor = resolver.query(data, projection, TrackPoint._ID + "> ? AND TrackPoint.TYPE IN (-2, -1, 0, 1)", new String[]{Long.toString(lastTrackPointId)}, null)) {```
* ignore type two during actual processing

Anyhow, the new type is a breaking change (for the current OSMDashboard release), so we need to upgrade the API version.

pstorch:
As long as we cannot do anything useful on the map with those trackpoints I would prefer the first option to add it to the where clause, then they are not even transferred.

dennisguse:
This code is ready for review and testing.

One problem.
Until now the GPX export would be "broken": SENSORPOINT TrackPoints are not exported (as they do not contain GPS coordinates).
However, SENSORPOINTs contain cumulative data (i.e., distance and altitude gain/loss), which then will not be exported.
As a quick workaround, I implemented the TrackStatsExtension #652.
Two solutions:
* remove cumulative data from GPX (i.e., `opentracks:distance`, `opentracks:gain`, and `opentracks:loss`) as it is only using TRACKPOINT data
* compute SENSORPOINT data and add it to the next TRACKPOINT.
  (probably the better solution, but it complicates the `GPXTrackExporter`).

@pstorch Due to the introduction of SENSORPOINT, a new release of OSMDashboard is required to ignore those.
I would also increase the protocol version to be on the safe side. What do you think?

@rgmf Let's merge+release the profile changes first as it provides more value :)

pstorch:
> @pstorch Due to the introduction of SENSORPOINT, a new release of OSMDashboard is required to ignore those.
> I would also increase the protocol version to be on the safe side. What do you think?

I don't think an increase in protocol version is necessary. The `type` field is already available. There is "only" a new value. The query works with the older OpenTracks.

pstorch:
Anything else to change in OSMDashboard? I think I should release a new version soon, because of the sensor delay bug.

dennisguse:
@pstorch To the best of my knowledge OSMDashboard is prepared for this PR.

dennisguse:
@pstorch one thought about the statistics for the share pic functionality of OSMDashboard: SENSORPOINTs contain cumulative sensor data (i.e., distance); is this a problem for OSMDashboard if those are ignored? [don't know how the stats are computed...

pstorch:
@dennisguse OSMDashboard reads the statistic provided by OpenTracks and displays them as is, no computation. So the question is how OpenTracks calculates them :wink:

dennisguse:
@pstorch Perfect :)
Just woke up and had this thought :D

rgmf (approved):
It works like a charm. Fantastic feature!!!

I've read the code and I've tried to understand. The first thing very well, the second one not as well :roll_eyes: Maybe, I've should been read it with more attention :rofl: 

Seriously now, great job ;)

dennisguse:
@rgmf Sounds great!
I still would like to wait for #1012 as downgrading from this is not possible (we would need two releases to ship the db downgrade code etc). Releasing only the layout functionality, allows people to be able to downgrade (if there is some trouble with something). After some waiting time, we can release this then :)

dennisguse:
Found and fixed a regression: for creating a marker, the lastStoredTrackPointWithLocation should be used as markers require a valid location.

Issue #1035, opened 2021-11-24
NPE reported via Google Play for v3.22.0


dennisguse:
Seemingly, the button delay in controller fragment should check if should still handle the event (is activity still in running)?

```
java.lang.NullPointerException: 
  at de.dennisguse.opentracks.ControllerFragment.lambda$onTouch$1$de-dennisguse-opentracks-ControllerFragment (ControllerFragment.java:178)
  at de.dennisguse.opentracks.ControllerFragment$$ExternalSyntheticLambda1.run (Unknown Source:4)
  at android.os.Handler.handleCallback (Handler.java:938)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loop (Looper.java:233)
  at android.app.ActivityThread.main (ActivityThread.java:7892)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:656)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:967)
```

PR #1036, opened 2021-11-24
Fix NPE #1035

dennisguse:


src/main/java/de/dennisguse/opentracks/ControllerFragment.java
```diff
@@ -198,8 +211,12 @@ public boolean onTouch(final View view, MotionEvent event) {
 
             buttonDelay = () -> {
                 view.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_CLICKED);
-                getCallback().recordStop();
-                ActivityUtils.vibrate(getContext(), 1000);
+                Context context = getContext();
```
pstorch, line 214: 2x getContext()? Maybe the `Context context`can be inlined in line 218.

src/main/java/de/dennisguse/opentracks/ControllerFragment.java
```diff
@@ -114,6 +118,12 @@ public void onResume() {
         trackRecordingServiceConnection.startConnection(getContext());
```
pstorch, line 118: Also Nullcheck this?

pstorch:
If `getContext()` and `getActivity()` in the `ControllerFragment` can be null, maybe the other occurrences should also be checked?

src/main/java/de/dennisguse/opentracks/ControllerFragment.java
dennisguse: not necessary as this will be created via onCreate.
The NPE's are happening in the delayed callbacks.

PR #1040, opened 2021-12-04
Aggregated stats for only selected tracks.

### LINKED ISSUE #1039: Feature request: Export all/selected tracks to CSV
Hello.

It would be great to have possibility to export all/selected tracks to simple CSV, which will contain basic info of the track (date, distance, avg speed, description, etc.). Every track as a line.

As a biker, sometimes I need some quick info, how many kilometers I've ridden since some exact date (e.g. for maintenace - how many km I have with new chain etc,). 
Or maybe do some other simple analytics which I am struggling to make from several GPX files. I haven't found any suitable SW to do so.

Other alternative could be to show Aggregated stats from selected tracks (maybe add some more stats).

It is different feature reqest then recent proposal to export trackpoints to CSV.

rgmf:
**Describe the pull request**
Aggregated stats has been extended to see aggregated stats only for tracks selected in the contextual menu from TrackListActivity.

Users that select tracks (open contextual menu) can see now a new item menu called "Aggregated Stats".

AggregatedStatisticsActivity expect an optional list of Track.Id and show only aggregated stats from those Track.Ids. If this list is empty or not exists then it shows aggregated stats from all tracks.

It partially fixes #1039 (see comment: https://github.com/OpenTracksApp/OpenTracks/issues/1039#issuecomment-985834549)

**Link to the the issue**
#1039

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/AggregatedStatisticsActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/AggregatedStatisticsActivity.java
@@ -5,11 +5,16 @@ import android.view.View;
 
 import androidx.lifecycle.ViewModelProvider;
 
+import java.util.List;
+
 import de.dennisguse.opentracks.adapters.AggregatedStatisticsAdapter;
+import de.dennisguse.opentracks.content.data.Track;
 import de.dennisguse.opentracks.databinding.AggregatedStatsBinding;
 import de.dennisguse.opentracks.viewmodels.AggregatedStatisticsModel;
 
 public class AggregatedStatisticsActivity extends AbstractActivity {
+
+    public static final String EXTRA_TRACK_IDS = "track_ids";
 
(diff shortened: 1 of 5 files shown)


dennisguse:
Regarding #1045: could we extend this to include this?

Technically, it is "just" adding a UI to define filters and the db part is almost done already, right?

rgmf:
> Regarding #1045: could we extend this to include this?
> 
> Technically, it is "just" adding a UI to define filters and the db part is almost done already, right?

I think it's only an UI subject. I think a filter date and category would be great. If you don't mind I'll try to do it.

Anyway, this PR can be merged because it's related with track list more than aggregated stats activity.

dennisguse:
Works for me :)

dizmoduck:
it will be a good idea if you could filter by description 

For example, under description I have the various running clubs and race numbers for the individual races I participate in

PR #1042, opened 2021-12-08
Store timezone offset when starting a recording

### LINKED ISSUE #301: Proper handling of timezone data
**Issue**: Many running trackers do not properly handle dates and times, especially when accounting for timezones.  If for example, I take a vacation to Europe and run in London at 8AM (GMT = UTC), and return home to new york (EST = UTC-5), it would show my london run time was at 3AM.  Furthermore, if I run in new york at 8AM during eastern standing time (UTC-5), when the clocks springs ahead for daylight savings time (DST = UTC-4), it would be good to see my running time still as 8AM instead of 9AM.  Keep in mind that the time is correctly stored already because its in UTC. Only my timezone offset has changed.

**Solution**: Proposing adding an additional column in tracks table for managing timezone offset of the current workout (tzoffset).  The timezone can be retrieved from phone.  The date and time is already stored properly using UTC regardless of the timzone because thats how GPS units timestamp the data.  Storing the timezone offset is also a good idea to prevent having to do daylight savings calculations which could get messy if daylight savings times are ever done away with.  For example, when I run in New York today, the tzoffset column can be stored as '-400'.  When the time is shown (starttime column), you could offset with offset hours+minutes.  I included the -400 instead of -4 because offsets can can also be in 30 minutes increments as is the case with Newfoundland in Canada.  So an offset of 4 hours and 30 minutes can be '430' (reference: https://en.wikipedia.org/wiki/List_of_UTC_time_offsets).  For historical workouts, if tzoffset column is NULL, consider it the current timezone.  This will apply to all activities going forward.

dennisguse:
This PR adds a time offset to the tracks; all times remain stored as Instant (i.e., UTC).

Data migration: existing Track will use current time offset of the phone.
Import: use the time offset from the KML/GPX

Note: time offset is stored in seconds and denotes the difference from UTC.

Note: PR is based upon #1033.

Fixes #301.

* [x] add offset column to Track table
* [x] add field to Track
* [x] add check in ExportImportTest (export/import for GPX and KML)
* [ ] use offset for UI renderings: StatisticsRecordedFragment
* [ ] use offset for naming tracks (default name option)
* [ ] use offset for UI renderings: TrackListActivity
* [ ] use offset for UI renderings: MarkerListActivity
* [x] rebase onto main after merging #1033

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackExporterTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/exporter/KmlTrackExporterTest.java
@@ -1,5 +1,7 @@
 package de.dennisguse.opentracks.io.file.exporter;
 
+import static org.junit.Assert.assertEquals;
+
 import android.content.Context;
 
 import androidx.test.core.app.ApplicationProvider;
@@ -10,11 +12,10 @@ import org.junit.runners.JUnit4;
 
 import java.io.ByteArrayOutputStream;
 import java.time.Instant;
+import java.time.ZoneOffset;
 
 import de.dennisguse.opentracks.content.data.TrackPoint;
 import de.dennisguse.opentracks.io.file.TrackFileFormat;
-
-import static org.junit.Assert.assertEquals;
 
 @RunWith(JUnit4.class)
 public class KmlTrackExporterTest {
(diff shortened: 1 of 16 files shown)


dennisguse:
@iointerrupt would have time to test this?
PS: only the internal data handling is implemented, so you need to record and export for now.
Using the time offset in the UI is still to be done.

rgmf:
@dennisguse in AggregatedStatisticsActivity we are filtering Tracks by datetime (using phone Locale Zone). Shall I review this to adapt to this PR? I mean, TrackSelection should handle starttime_offset, right?

dennisguse:
@rgmf Valid point; the UI is not yet adjusted, so the data is stored, but the UI doesn't use it.
For now, this is just used for export and imports.
If you have time for this that would be awesome.
I updated the issue and added this as an open point.

rgmf:
@dennisguse okay, I see you are asigned to #301. I have some time so if you don't mind or you want to do other things then change the assignation and I do the list of things you have written out here: https://github.com/OpenTracksApp/OpenTracks/issues/301#issuecomment-1002138664

dennisguse:
@rgmf feel free to :)
I guess, I will try to refactor the TrackRecordingService in the next (free) days.

PR #1049, opened 2021-12-15
fix persistable directory permissions

pstorch:
I was wondering why the export directory vanishes. I found a mistake in persisting the uri permission.
I also fixed the TODO comment to implement the release of the uri permission in case the directory chooser is canceled or a new directory is chosen.


dennisguse (approved):
Looks great!

Issue #1052, opened 2021-12-18
TrackRecordingService: service connection leaks


dennisguse:
I was able to reproduce again. Starting the track recording, push the power button, so the screen of you phone turns black and OT runs in the background. Wait until the first notification. Then unlock the phone again and stop the recording. Go back to the track list and you see that there is still a track recording running.

This is the logcat:

```

2021-12-17 14:48:34.411 29612-29612/de.dennisguse.opentracks.debug E/ActivityThread: Service de.dennisguse.opentracks.services.TrackRecordingService has leaked ServiceConnection android.speech.tts.TextToSpeech$Connection@284d451 that was originally bound here
    android.app.ServiceConnectionLeaked: Service de.dennisguse.opentracks.services.TrackRecordingService has leaked ServiceConnection android.speech.tts.TextToSpeech$Connection@284d451 that was originally bound here
        at android.app.LoadedApk$ServiceDispatcher.<init>(LoadedApk.java:1941)
        at android.app.LoadedApk.getServiceDispatcherCommon(LoadedApk.java:1813)
        at android.app.LoadedApk.getServiceDispatcher(LoadedApk.java:1792)
        at android.app.ContextImpl.bindServiceCommon(ContextImpl.java:1899)
        at android.app.ContextImpl.bindService(ContextImpl.java:1829)
        at android.content.ContextWrapper.bindService(ContextWrapper.java:774)
        at android.speech.tts.TextToSpeech.connectToEngine(TextToSpeech.java:849)
        at android.speech.tts.TextToSpeech.initTts(TextToSpeech.java:819)
        at android.speech.tts.TextToSpeech.<init>(TextToSpeech.java:772)
        at android.speech.tts.TextToSpeech.<init>(TextToSpeech.java:751)
        at android.speech.tts.TextToSpeech.<init>(TextToSpeech.java:735)
        at de.dennisguse.opentracks.services.tasks.VoiceAnnouncement.start(VoiceAnnouncement.java:122)
        at de.dennisguse.opentracks.services.tasks.VoiceAnnouncementManager.restore(VoiceAnnouncementManager.java:62)
        at de.dennisguse.opentracks.services.tasks.VoiceAnnouncementManager.setFrequency(VoiceAnnouncementManager.java:105)
        at de.dennisguse.opentracks.services.TrackRecordingService$2.onSharedPreferenceChanged(TrackRecordingService.java:110)
[...]

PR #1053, opened 2021-12-19
Cleanup: reference resources manually.

dennisguse:
Using a generic lookup prevents compiler optimization.
... and the code was too hard for me figure out :)

Saw this while looking into the stacktrace from #1052.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/settings/PreferencesUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/settings/PreferencesUtils.java
@@ -32,12 +32,11 @@ import androidx.documentfile.provider.DocumentFile;
 import androidx.preference.PreferenceManager;
 
 import java.time.Duration;
-
-import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import de.dennisguse.opentracks.R;
 import de.dennisguse.opentracks.content.data.Distance;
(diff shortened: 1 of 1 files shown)


src/main/java/de/dennisguse/opentracks/settings/PreferencesUtils.java
```diff
@@ -552,30 +551,33 @@ public static int getLayoutColumnsByDefault() {
...
+                R.array.stats_custom_layout_fields_default_value_9,
+                R.array.stats_custom_layout_fields_default_value_10,
+                R.array.stats_custom_layout_fields_default_value_11,
+                R.array.stats_custom_layout_fields_default_value_12,
+                R.array.stats_custom_layout_fields_default_value_13,
+                R.array.stats_custom_layout_fields_default_value_14,
+                R.array.stats_custom_layout_fields_default_value_15,
+                R.array.stats_custom_layout_fields_default_value_16,
+                R.array.stats_custom_layout_fields_default_value_17,
+                R.array.stats_custom_layout_fields_default_value_18
```
rgmf, line 574: That piece of code is more clear and better to read.

But, if we decide to add a new field then we'll have to remember to add it into settings.xml and inside this stream in PreferencesUtils.java

I don't think we are going to add fields very often, so it's good for me.

rgmf (approved):

src/main/java/de/dennisguse/opentracks/settings/PreferencesUtils.java
rgmf: PS/ I've not tested it :D

dennisguse: :D

PR #1055, opened 2021-12-20
Filter option in Aggregated Stats

rgmf:
**Describe the pull request**
User can filter aggregated statistics by category (sport) and/or date range:

imagen

By default:
- The current week is selected (begin and end date with first day of this week and last day of this week).
- All categories are checked.

When user set a filter then AggregatedStatisticsActivity shows "clear filter" menu option instead of "filter" option to clear filters applied.

**Developers information**
FilterDialogFragment has been developed to be reused in other parts of OpenTracks if needed.

I've created an interface inside `ContentProviderUtils` called `ContentProviderSelectionInterface` that can be used to create selections from out of this class. Useful here for `Track` class to filter tracks and useful for other similar situations. In that way, `ContentProviderUtils.getTracks`, for example, can be re-used for different selections.

**Link to the the issue**
#1039

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/adapters/AggregatedStatisticsAdapter.java
+++ b/src/main/java/de/dennisguse/opentracks/adapters/AggregatedStatisticsAdapter.java
@@ -9,6 +9,9 @@ import android.widget.BaseAdapter;
 import android.widget.ImageView;
 import android.widget.TextView;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import de.dennisguse.opentracks.R;
 import de.dennisguse.opentracks.settings.PreferencesUtils;
 import de.dennisguse.opentracks.util.StringUtils;
@@ -17,7 +20,7 @@ import de.dennisguse.opentracks.viewmodels.AggregatedStatistics;
 
 public class AggregatedStatisticsAdapter extends BaseAdapter {
 
-    private final AggregatedStatistics aggregatedStatistics;
+    private AggregatedStatistics aggregatedStatistics;
     private final Context context;
 
     public AggregatedStatisticsAdapter(Context context, AggregatedStatistics aggregatedStatistics) {
@@ -64,6 +67,20 @@ public class AggregatedStatisticsAdapter extends BaseAdapter {
         }
 
         return convertView;
+    }
+
+    public AggregatedStatistics swapData(AggregatedStatistics aggregatedStatistics) {
+        this.aggregatedStatistics = aggregatedStatistics;
+        this.notifyDataSetChanged();
+        return aggregatedStatistics;
+    }
+
+    public List<String> getCategories() {
+        List<String> categories = new ArrayList<>();
+        for (int i = 0; i < aggregatedStatistics.getCount(); i++) {
+            categories.add(aggregatedStatistics.getItem(i).getCategory());
+        }
+        return categories;
     }
 
     private class ViewHolder {
--- a/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
@@ -73,6 +73,10 @@ public class ContentProviderUtils {
 
     private final ContentResolver contentResolver;
 
+    public interface ContentProviderSelectionInterface {
+        SelectionData buildSelection();
+    }
+
     public ContentProviderUtils(Context context) {
         contentResolver = context.getContentResolver();
     }
@@ -205,15 +209,10 @@ public class ContentProviderUtils {
         return tracks;
     }
 
-    public List<Track> getTracks(List<Track.Id> trackIds) {
-        if (trackIds == null || trackIds.isEmpty()) {
-            return getTracks();
-        }
-
-        String selection = String.format(TracksColumns._ID + " IN (%s)", TextUtils.join(",", Collections.nCopies(trackIds.size(), "?")));
-        String[] selectionArgs = trackIds.stream().map(id -> Long.toString(id.getId())).toArray(String[]::new);
+    public List<Track> getTracks(ContentProviderSelectionInterface selection) {
+        SelectionData selectionData = selection.buildSelection();
         ArrayList<Track> tracks = new ArrayList<>();
-        try (Cursor cursor = getTrackCursor(selection, selectionArgs, TracksColumns._ID)) {
+        try (Cursor cursor = getTrackCursor(selectionData.getSelection(), selectionData.getSelectionArgs(), TracksColumns._ID)) {
             if (cursor != null && cursor.moveToFirst()) {
                 tracks.ensureCapacity(cursor.getCount());
                 do {
--- a/src/main/java/de/dennisguse/opentracks/fragments/FilterDialogFragment.java
+++ b/src/main/java/de/dennisguse/opentracks/fragments/FilterDialogFragment.java
@@ -1,0 +1,188 @@
+package de.dennisguse.opentracks.fragments;
+
+import android.app.Dialog;
+import android.content.Context;
+import android.os.Bundle;
+import android.os.Parcel;
+import android.os.Parcelable;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.widget.CheckBox;
+import android.widget.DatePicker;
+import android.widget.GridLayout;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AlertDialog;
+import androidx.fragment.app.DialogFragment;
+import androidx.fragment.app.FragmentManager;
+
+import com.google.android.material.textfield.TextInputEditText;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.temporal.WeekFields;
+import java.util.ArrayList;
+import java.util.Locale;
+
+import de.dennisguse.opentracks.R;
+import de.dennisguse.opentracks.util.StringUtils;
+
+public class FilterDialogFragment extends DialogFragment {
+
+    private static final String TAG = FilterDialogFragment.class.getSimpleName();
+    public static final String KEY_FILTER_ITEMS = "filterItems";
+
+    private FilterDialogListener filterDialogListener;
+    private ArrayList<FilterItem> filterItems = new ArrayList<>();
+
+    public static void showDialog(FragmentManager fragmentManager) {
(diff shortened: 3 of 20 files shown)


dennisguse:
Did not read the code :)

Three things:
* categories might contain anything - the user can select something pre-defined, but is not limited to
* categories (at least the default one's) could have icons
* dates should contain years

rgmf:
> Did not read the code :)
> 
> Three things:
> 
>     * categories might contain anything - the user can select something pre-defined, but is not limited to

The list of categories come from the database, so user was the one who decided them.

> 
>     * categories (at least the default one's) could have icons

I could add the icon too.

> 
>     * dates should contain years

I'm using DateUtils.formatDateTime in the end. The dates you can see in the screenshot doesn't contain the year because it's the current year. If you select a 2020 date, then DateUtils.formatDate add the year.

src/main/java/de/dennisguse/opentracks/content/data/Track.java
```diff
@@ -177,4 +185,73 @@ public void writeToParcel(Parcel parcel, int i) {
...
     }
+
+    public static class Selection implements ContentProviderUtils.ContentProviderSelectionInterface {
+        private final List<Id> trackIds = new ArrayList<>();
+        private final List<String> categories = new ArrayList<>();
+        private Instant from;
+        private Instant to;
+
+        private Selection() {
+
```
dennisguse, line 196: empty line

src/main/java/de/dennisguse/opentracks/content/data/Track.java
```diff
@@ -177,4 +185,73 @@ public void writeToParcel(Parcel parcel, int i) {
...
+        }
+
+        public static Selection create() {
```
dennisguse, line 199: What s the benefit of a factory pattern here?

src/main/java/de/dennisguse/opentracks/content/data/Track.java
```diff
@@ -177,4 +185,73 @@ public void writeToParcel(Parcel parcel, int i) {
...
+
+        public Selection addCategory(String category) {
+            if (!this.categories.contains(category)) {
+                this.categories.add(category);
+            }
+            return this;
+        }
+
+        @Override
+        public Pair<String, String[]> buildSelection() {
```
dennisguse, line 224: Why not create a datatype for this pair?

src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
```diff
@@ -205,15 +210,10 @@ public void deleteTrack(Context context, @NonNull Track.Id trackId) {
...
     }
 
-    public List<Track> getTracks(List<Track.Id> trackIds) {
-        if (trackIds == null || trackIds.isEmpty()) {
-            return getTracks();
-        }
-
-        String selection = String.format(TracksColumns._ID + " IN (%s)", TextUtils.join(",", Collections.nCopies(trackIds.size(), "?")));
-        String[] selectionArgs = trackIds.stream().map(id -> Long.toString(id.getId())).toArray(String[]::new);
+    public List<Track> getTracks(ContentProviderSelectionInterface selection) {
```
dennisguse, line 213: Is this @NonNull?

src/main/java/de/dennisguse/opentracks/fragments/FilterDialogFragment.java
```diff
@@ -0,0 +1,186 @@
...
+
+            itemsLayout.addView(view);
+        }
+
+        DatePicker datePickerFrom = layout.findViewById(R.id.filter_date_picker_from);
+        DatePicker datePickerTo = layout.findViewById(R.id.filter_date_picker_to);
+        TextInputEditText dateFrom = layout.findViewById(R.id.filter_date_edit_text_from);
+        TextInputEditText dateTo = layout.findViewById(R.id.filter_date_edit_text_to);
+
+        CalendarUtil calendarFirstWeekDay = CalendarUtil.ofFirstDayOfTheWeek();
```
dennisguse, line 83: somehow this feels complicated.
No better suggestion yet :)

src/main/java/de/dennisguse/opentracks/util/CalendarUtil.java
```diff
@@ -0,0 +1,72 @@
+package de.dennisguse.opentracks.util;
+
+import android.content.Context;
+import android.text.format.DateUtils;
+import android.util.Log;
+
+import java.util.Calendar;
```
dennisguse, line 7: Not a fan of Calendar.
Could we use the newer classes for date handling?

src/main/java/de/dennisguse/opentracks/viewmodels/AggregatedStatisticsModel.java
```diff
@@ -21,18 +20,26 @@ public AggregatedStatisticsModel(@NonNull Application application) {
         super(application);
     }
 
-    public LiveData<AggregatedStatistics> getAggregatedStats(@Nullable List<Track.Id> trackIds) {
+    public LiveData<AggregatedStatistics> getAggregatedStats(Track.Selection selection) {
```
dennisguse, line 23: nullable?

src/main/java/de/dennisguse/opentracks/viewmodels/AggregatedStatisticsModel.java
```diff
@@ -21,18 +20,26 @@ public AggregatedStatisticsModel(@NonNull Application application) {
...
         return aggregatedStats;
     }
 
-    private void loadAggregatedStats(@Nullable List<Track.Id> trackIds) {
+    public void updateSelection(Track.Selection selection) {
+        loadAggregatedStats(selection);
+    }
+
+    public void clearSelection() {
+        loadAggregatedStats(Track.Selection.create());
```
dennisguse, line 36: Not yet sure, if Selection should be an inner class of Track.
It somehow feels like separate class that is used for selecting tracks.

src/main/res/values/strings.xml
```diff
@@ -239,6 +239,9 @@ limitations under the License.
     <string name="generic_click_twice_cancel">Click back again to cancel</string>
     <string name="generic_choose_an_option">Choose an option</string>
     <string name="generic_undo">Undo</string>
+    <string name="generic_filter">Filter</string>
+    <string name="generic_from_date">Begin Date</string>
```
dennisguse, line 243: From?

src/main/res/values/strings.xml
```diff
@@ -239,6 +239,9 @@ limitations under the License.
...
+    <string name="generic_to_date">End Date</string>
```
dennisguse, line 244: To?

dennisguse (approved):
Looks pretty good.
Did not try it out :)

dennisguse:
Nit picking: a UI-based (espresso) test would be awesome

rgmf:
> Nit picking: a UI-based (espresso) test would be awesome

Oohhh, yes, count on it.

src/main/java/de/dennisguse/opentracks/adapters/AggregatedStatisticsAdapter.java
```diff
@@ -66,6 +69,20 @@ public View getView(int position, View convertView, ViewGroup parent) {
...
 
+    public AggregatedStatistics swapData(AggregatedStatistics aggregatedStatistics) {
+        this.aggregatedStatistics = aggregatedStatistics;
+        this.notifyDataSetChanged();
+        return aggregatedStatistics;
+    }
+
+    public List<String> getCategories() {
+        List<String> categories = new ArrayList<>();
+        for (int i = 0; i < aggregatedStatistics.getCount(); i++) {
```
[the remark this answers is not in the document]
rgmf: I prefer not to do that because It would result in `AggregatedStatistics` modifications. Maybe, creating `getCategories` method inside `AggregatedStatistics`. Would it worth?

src/main/java/de/dennisguse/opentracks/content/data/Track.java
rgmf: I like more the syntax `Selection.create().addTrackId().addCategory()` than `new Selection().addTrackId().addCategory()`.

Anyway, re-reading my argument seems like a stupid one :thinking: :laughing:  so I'm going to change it.

src/main/java/de/dennisguse/opentracks/content/provider/ContentProviderUtils.java
rgmf: We already have `getTracks()` method so if someone uses `getTracks(selection)` s-he should pass a selection.

src/main/java/de/dennisguse/opentracks/fragments/FilterDialogFragment.java
rgmf: I'm going to think about this.

src/main/java/de/dennisguse/opentracks/viewmodels/AggregatedStatisticsModel.java
rgmf: I think so, because it's a track's selection. If you remove Track class then you have to remove Track.Selection one.

src/main/res/values/strings.xml
rgmf: Okay. So I rename generic_from_date to generic_from to make it even more generic :smile:

src/main/java/de/dennisguse/opentracks/viewmodels/AggregatedStatisticsModel.java
dennisguse: That's true, but not every Track has a selection and the selection is only relevant for the filtering.
So for me, the selection is some structure that helps us to find the correct tracks, but afterwards the tracks have nothing to with it.
And it somehow only is used by one feature (at the moment) while the tracks are used almost everywhere.

src/main/java/de/dennisguse/opentracks/util/CalendarUtil.java
rgmf: BTW: What do you do to keep up with Java news? Any webpage/resource/RSS?

I didn't know there were better classes to use instead of Calendar.

dennisguse:
@rgmf Happy X-Mas ;)

rgmf:
@dennisguse Merry Christmas! 🎄😀

dennisguse:
@rgmf Forgot to answer you in some of thread some days ago: https://www.baeldung.com/java-8-date-time-intro
And in general: this webpage is usually the go to place for knowledge around java.
Almost every time I search for something this page comes up :)

PR #1069, opened 2021-12-29
BLE: consider PineTime with InfiniTime to provide a heart rate sensor.

### LINKED ISSUE #1063: pine time
I am using pine time watch but it is not connected when I asked for heart rate connection. It couldn't find the watch. Can this implemented? Gadgetbridge has no problem and also fittotrack can connect the watch. So if this is a problem ? Here it is or is the watch not included in the software? I hope it will be soon implemented.

Yours

Johan**Is your feature request related to a problem? Please describe.**

**Additional context**

dennisguse:
Fixes #1063.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/util/BluetoothUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/util/BluetoothUtils.java
@@ -48,7 +48,8 @@ public class BluetoothUtils {
     public static final List<UUID> HEART_RATE_SUPPORTING_DEVICES = Collections.unmodifiableList(Arrays.asList(
             BluetoothUtils.HEART_RATE_SERVICE_UUID,
             //Devices that support HEART_RATE_SERVICE_UUID, but do not announce HEART_RATE_SERVICE_UUID in there BLE announcement messages (during device discovery).
-            UUID.fromString("0000fee0-0000-1000-8000-00805f9b34fb") //Miband3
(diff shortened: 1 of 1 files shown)


dennisguse:
@JohanvdS
@JF002 
Does this fix the issue?

JF002:
I've just checked : when I disable the option "Only show bluetooth devices that announce required servces", OpenTracks connects to Infinitime correctly (on main branch).

On this branch, with the option enabled, it doesn't list my pinetimes. I checked : the UUID is wrong, is must be read backwards:

```
00001530-1212-efde-1523-785feabcd123
```

However, I'm not sure this is a good idea to use this UUID, as it's not specific to InfiniTime or the PineTime : it's the UUID of the Nordic Semiconductor DFU service, which can be exposed by other devices (most probably those based on an NRF52 MCU).

I guess the best solution would be to add the HR UUID in the ADV packets sent by InfiniTime. What are the implications of adding multiple UUID in ADV?

I created a feature request in InfiniTime : https://github.com/InfiniTimeOrg/InfiniTime/issues/895. I don't think I'll be able to work in it soon, as I'm currently working on some memory optimizations right now...

dennisguse:
@JF002 nice catch that the UUID is the other way around :)
PS: moved the discussion to https://github.com/InfiniTimeOrg/InfiniTime/issues/895

PR #1075, opened 2022-01-01
Try to takkle down export dir issues

pstorch:
I'm still having problems with a nextcloud export directory. Sometimes (after new install of a nightly), the directory vanishes. Vanishes as in the settings activity doesn't show anything, even not "not set". So I changed the display of the name to the uri and added an indicator if it is writable.
I also added a check if the post-workout export is enabled, but the directory is null or not writable. 
Additionally I only release the uri permission, if the new selected directory is different to the previous one.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/settings/ImportExportSettingsFragment.java
+++ b/src/main/java/de/dennisguse/opentracks/settings/ImportExportSettingsFragment.java
@@ -66,7 +66,11 @@ public class ImportExportSettingsFragment extends PreferenceFragmentCompat {
         instantExportDirectoryPreference.setSummaryProvider(preference -> {
             DocumentFile directory = PreferencesUtils.getDefaultExportDirectoryUri(getContext());
             //Use same value for not set as Androidx ListPreference and EditTextPreference
-            return directory != null ? directory.getName() : getString(R.string.not_set);
+            if (directory == null) {
+                return getString(R.string.not_set);
+            }
+
+            return directory.getUri().toString() + (directory.canWrite() ? "" : getString(R.string.export_dir_not_writable));
         });
     }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/de/dennisguse/opentracks/settings/ImportExportSettingsFragment.java
```diff
@@ -66,7 +66,9 @@ private void setExportDirectorySummary() {
         instantExportDirectoryPreference.setSummaryProvider(preference -> {
             DocumentFile directory = PreferencesUtils.getDefaultExportDirectoryUri(getContext());
             //Use same value for not set as Androidx ListPreference and EditTextPreference
-            return directory != null ? directory.getName() : getString(R.string.not_set);
+            return directory != null
+                    ? directory.getUri().toString() + (directory.canWrite() ? "" : getString(R.string.export_dir_not_writable))
```
dennisguse, line 70: Somehow too many ifs in one line :)

dennisguse (approved):
Looks good.

By the way: is there any chance that we can create a test case for this?

src/main/java/de/dennisguse/opentracks/settings/ImportExportSettingsFragment.java
pstorch: You are right. I'll rework it.

pstorch:
I've untangled the nested if conditions.

Regarding tests: I looked into it, but didn't see an easy way to do it. One would need to create an URI in the preferences with different permissions and which later can be converted to a `DocumentFile` with the `DocumentFile.fromTreeUri` and enqueuing to the `ExportService` in the background. 

I also saw that the `ExportService` needs some renovation because the `JobIntentService` is deprecated. I can upgrade this in a next PR.

pstorch:
> I also saw that the `ExportService` needs some renovation because the `JobIntentService` is deprecated. I can upgrade this in a next PR.

Hmmm, that was a bold statement of mine. It's not that easy and as I look into it I'm not sure if this is the right approach. The Jobs in Android are for background work, but we have an interaction with the UI here (show progress and errors). So at the moment I think it should be a foreground service instead. But that's a completely different discussion.

rgmf:
> I also saw that the `ExportService` needs some renovation because the `JobIntentService` is deprecated. I can upgrade this in a next PR.

:dizzy_face: 

It sounds like a big change to do.

PR #1081, opened 2022-01-06
Bugfix: use monotonic clock.

### LINKED ISSUE #800: Crash while recording: clock went backwards?
Reported via Google PlayStore.

I guess, we have a case in which the clock was adjusted while recording.
And the TrackStatistics then react with a RuntimeException.

Just happened once so far since we implemented using the _system clock_ instead of the _gps time_.

```
java.lang.RuntimeException: 
  at de.dennisguse.opentracks.stats.TrackStatistics.setStopTime (TrackStatistics.java:159)
  at de.dennisguse.opentracks.stats.TrackStatisticsUpdater.internalAddTrackPoint (TrackStatisticsUpdater.java:123)
  at de.dennisguse.opentracks.stats.TrackStatisticsUpdater.addTrackPoint (TrackStatisticsUpdater.java:108)
  at de.dennisguse.opentracks.services.TrackRecordingService.insertTrackPoint (TrackRecordingService.java:652)
  at de.dennisguse.opentracks.services.TrackRecordingService.newTrackPoint (TrackRecordingService.java:561)
  at de.dennisguse.opentracks.services.handlers.HandlerServer.lambda$sendTrackPoint$0$HandlerServer (HandlerServer.java:59)
  at de.dennisguse.opentracks.services.handlers.-$$Lambda$HandlerServer$u7UpyDYRGPit7CRfg_glQph5Wm8.run (Unknown Source:6)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:919)
```

dennisguse:
`System.currentTimeMillis()` is not monotonic and might jump forward/backward whenever the local device time is updated.
This results in crashes while recording:
1. if the time jumps before start of a current segment
2. if the time jumps before the time of the previous trackpoint

Exceptions are nowadays created in TrackStatistics to prevent invalid data, leading to crashes.

We now use a clock that is relative to it's creation while using the time since boot of the device (guaranteed to be monotonic).

https://developer.android.com/reference/android/os/SystemClock

Fixes #800.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/services/handlers/MonotonicClock.java
+++ b/src/main/java/de/dennisguse/opentracks/services/handlers/MonotonicClock.java
@@ -1,0 +1,40 @@
+package de.dennisguse.opentracks.services.handlers;
+
+import android.os.SystemClock;
+
+import java.time.Clock;
+import java.time.Instant;
+import java.time.ZoneId;
+
+/**
+ * A monotonic clock relative to the device's clock at time of the instantiation.
+ * Uses device time since startup.
+ * Replacement for java.time.Clock as it may not be monotonic (i.e., can jump back and forwards).
+ */
+public class MonotonicClock extends Clock {
+
+    private final long epochAtCreation;
+
+    private final long elapsedRealtimeAtCreation;
+
+    public MonotonicClock() {
+        epochAtCreation = Instant.now().toEpochMilli();
+        elapsedRealtimeAtCreation = SystemClock.elapsedRealtime();
+    }
+
+    @Override
+    public Instant instant() {
+        long current = (SystemClock.elapsedRealtime() - elapsedRealtimeAtCreation);
+        return Instant.ofEpochMilli(epochAtCreation + current);
+    }
+
+    @Override
+    public ZoneId getZone() {
+        throw new RuntimeException("Not implemented");
+    }
+
+    @Override
+    public Clock withZone(ZoneId zone) {
+        throw new RuntimeException("Not implemented");
+    }
+}
--- a/src/main/java/de/dennisguse/opentracks/services/handlers/TrackPointCreator.java
+++ b/src/main/java/de/dennisguse/opentracks/services/handlers/TrackPointCreator.java
@@ -30,7 +30,7 @@ public class TrackPointCreator implements BluetoothRemoteSensorManager.SensorDat
     private final Callback service;
 
     @NonNull
-    private Clock clock = Clock.systemUTC();
+    private Clock clock = new MonotonicClock();
 
     private final GPSHandler gpsHandler;
     private BluetoothRemoteSensorManager remoteSensorManager;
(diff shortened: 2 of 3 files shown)


src/main/java/de/dennisguse/opentracks/services/handlers/MonotonicClock.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * A monotonic clock relative to the device's clock at time of the instantiation.
+ * Uses device time since startup.
+ * Replacement for java.time.Clock as may not be monotonic.
+ */
+public class MonotonicClock extends Clock {
+
+    private final long startEpoch;
+
+    private final long elapsedRelatimeAtInstation;
```
rgmf, line 18: Did you mean `elapsedRealtimeAtInstantation`?

rgmf (changes requested):

src/main/java/de/dennisguse/opentracks/services/handlers/MonotonicClock.java
dennisguse: Yes ;)

dennisguse:
@rgmf Any idea why `StringUtilsTest.testFormatDateTime()` fails?
It also fails for me locally on `main`. And on CI.
Weird as it was working until now.

rgmf:
> @rgmf Any idea why `StringUtilsTest.testFormatDateTime()` fails? It also fails for me locally on `main`. And on CI. Weird as it was working until now.

I'll look into it tomorrow ;)

dennisguse:
It started failing here: https://github.com/OpenTracksApp/OpenTracks/actions/runs/1649899045
Previous test executions in the CI were okayish (except the race condition that should be fixed by event-driven BLE data handling).

pstorch:
I'm thinking if it would be better to ignore `time.isNegative()` and set the time difference to 0 in this case.
Don't throw an Exception, just log a warning. 
Would we ignore some severe error conditions?

dennisguse:
@pstorch to be honest: I guess it wouldn't matter.
And now comes the argument why this still makes sense :)
And the argument is almost academic. :sunglasses:

The point here is that in real life time just moves in dimension anyhow.
Just because my clock becomes out of sync (with whatever absolute time source we trust) and we have tech to synchronize to an absolute (more up to date) value, we get this jumping of absolute time.

But during the recording, I don't need an absolute time, but only the relative duration to the recording start (aka how many seconds ago did I press start). There is no need to get an accurate (in terms of re-synchronized jumping) absolute time. We just store it, because most data formats want it (and it would be too some degree painful). Technically, it just like moving time and total time. 
From a data perspective, only the timestamp of SEGMENT_START_MANUAL must be absolute while all others would be sufficient as relative time (reference to the previous SEGMENT_START_MANUAL). 
And this SEGMENT_START_MANUAL should be in sync with the real world (or whichever clock the user wants to have on his phone).

The ugly part: the amount of time, we would "ignore" using `System.currentTimeMillis()` is very likely marginal (likely milliseconds at most).
But why take the chance especially as it makes the code ugly at some places?
... especially if we have a local (relative) time source available. 

PS/ general refactoring plan: all recording related variables in TrackRecordingService should be instantiated on recording start. And then the monotonic clock has actually the real start time.

dennisguse:
I start testing this :)

PR #1083, opened 2022-01-07
StringUtilsTest.testFormatDateTime

rgmf:
Trying to fix strange fails in this test.

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/util/StringUtilsTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/util/StringUtilsTest.java
@@ -20,8 +20,6 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
 import android.content.Context;
-import android.content.res.Configuration;
-import android.content.res.Resources;
 
 import androidx.test.core.app.ApplicationProvider;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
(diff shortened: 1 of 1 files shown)


rgmf:

We are doing something like this:

```java
"Jan 2, 2022 10:30:15 AM".contains("Jan 2, 2022 10:30:15")
```

So we are ignoring AM/PM/a.m./p.m. at the end of the string. I think it's not that important, at least we are sure the datetime is a "good" datetime.

dennisguse (approved):
Looks good :)

dennisguse:
Well oehm my LineageOS still disagree: 
```
java.lang.AssertionError: Jan 2, 2022 10:15:30 vs Jan 2, 2022, 10:15:30 AM
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
```

using the following assert:
```
        assertTrue(expectedFormat + " vs " + formatDate, formatDate.contains(expectedFormat));
```


I am somehow start to believe, we should just remove this test...

rgmf:
> I am somehow start to believe, we should just remove this test...

Yes, it's too much unpredictable and it shoudn't.

We are using official Java library but formatter results in different results depending on the device. I can't understand what's happening :disappointed:

dennisguse:
Sometimes it is just confusing and we have to deal with it :)

PR #1084, opened 2022-01-07
TrackRecordingService: add tests for state changes.

dennisguse:
Adding some test coverage and cleaning some tests for TrackRecordingService.

Test are integration test that use all components.
* [x] state changes (+ plus found some bugs)
* [x] actual data recording with verification of TrackStatistics (update tests in TrackRecordingServiceTestRecording and remove TrackRecordingServiceTestStatistics)
* [ ] ~replace TestDataUtil.createTrack() as it does not creates a real track anymore; use TrackRecordingService instead~


dennisguse:
Found some additional bugs:
* It could be that stopping GPS via UI doesn't work (if no recording is running)
* TrackRecordingManager handles GPS vs. sensor logic not nicely (i.e., store GPS locations depending on the distance to the previous stored GPS location and not the sensor data); don't now yet how to fix this :/

PR #1099, opened 2022-01-18
Show on lockscreen: hide controller DOESN'T WORK.

dennisguse:
Any ideas why this doesn't work?

Locking the device works, but unlocking is not triggering anything and thus the UI is not updated.


dennisguse:
Doesn't work, so ...

Issue #1095, opened 2022-01-15
No file manager found


Niekto350:
**Describe the bug**
I've changed the file manager on my phone to simple file manager. I cannot do any file related actions like import/export because the app says no file manager is found (some actions fails silently).

The same issue exist on the offline dashboard companion app (https://github.com/OpenTracksApp/OSMDashboard). This one shows the notification about missing file manager. So basically I cannot specify a location for the map files and have a blank screen with my track recording.

This is the file manager I'm using:
https://play.google.com/store/apps/details?id=com.simplemobiletools.filemanager.pro&hl=en_US&gl=US
https://f-droid.org/en/packages/com.simplemobiletools.filemanager.pro/

**To Reproduce**
1. Uninstall default phone file manager / switch to simple file manager.
2. Click on any file manager related action in the app (for example export).

**Technical information**
Using android 10 and latest OpenTracks and Simple File manager.

pstorch:
Unfortunately it looks like this FileManager doesn't provide DOCUMENTS_PROVIDER functionality which is needed. 
OpenTracks and OSMDashboard need a DocumentProvider to access files.

dennisguse:
Can we do something about the silent failures?

pstorch:
Yes, we could catch the ActivityNotFoundException as in OSMDashboard and show a message:

https://github.com/OpenTracksApp/OSMDashboard/blob/34ba6b91d0aaffe0c12d4653fa38655d5f388227/src/main/java/de/storchp/opentracks/osmplugin/BaseActivity.java#L235

Niekto350:
Thank you, I've opened a ticket here https://github.com/SimpleMobileTools/Simple-File-Manager/issues/550

Do you need the file manager to do some more complex stuff like file transfer or it's used just to generate the file path? If all you need is a file path string, could we do something like text input to type the path manually?

pstorch:
@Niekto350 it's more than just the file path. Android has this Storage Access Framework, which allows Apps without general file permission to access files outside of it's private space via a DocumenProvider App like a FileManager (can also be a remote cloud storage app).

Niekto350:
Thanks for the explanation.

pstorch:
> Can we do something about the silent failures?

I'll implement the message.

pstorch:
@Niekto350 if you like you can test our nightly build. Would be good to know if the error message appears that no compatible file manager is found

Niekto350:
[...]

Issue #1080, opened 2022-01-05
still some issues with Export to NextCloud


mega-stoffel:
**Describe the bug**
As mentioned already in #1025, my export to a NextCloud shared directory is sometimes not working. Most of the tracks are exporting perfectly fine after I record them. But unfortunately sometimes not. I haven't found a pattern, yet.
I tried to get a logcat output and I guess, this might be the error message. It's maybe not even related to OpenTracks, but some misbehavior of NextCloud?

```
[01-05 21:59:45.684 15568:15621 E/ExportUtils]Unable to open exportDocumentFile content://org.nextcloud.documents/tree/1405706866%2F12103/document/1405706866%2F16011java.io.FileNotFoundException: Failed to open document for writing 54f6640c_2022-01-04 08_34.gpx
	at android.database.DatabaseUtils.readExceptionWithFileNotFoundExceptionFromParcel(DatabaseUtils.java:149)
	at android.content.ContentProviderProxy.openAssetFile(ContentProviderNative.java:631)
	at android.content.ContentResolver.openAssetFileDescriptor(ContentResolver.java:1523)
	at android.content.ContentResolver.openOutputStream(ContentResolver.java:1229)
	at android.content.ContentResolver.openOutputStream(ContentResolver.java:1205)
	at de.dennisguse.opentracks.util.ExportUtils.exportTrack(ExportUtils.java:49)
	at de.dennisguse.opentracks.io.file.exporter.ExportService.onHandleWork(ExportService.java:62)
	at androidx.core.app.JobIntentService$CommandProcessor.doInBackground(JobIntentService.java:396)
	at androidx.core.app.JobIntentService$CommandProcessor.doInBackground(JobIntentService.java:387)
	at android.os.AsyncTask$3.call(AsyncTask.java:378)	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.lang.Thread.run(Thread.java:919)
```
**Technical information**
 - Device: Fairphone FP3
 - OS: Android 10
 - OpenTracks version: latest from FDroid, 3.23.0
 - OpenTracks commit id (only for nightly builds): [e.g. 4b333bb]

pstorch:
Thanks for the logcat. I'm still analyzing this.
[...]

PR #1100, opened 2022-01-19
Change handling of Uri and DocumentFile for #1080 and #1095

pstorch:
The Uris from Intent.getData() and DocumentFile.getUri() seems to be different. I think PreferencesUtils should only work with the original Uri as well as the permission handling logic.

IntentUtils.toDocumentFile can be used to convert it to a DocumentFile later on to actually work with.

Adds "No file manager found" message if no activity with OPEN_DOCUMENT_TREE is found. #1095

Removed deprecation of `startActivityForResult`.

I'm not sure if `IntentUtils.toDocumentFile(getContext(), directoryUri);` is at the right place. Maybe `FileUtils` is a better place.

This PR improves the situation a bit, but still doesn't solve the lost write permission on the next cloud folders. I'm still investigating.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -639,4 +639,5 @@ limitations under the License.
     <string name="field_set_primary">Primary</string>
     <string name="field_set_secondary">Secondary</string>
     <string name="field_remove_from_layout">Remove</string>
+    <string name="no_compatible_file_manager_installed">No compatible file manager installed</string>
 </resources>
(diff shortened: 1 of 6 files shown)


src/main/res/values/strings.xml
```diff
@@ -639,4 +639,5 @@ limitations under the License.
     <string name="field_set_primary">Primary</string>
     <string name="field_set_secondary">Secondary</string>
     <string name="field_remove_from_layout">Remove</string>
+    <string name="no_file_manager_found">No file manager found</string>
```
dennisguse, line 642: No compatible file manager installed?

dennisguse (approved):
Just from scrolling over it: looks good :)

dennisguse:
@pstorch regarding into which class to put it: the package restructuring is not yet finished.
We just have too many files to move around at once. 
So, just put the stuff were it makes sense to you :)

src/main/res/values/strings.xml
pstorch: Yes, that wording is better, changed it.

PR #1113, opened 2022-01-28
Refactoring TrackRecordingService for #1052

dennisguse:
... not yet sure how the final result will look like :sunglasses:


dennisguse:
@pstorch Actually, this PR includes the  "stop service after triggering export"-change.
Might make a difference or not regarding #1080.

Issue #482, opened 2020-10-30
While recording handle GPS disabled


dennisguse:
While recording a user might disable the GPS in Android.
What should we do then?

At the moment, we just continue recording, but we don't get any GPS coordinates delivered.

rgmf:
I like how OT handles this situation: it shows a notification on the notification bar and stop moving time (total time go on).

But it's also truth that user can overlook the notification bar and don't realize that GPS was disabled. I'd add an SnackBar or something like this.

dennisguse:
@rgmf Would you have the time to build some GPS feedback for the StatisticsRecordingFragment?
Some of the next steps is to make GPS optional and then it should be visible to user if it is on or not...
Technically, it is not an error anymore :)
(although we still request the permission  #1061)

rgmf:
@dennisguse okay, I'll give it a spin, so I'll see the best way to make GPS status visible to the user.

dennisguse:
@rgmf This was fixed right?

rgmf:
@dennisguse yes, it is!!! :tada:

PR #1114, opened 2022-01-29
Track recording without GPS: warn user with a Snackbar.

rgmf:
**Describe the pull request**
OpenTracks warns user if s-he starts recording with GPS off.

The Snackbar can be dismissed and it won't show up anymore. Anyway, It would be great to add a setting to disable this warning because some users can consider it a little invasive :thinking: 

Also, I've changed the theme for Snackbar. It took me a long time so I've decided to push this code with this theme but I'm not sure about it. I think it would be better let the default theme for the Snackbar (background color dark gray and foreground orange).

Above all, I create this draft PR to discuss about all these stuffs.

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/482

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
@@ -2,7 +2,6 @@ package de.dennisguse.opentracks;
 
 import android.app.AlertDialog;
 import android.content.Intent;
-import android.content.SharedPreferences;
 import android.content.SharedPreferences.OnSharedPreferenceChangeListener;
 import android.os.Build;
 import android.os.Bundle;
@@ -17,6 +16,7 @@ import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
 import androidx.viewpager2.adapter.FragmentStateAdapter;
 
+import com.google.android.material.snackbar.Snackbar;
 import com.google.android.material.tabs.TabLayoutMediator;
 
 import java.util.List;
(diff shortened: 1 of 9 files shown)


dennisguse:
@rgmf How do I turn the GPS off so the snackbar is shown?
I enabled/disabled location via the system settings (a) before starting a recording and (b) during an ongoing recording, but it did not show any effect :/

rgmf:
> @rgmf How do I turn the GPS off so the snackbar is shown? I enabled/disabled location via the system settings (a) before starting a recording and (b) during an ongoing recording, but it did not show any effect :/

Ops, I added a bug in the las commit... I'll fix it this afternoon ;)

Thank you for testing.

rgmf:
@dennisguse I think now is working.

dennisguse:
Will have to check it tomorrow: somehow it doesn't work and the new AndroidStudio has some issues with breakpoints :/

dennisguse:
... https://developer.android.com/studio/known-issues
Breakpoints are broken in AndroidStudio 2021.1.1.....
:boom:

src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
```diff
@@ -406,4 +405,20 @@ private void onRecordingStatusChanged(TrackRecordingService.RecordingStatus stat
...
         setScreenOnPolicy();
     }
-}
+
+    private void onGpsStatusChanged(GpsStatusValue gpsStatusValue) {
+        if (gpsStatusValue.isGpsStarted() && snackbar != null && snackbar.isShown()) {
+            snackbar.dismiss();
+            return;
+        }
+        if (viewBinding.trackRecordingCoordinatorLayout == null || gpsStatusValue != GpsStatusValue.GPS_DISABLED) {
```
dennisguse, line 414: null check not required, right?
viewbinding could be null (but shouldn't), but all it's child are nonnull :)

dennisguse (commented):
Works like charm :)

I really like the UI!

dennisguse:
@rgmf @pstorch debugging only works for me, if I downgrade to Android Gradle plugin 7.0.4

src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
rgmf: Yes, you are right and so is AndroidStudio :)

rgmf:
> @rgmf @pstorch debugging only works for me, if I downgrade to Android Gradle plugin 7.0.4

I can do it with 7.1.0 Gradle version.

PR #1124, opened 2022-02-07
Bugfix: speed showed sometimes 0.0 in UI.

### LINKED ISSUE #1117: Recording: speed shows 0.0 most of the time.
**Describe the bug**
When a track recording is on you can see that speed is 0.0 most of the time but average is computed and showed correctly.

Every once in a while you can see the current speed that vanishes quickly to show 0.0 again.

I've seen this issue using GPS speed (without speed sensor). I couldn't test it with speed sensor yet.

**To Reproduce**
1. Start a recording.
2. Look at speed.

**Technical information**
 - Device: Pixel 4a
 - OS: Android 12
 - OpenTracks version: debug version (current main branch).

dennisguse:
Fixes #1117.

Should only fix the problem for using the GPS.
BLE sensors might still be affected if they were - I had no time to check it.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
+++ b/src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
@@ -40,6 +40,9 @@ class TrackRecordingManager implements SharedPreferences.OnSharedPreferenceChang
     private TrackStatisticsUpdater trackStatisticsUpdater;
 
     private TrackPoint lastTrackPoint;
+    private TrackPoint lastTrackPointUIWithSpeed;
+    private TrackPoint lastTrackPointUIWithAltitude;
+
     private TrackPoint lastStoredTrackPoint;
     private TrackPoint lastStoredTrackPointWithLocation;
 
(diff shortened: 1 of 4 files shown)


rgmf:
@dennisguse I tested it this morning driving to work and it doesn't work yet.

It's really strange because we can divide the (bad) behavior into two possibilities:
1. When I drove below a certain speed (let's say 40km/h) I saw how speed jumb between 0.0 and the current speed.
2. When I drove above a certain speed (more than 40km/h) I only saw 0.0 as speed.

I unknown the speed threshold for these two behaviors and I cannot understand such a odd behavior.

dennisguse:
I guess it is GPS precision - the faster you go the less precise the measurements and thus less measurements might be reported by the GPS as well as ignored by OpenTracks due the accuracy.

Let me think about it ;)
PS elevation should also be affected.

rgmf:
> PS elevation should also be affected.

I didn't realize.

dennisguse:
@rgmf Now it should be working ;)
I am not a big fan of the solution, but it hopefully gets the job done :D

rgmf:
@dennisguse if it works then it's enough for now. Anyway, there's room for improvements.

Installing... and I'll test it driving back to home.

rgmf:
@dennisguse it worked like a charm using GPS speed.

rgmf (approved):

Issue #1116, opened 2022-02-02
Optionally provide live track updates to other apps


arjan-s:
**Is your feature request related to a problem? Please describe.**
With the addition of the intent API (https://github.com/OpenTracksApp/OpenTracks/pull/1072 and https://github.com/OpenTracksApp/OpenTracks/issues/1103) it is now possible for Gadgetbridge and other apps to start and end recordings while the phone's display is still locked. There is currently no ability for the same apps to receive live track updates (like speed and distance) from OpenTracks.

**Describe the solution you'd like**
My preferred solution for providing live track updates is to implement the explicit intent approach. Perhaps the settings screen could even have a "target package ID" setting or something like that, so give the user choice. I'm thinking about the way [Tiny Weather Forecast Germany](https://f-droid.org/en/packages/de.kaffeemitkoffein.tinyweatherforecastgermany/) does it [here](https://codeberg.org/Starfish/TinyWeatherForecastGermany/src/branch/master/app/src/main/java/de/kaffeemitkoffein/tinyweatherforecastgermany/GadgetbridgeAPI.java#L112).

**Describe alternatives you've considered**
Alternatively, since Gadgetbridge already has access to all notifications, it could try to parse the OpenTracks ongoing notification and extract the relevant data from it. This is, however, a bad practice and prone to break in the future. Also, implementation of this would be a problem due to translations.

The existing dashboard API is able to provide live updates, but it has to be started manually for each new track. The proposed solution is for automatic live updates and hopefully doesn't require a lot of extra code.

dennisguse:
What is your use case?
Assumption: the currently recording Track
And which data of a Track do you need?
* last TrackPoints
* all TrackPoints
* overall statistics

We could re-use the DashboardAPI and allow starting it automatically returning the URIs (incl. temp. read permissions) on Intent.

arjan-s:
Sorry to keep bothering you with these questions!

My use case is that wearables supported by Gadgetbridge could display the overall statistics of a currently recording track on their watchface or in a workouts app.
Perhaps the URI-on-intent method could work. I haven't looked into the dashboard API yet though. Could it allow an app without internet permissions to periodically (or streaming) receive the overall statistics?

dennisguse:
Understood.

Internet/ no Internet permission doesn't matter.
The URIs are pointing to the OpenTracks contentprovider.
What you are interested in is the Track URI.

Relevant code:
[...]

PR #1125, opened 2022-02-10
Start Dashboard API from StartRecording API

arjan-s:
**Describe the pull request**
This PR adds the ability for StartRecording to automatically start the Dashboard API for the requesting app. The requesting app needs to include `ORIGINATING_PACKAGE` and `STATISTICS_CLASS` in the intent for this functionality to activate.
Gadgetbridge has a [PR here](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/2577) that needs this PR for its functionality.

**Link to the the issue**

**License agreement**


src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
```diff
@@ -58,6 +58,20 @@ private IntentDashboardUtils() {
...
     public static void startDashboard(Context context, boolean isRecording, Track.Id... trackIds) {
+        startDashboard(context, isRecording, null, null, trackIds);
+    }
+
+    /**
+     * Send intent to show tracks on a map (needs an another app) as resource URIs. By providing a
+     * targetPackage and targetClass an explicit intent can be sent, thus bypassing the need for the
+     * user to select an app.
+     *
+     * @param context  the context
```
dennisguse, line 69: not necessary as the variable names already tell the story.

src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
```diff
@@ -58,6 +58,20 @@ private IntentDashboardUtils() {
      * @param trackIds the track ids
      */
     public static void startDashboard(Context context, boolean isRecording, Track.Id... trackIds) {
+        startDashboard(context, isRecording, null, null, trackIds);
+    }
+
+    /**
+     * Send intent to show tracks on a map (needs an another app) as resource URIs. By providing a
```
dennisguse, line 65: One line per sentence?

src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
```diff
@@ -87,6 +101,10 @@ public static void startDashboard(Context context, boolean isRecording, Track.Id
         clipData.addItem(new ClipData.Item(uris.get(MARKERS_URI_INDEX)));
         intent.setClipData(clipData);
 
+        if ((targetPackage != null) && (targetClass != null)) {
```
dennisguse, line 104: Should we print a warning if only one was set?
And a logcat output would be nice for debugging to check if actually something was received.

src/main/java/de/dennisguse/opentracks/publicapi/StartRecording.java
```diff
@@ -1,11 +1,23 @@
...
+import de.dennisguse.opentracks.data.models.Track;
 import de.dennisguse.opentracks.services.TrackRecordingService;
+import de.dennisguse.opentracks.util.IntentDashboardUtils;
 
 public class StartRecording extends AbstractAPIActivity {
 
     protected void execute(TrackRecordingService service) {
-        service.startNewTrack();
+        Track.Id trackId = service.startNewTrack();
+        Intent intent = getIntent();
```
dennisguse, line 14: Assignment only used once.

src/main/java/de/dennisguse/opentracks/publicapi/StartRecording.java
```diff
@@ -1,11 +1,23 @@
...
+        Bundle bundle = intent.getExtras();
+        if (bundle != null) {
+            String originatingPackage = bundle.getString("ORIGINATING_PACKAGE", null);
```
dennisguse, line 17: Could we do something like this?
https://stackoverflow.com/a/6783661

rgmf:
> @pstorch @rgmf What do you think?

This empowers the user. I mean, OpenTracks get concern about user privacy but at the same time offers them the possibility of using the data OpenTracks record for other porpose like upload them to a web service through Internet and things like that by using a third app.

It's a feature I won't use for now, I'm not planing to use such a feature, altough I'm curios about how Gadgetbridge can handle OpenTracks data to show them on the watch in live.

src/main/java/de/dennisguse/opentracks/publicapi/StartRecording.java
pstorch: Hmm, `ORIGINATING_PACKAGE` can be freely set by the calling application, right?
What about an API key generated/shown in the OpenTracks settings, which needs to be copied into the calling application, which then needs to provide as an bundle string?

src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
arjan-s: Fixed

src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
arjan-s: Fixed

src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
arjan-s: Fixed

src/main/java/de/dennisguse/opentracks/publicapi/StartRecording.java
arjan-s: Fixed

src/main/java/de/dennisguse/opentracks/publicapi/StartRecording.java
arjan-s: I'm afraid in my use case using `startActivityForResult()` isn't an option. It's only available in `Activity`, while the context in Gadgetbridge that's starting this intent is a `Service`.

arjan-s:
Thanks all for your reviews and comments! I've resolved most issues with the code, leaving just one open for further discussion for now. To help with the privacy concern, I've added a second setting for specifically allowing to start the Dashboard API automatically, and put them in a sub menu to make things look cleaner.

src/main/java/de/dennisguse/opentracks/publicapi/StartRecording.java
arjan-s: BTW I like the API key suggestion, but I'm holding off on implementing it until after further reviews/comments.

src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
```diff
@@ -87,6 +98,13 @@ public static void startDashboard(Context context, boolean isRecording, Track.Id
...
+            Log.i(TAG, "Starting dashboard activity with explicit intent (package=" + targetPackage + ", class=" + targetClass + ")");
+            intent.setClassName(targetPackage, targetClass);
+        } else {
+            Log.i(TAG, "Starting dashboard activity with generic intent (package=" + targetPackage + ", class=" + targetClass + ")");
```
dennisguse, line 105: targetPackage and targetClass will not be used, so why log them?

src/main/res/values/strings.xml
```diff
@@ -444,10 +444,14 @@ limitations under the License.
     <string name="settings_stats_units_title">Preferred units</string>
 
     <!-- Settings public api -->
+    <string name="settings_api_title">Public API</string>
     <string name="settings_public_api_enabled_title">Public API</string>
     <string name="settings_public_api_enabled_summary_on">Other installed applications can start or stop recordings.</string>
     <string name="settings_public_api_enabled_summary_off">Only OpenTracks can start and stop recordings.</string>
     <string name="settings_public_api_disabled_toast">OpenTracks Public API is disabled: it can be enabled in the settings.</string>
+    <string name="settings_public_api_dashboard_enabled_title">Allow Dashboard API automatic start</string>
```
dennisguse, line 452: Automatic data transfer

src/main/res/values/strings.xml
```diff
@@ -444,10 +444,14 @@ limitations under the License.
...
+    <string name="settings_public_api_dashboard_enabled_summary_on">Other installed applications can automatically start the Dashboard API for live tracks and statistics</string>
```
dennisguse, line 453: "An application that started a recording, can also access the recorded data."

src/main/res/values/strings.xml
```diff
@@ -444,10 +444,14 @@ limitations under the License.
...
+    <string name="settings_public_api_dashboard_enabled_summary_off">The Dashboard API for sharing live tracks and statistics can only be started manually</string>
```
dennisguse, line 454: "Data is not shared with application."

dennisguse (approved):
Looks good to me.
The functionality can be disabled, so that is fine!

src/main/java/de/dennisguse/opentracks/publicapi/StartRecording.java
```diff
@@ -1,11 +1,25 @@
...
 import de.dennisguse.opentracks.services.TrackRecordingService;
+import de.dennisguse.opentracks.settings.PreferencesUtils;
+import de.dennisguse.opentracks.util.IntentDashboardUtils;
 
 public class StartRecording extends AbstractAPIActivity {
+    private static final String TAG = StartRecording.class.getSimpleName();
 
     protected void execute(TrackRecordingService service) {
-        service.startNewTrack();
+        Track.Id trackId = service.startNewTrack();
```
dennisguse, line 14: One thing: what happens if the trackId is null?
This might happen if already a recording is running and then we should probably not send the data, right?

src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
arjan-s: This is because of your previous comment above: "And a logcat output would be nice for debugging to check if actually something was received."

src/main/java/de/dennisguse/opentracks/publicapi/StartRecording.java
arjan-s: Agreed. Added an extra check for that.

src/main/res/values/strings.xml
arjan-s: Improved the text.

src/main/res/values/strings.xml
arjan-s: Improved the text.

src/main/res/values/strings.xml
arjan-s: Replaced

README.md
```diff
@@ -118,21 +118,26 @@ __Only required permission:__
...
 * **Start a recording:**  `de.dennisguse.opentracks.publicapi.StartRecording`
 
-* For testing via adb: `adb shell am start -n "package/activity"`
+For testing via adb: `adb shell am start -n "package/class"`
+
+The Start/Stop API is disabled by default to protect the user's privacy, but it can easily be enabled in the settings.
 
 ## Custom Dashboards (incl. map)
-As of v3.3.1, OpenTracks supports custom dashboards.
+As of v3.3.1, OpenTracks supports custom dashboards for displaying previously recorded and live tracks.
```
dennisguse, line 139: We should also include Gadgetbridge here then.

README_API.md
```diff
@@ -17,11 +19,16 @@ These URIs notify observers if on content changes (e.g., a new `TrackPoint` was
...
 Limitations:
 * only tracks selected by the user are exposed,
-* `Waypoints` are not exposed,
+* `Waypoint`s are not exposed,
 * access is only granted temporarily and automatically revoked (see `FLAG_GRANT_READ_URI_PERMISSION`),
-* no write access is possible, and
-* neither track recordings can be started, paused, or stopped.
+* no write access is possible
+
+The Data API can also be started automatically by leveraging Extra fields in the explicit Intent for the Start/Stop API.
```
dennisguse, line 26: DashboardAPI?

README_API.md
```diff
@@ -17,11 +19,16 @@ These URIs notify observers if on content changes (e.g., a new `TrackPoint` was
...
+An application that's starting a recording with an explicit Intent can add the `STATS_TARGET_PACKAGE` and `STATS_TARGET_CLASS` Extra fields to let OpenTracks know it wants to receive the Data API URIs.
```
dennisguse, line 27: Make _Extra fields_?

src/main/java/de/dennisguse/opentracks/util/IntentDashboardUtils.java
dennisguse: Then I would move this into a separate log beforehand, but not important.

dennisguse (approved):
Still looks great.
Just some nitpicking :)

dennisguse:
@rgmf @pstorch what do you think about the current solution (aka two preferences: one for Public API and then 2nd one for data transmission)?

dennisguse:
@arjan-s would it be possible to implement tests for the different options?

README_API.md
arjan-s: Actually, I used "Data API" because it's in the title of this file. And "Start/Stop API" because it's new and is just called "API" in README.md

arjan-s:
> @arjan-s would it be possible to implement tests for the different options?

I'm afraid I have zero experience with this... Wouldn't even know where to start.

pstorch:
> @rgmf @pstorch what do you think about the current solution (aka two preferences: one for Public API and then 2nd one for data transmission)?

For me it's fine because I don't need it (at the moment). But if I would need it, I think I want to have more control which app I want to allow to read the tracks data. At the moment I can only allow "Automatic data transfer" fall all. What about the idea with a shared secret (aka api key)?

dennisguse:
@pstorch For me, both options work.
API key sounds nicer; only drawback: more configuration effort for the user.
Alternative: we could also extend StartRecording activity to be request confirmation from the user on the first request for a destionation (and then store the info in the preferences for next time).

@arjan-s This might be helpful, but I have never used it: https://developer.android.com/training/testing/espresso/intents
The more logic we have the higher the chances for bugs and then tests make our life easier in the future.

arjan-s:
Another alternative: make the API key optional, so only the most privacy minded users have to do extra configuration.

pstorch:
If it works for you, then I'm totally fine. We don't need to make it more complicated than necessary.

dennisguse:
Then let's merge this for now :)

@arjan-s would you squash the commits into 1 or 2?

And add `Fixes #1116.` as a 2nd line to the commit message?
(then Github links the commit and the issue together and also closes the issue after merging)

arjan-s:
Done. Sorry I don't have time to learn Android app testing right now... But I'll look at it some time soon.

vanous:
This PR seems to be the correct place to add this comment. We get some questions about this new, cool inter-communication between our ~~islands~~ apps :) . I have tried to document this as best as i can, if needed feel free to send the users to our wiki: https://codeberg.org/Freeyourgadget/Gadgetbridge/wiki/OpenTracks-API . Also, feel free to fix errors in the article or let me know and i can do the editing.

To give you a quick summary of Gb current status: Fossil watch can use this in fully integrated fashion. Mibands/Amazfits can use this as an event triggered by a button (if they support button actions) but this does not push live updates to the band/watch. Debug menu (in Gb) has start/stop/status. And, the Settings in Gb has a pre-populated list of package names for OpenTracks, to make setting of the package name easy.

PR #1136, opened 2022-02-21
for #1080: maybe the different type of Contexts are the problem

pstorch:
Still not sure why the persisted URI permission gets lost, but I saw that we use different contexts to persist the URI and to get the DocumentFile from it. Let's use ApplicationContext everywhere because sometimes it's an Activity and sometimes a Service.

Also stumbled upon a RuntimeException after exporting because missing breaks in a switch statement.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/io/file/exporter/ExportServiceResultReceiver.java
+++ b/src/main/java/de/dennisguse/opentracks/io/file/exporter/ExportServiceResultReceiver.java
@@ -32,8 +32,10 @@ public class ExportServiceResultReceiver extends ResultReceiver {
         switch (resultCode) {
             case RESULT_CODE_SUCCESS:
                 receiver.onExportSuccess(trackId);
+                break;
             case RESULT_CODE_ERROR:
                 receiver.onExportError(trackId);
+                break;
             default:
                 throw new RuntimeException("Unknown resultCode.");
         }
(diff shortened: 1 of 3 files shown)


dennisguse:
If it works for you :)

PS/ I would also add NextCloud into the commit message.

pstorch:
@dennisguse I've fixed the commit message. Ok to merge?

dennisguse:
@pstorch Just got this via Play Store's crash reporting.
Should be fixed by this PR, but it still seemed to make trouble - don't know how often; just saw it recently and for the first time.

java.lang.RuntimeException: 
  at de.dennisguse.opentracks.io.file.exporter.ExportServiceResultReceiver.onReceiveResult (ExportServiceResultReceiver.java:38)
  at android.os.ResultReceiver$MyRunnable.run (ResultReceiver.java:50)
  at android.os.Handler.handleCallback (Handler.java:938)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loopOnce (Looper.java:226)
  at android.os.Looper.loop (Looper.java:313)
  at android.app.ActivityThread.main (ActivityThread.java:8582)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:563)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1133)
```

pstorch:
@dennisguse yes, this is the Exception from v3.25.0. Should we publish a bugfix release v3.25.1, soon?

dennisguse:
@pstorch Done ;)
If I am too slow (or something is urgent), feel free to craft releases yourself - just use `RELEASE.sh` :)

Issue #1139, opened 2022-02-24
CI is failing: 0848897ed028ba5e33287c537376d70d20d70f06


dennisguse:
Since 0848897ed028ba5e33287c537376d70d20d70f06, the CI is reporting build failures.
Don't know exactly why, but at least it is reproducible...

PR #1148, opened 2022-03-05
for #1139: fix NPE and locale dependent tests

pstorch:


dennisguse:
Why should this be necessary? IsStarted() should already cover this case (although somehow implicitly), right?

pstorch:
Looks like a race condition. In the `onStop()` method first `remoteSensorManager` and `altitudeSumManager` are stopped and nulled, then the context is nulled (which is used in `isStarted()`). If in between a new TrackPoint is recorded, then we get the NPE.

dennisguse:
Grummel ;)

Sounds about right :/
Technically, we then need to keep the reference to the managers in a local variable as even after the null check a race condition might occur. 
We some code like this in TrackRecordingService...

pstorch:
The espresso test errors with target api 30 looks like a locale issue. I get the same errors when I change my locale to german in the emulator.

dennisguse:
That would be amazing!!! Nice idea!

pstorch:
Strange, the `EspressoEditTrackRecordingTest` is now working locally with a "german" emulator, but still not happy on CI pipeline.

pstorch:
Interesting the tests on "pull_request_target" are running again when I push new commits, but use the commit I initially pushed when opening the pull request. So they fail over and over again.

pstorch:
At least the test 23 push is now green.
The test 30 still has problems opening the overflow menu and find the edit item.
And a problem with at `TrackRecordingServiceTestStateMachine.invalidStateChanges(TrackRecordingServiceTestStateMachine.java:223)`. Not sure what happens here. I think sometimes this is not happening. So I guess here is also some kind of timing issue.

After spending some time here I remember why I don't like UI tests :see_no_evil:

src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTestStateMachine.java
```diff
@@ -101,13 +101,14 @@ public void tearDown() throws TimeoutException {
...
     @Test
-    public void initialState() {
+    public void initialState() throws InterruptedException {
         // given
         List<Track> tracks = contentProviderUtils.getTracks();
         assertTrue(tracks.isEmpty());
 
         // when
         // noop
+        Thread.sleep(1000);
```
dennisguse, line 111: This should not be necessary; the other Thread.sleep()s are necessary as we check the observables and those are executed in a different thread (I presume).
isRecording() uses the value object directly, which is set in the same thread.

src/androidTest/java/de/dennisguse/opentracks/LocaleRule.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Locale;
+
+public class LocaleRule implements TestRule {
+
+    private final Locale[] mLocales;
+    private Locale mDeviceLocale;
+
+
+    public LocaleRule(Locale... locales) {
+        mLocales = locales;
```
dennisguse, line 22: Assert that the array is not null and length > 0?

src/androidTest/java/de/dennisguse/opentracks/ui/aggregatedStatistics/EspressoAggregatedFilterTest.java
```diff
@@ -73,12 +76,33 @@ public void espressoAggregatedFilterTest() {
...
+        checkBox.perform(waitFor(2000));
 
         // check there's an edit text for "from date"
         onView(withId(R.id.filter_date_edit_text_from)).check(matches(isDisplayed()));
 
         // check there's an edit text for "to date"
         onView(withId(R.id.filter_date_edit_text_to)).check(matches(isDisplayed()));
     }
+
+    private static ViewAction waitFor(final long duration_ms) {
```
dennisguse, line 88: Should we generalize this?
We are using it somewhere else already, right?

dennisguse (approved):
Nice work :)
And UI based integration tests (especially on Android) are the worst!
@pstorch :1st_place_medal:

src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTestStateMachine.java
pstorch: This was an attempt to fix the assert with NOT_RECORDING 3 lines down. I'll remove it.

src/androidTest/java/de/dennisguse/opentracks/LocaleRule.java
pstorch: Ok, then the check further down can be removed

pstorch:
Tests are failing again 😞

dennisguse:
Should we merge this?
I mean it improves the situation...

pstorch:
Yes, it improves the situation a tiny bit. But some tests are still flaky. We can merge it as a first step and continue.

dennisguse:
@rgmf @pstorch as this is now merged, I would create a new release on the 15th?

PR #1153, opened 2022-03-10
Add information about Wahoo Tickr X

brianjmurrell:
**Describe the pull request**
Cadence is a bit of an unknown.  OpenTracks sees the Tickr X as being able to provide
cadence but doesn't display anything for it while walking on the treadmill, for example.

The Wahoo Fitness app does show cadence from the Tickr X while walking on the
treadmill.

**License agreement**

### DIFF
--- a/README_TESTED_SENSORS.md
+++ b/README_TESTED_SENSORS.md
@@ -59,6 +59,7 @@ We do not support type of movement (i.e., walking vs running).
 
 * Polar Stride
 * Stryd (incl. power)
+* Wahoo Tickr X (Cadence appears to be reported in SPM rather than RPM)
 
 ## Tested Barometric Sensor Smartphones
 


dennisguse:
The Tickr provides a running sensor, right?
If so, it should it be in the same section as the Stryd.

And can you check if the sensor also provides distance?
Just start a recording outdoors while GPS is disabled.
And this may also work on a treadmill.

About cadence: it should be provided in the same message as distance and speed...

But it may also be treated as a cycling cadence.
Can you select the sensor there?
Just saw in there documentation that they support indoor spinning.

brianjmurrell:
Not yet I'm afraid.  Just haven't had time to massage the PR yet unfortunately.

brianjmurrell:
Updated.

PR #1157, opened 2022-03-16
Android 13 preparation

dennisguse:
New things in Android 13: https://developer.android.com/about/versions/13/features

- [x] Themed icon


dennisguse:
image

PR #1166, opened 2022-04-03
Adhere to recording distance interval for TrackPoint storage.

### LINKED ISSUE #1059: Optimize idle/non-idle detection
We detect idle/non-idle changes only using the GPS/ sensor-provided speed.
As the speed is inherently noisy (especially for GPS), this leads to creating to much TrackPoints due to idle/non-idle changes.

Some ideas:
* filter the speed (using the speedbuffer from TrackStatistics)
* require a certain time before switching between idle/non-idle
* require a certain amount of GPS locations switching between idle/non-idle

Background infos: 
We removed idle/non-idle TrackPoints, but needed to re-implement them as otherwise the moving time cannot be computed.
Reverting the changes was done here: https://github.com/OpenTracksApp/OpenTracks/pull/1047

Version without idle/non-idle detection: v3.22.0

dennisguse:
Fixes #1059.

A valid LastTrackPoint (with distance) is not stored if it did not met the recording distance requirement and no change idle/non-idle took place.

@vlmendz @niekto350 could you test this?

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
@@ -172,7 +172,7 @@ public class ExportImportTest {
         track = contentProviderUtils.getTrack(trackId);
         trackPoints = TestDataUtil.getTrackPoints(contentProviderUtils, trackId);
         markers = contentProviderUtils.getMarkers(trackId);
-        assertEquals(11, trackPoints.size());
+        assertEquals(10, trackPoints.size());
         assertEquals(2, markers.size());
     }
 
--- a/src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTestRecording.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/services/TrackRecordingServiceTestRecording.java
@@ -612,6 +612,55 @@ public class TrackRecordingServiceTestRecording {
         ), TestDataUtil.getTrackPoints(contentProviderUtils, trackId));
     }
 
+    @MediumTest
+    @Test
+    public void testRecording_gpsOnly_recordingDistance_movement_non_idle() {
+        // given
+        String startTime = "2020-02-02T02:02:02Z";
+        TrackPointCreator trackPointCreator = service.getTrackPointCreator();
+        trackPointCreator.setClock(startTime);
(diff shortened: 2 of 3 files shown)


vlmendz:
Results of the recording test:

1) Recording dual locations when in motion was corrected.

2) At stops, the recording of unnecessary points seems to have improved, but not a few continue to be recorded.

My recommendations:

1) Merge this PR.

2) Try a recording logic similar to the one I posted in #1059.

dennisguse:
@pstorch @rgmf does the code makes sense to you?
Changes are again in a critical component...

rgmf:
> @pstorch @rgmf does the code makes sense to you? Changes are again in a critical component...

The code is too complex, there are a lot of "lastTrackPointWhatever", a lot cases... I guess it's complicated but it's what it is.

About the new changes, I cannot figure out a better option to ignore "lastTrackPoint" to avoid that undesirable lastTrackPoint, so good for me.

Anyway, I'm going to install this branch on my phone and test it. As it's critical I'd wait to merge it until we all test it in real situation and we'll be sure everything is working well.

PS/ What an awesome work you and @vlmendz are doing!!!

rgmf:
First test this morning. It looks well. The same segment, same zoom, a track recorded before this PR (before) and a track recorded with this PR (now):

image

dennisguse:
Looks pretty good!
Thanks for testing.

PS we have test coverage now (at least some) and thus refactoring becomes safer 😇

vlmendz:
For a better understanding of what actually happens during stops:

The following image shows a fragment of a track corresponding to a stop at a traffic light, with a duration of 1m8s.

Parada 1 08

In red are shown the segments with less than 10m in length.

Filter values used for the recording:

recordingDistanceInterval = 10 m
Idle threshold = 0.694 m/s (2.5 km/h)

It can be seen that during that minute 8 points were recorded because the idle/non-idle state changed 4 times (see the speed values at each point).

The problem with this is that it increases the distance and time in motion with each stop.

In the same image it can be inferred that by using the recordingDistanceInterval value to control the end of the idle state, unnecessary idle/non-idle state changes could have been avoided and only the 2 really necessary points would be recorded (the first and the last one with speed lower than the idle threshold).

 This is what is intended by the recording logic I put in #1059. In the following image the green line represents the 2 points that should have been recorded using this algorithm:

Parada 1 08 corregida 2

Issue #1151, opened 2022-03-09
Stopped in the middle of a workout


brianjmurrell:
**Describe the bug**
First time using OpenTracks (so no idea of frequency or reproducibiliy although statistically I guess it's 100% of the time so far :smile:) I was walking on a treadmill with a Tickr X connected to OpenTracks.  Everything was going really well until OpenTracks simply suddenly stopped recording -- as if I'd hit (and held, as one is required to do -- which simply didn't happen as the phone was sitting on the console of a treadmill at the time) the stop button.  I had to quickly press start again to create a new activity while I was on the Treadmill.  Unfortunately now I have two activities for what was actually just a single workout.

**To Reproduce**
Basic OpenTracks BLE sensor pairing and starting an activity.

**Technical information**
 - Device: OnePlus 6
 - OS: LineageOS 18.1 - March 1, 2022
 - OpenTracks version: v3.25.1; Version code: 4755

I'm really hoping this was a once-in-a-lifetime anomaly, but as a software engineer I know that is not likely.  :-(

I wish I had logcat output, but my phone was not logcatting anywhere at the time.  It typically isn't.  And I had 80% more of my workout to complete so I couldn't stop and try to grab what was in the buffer when it happened.

brianjmurrell:
As an aside, is this what "Resume track" in a track's menu is for?  Would that have at least added the second activity to the first and kept it all in one activity?

FWIW, I find this issue quite disturbing as I only happened to notice it happened because my phone was sitting right in front of me on a treadmill console.  Had I been outdoors with my phone in my pocket, I would have never noticed and would have lost 80% of my activity data.

dennisguse:
Was you screen on the whole time?
And did you have a GPS signal or just the sensor?
Sounds like a race condition in the UI code (some kind of null pointer?).
So, running with display off should be fine (recent personal experience) :)

PS/ resume track allows to continue the recording if it was stopped or OpenTracks crashed.

@brianjmurrell Can you add the Tickr X to https://github.com/OpenTracksApp/OpenTracks/blob/main/README_TESTED_SENSORS.md?

brianjmurrell:
> Was you screen on the whole time?

Indeed.  I did have the option to keep the screen on enabled.

> And did you have a GPS signal or just the sensor?

Just the sensor as it was indoors on the treadmill.

> Sounds like a race condition in the UI code (some kind of null pointer?).

What's the best way to hunt that sort of thing down?

[...]

PR #1167, opened 2022-04-03
Bugfix: race condition when using multiple sensors

dennisguse:
See https://github.com/OpenTracksApp/OpenTracks/issues/1151#issuecomment-1086693660

Solution: properly synchronize the two methods.

TODO:
- [ ] actually trying it out
- [x] remove TODO commit before merging (catches exception and logs); revert back to crash if time moves backward
- [x] fix tests

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/services/handlers/TrackPointCreator.java
+++ b/src/main/java/de/dennisguse/opentracks/services/handlers/TrackPointCreator.java
@@ -1,6 +1,7 @@
 package de.dennisguse.opentracks.services.handlers;
 
 import android.content.Context;
+import android.location.Location;
 import android.util.Log;
 import android.util.Pair;
 
@@ -47,7 +48,7 @@ public class TrackPointCreator implements BluetoothRemoteSensorManager.SensorDat
         this.gpsHandler = gpsHandler;
     }
 
-    public void start(@NonNull Context context) {
+    public synchronized void start(@NonNull Context context) {
         this.context = context;
 
         gpsHandler.onStart(context);
@@ -71,7 +72,7 @@ public class TrackPointCreator implements BluetoothRemoteSensorManager.SensorDat
         gpsHandler.onStop();
     }
 
-    public void reset() {
+    public synchronized void reset() {
         if (remoteSensorManager == null || altitudeSumManager == null) {
             Log.d(TAG, "No recording running and no reset necessary.");
             return;
@@ -80,7 +81,7 @@ public class TrackPointCreator implements BluetoothRemoteSensorManager.SensorDat
         altitudeSumManager.reset();
     }
 
-    private SensorDataSet fill(TrackPoint trackPoint) {
+    private SensorDataSet addSensorData(TrackPoint trackPoint) {
         if (!isStarted()) {
             Log.w(TAG, "Not started, should not be called.");
             return null;
@@ -98,7 +99,7 @@ public class TrackPointCreator implements BluetoothRemoteSensorManager.SensorDat
         return sensorDataSet;
     }
 
-    public void stop() {
+    public synchronized void stop() {
         gpsHandler.onStop();
 
         if (remoteSensorManager != null) {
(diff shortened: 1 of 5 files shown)


dennisguse:
I am not sure this will solve our race condition.
What we need to make sure is that __all__ calls to `TrackPointCreator.createNow()` result in an in-interruptible call to `TrackPointCreator.onNewTrackPoint()`.
But this is not true for the `GPSHandler` as it first creates a TrackPoint (incl. `createNow()`) and then does some interruptible checks.

In short: TrackPointCreator needs to create the TrackPoints, not the GPSHandler.

dennisguse:
@brianjmurrell ready for testing.
This time the race condition is prevented using an object lock and the whole time generation+storage should never be interupted.


Don't know why this happens and would love to the some documentation. I mean, we have a race condition without threading ;)

dennisguse:
@rgmf Code did not change that much; only thing is that we now the whole adding to TrackPointCreator is synchronized and should therefore enforce an order (and not be interrupted/ multi-threading shouldn't not happen inside).

I am still not sure why this happens exactly:
a) we have multiple threads (IMHO we don't)
b) we have one thread (i.e., managed by one handler), but it might be interrupted to handle another event while it is processing an event.

PS/ Seemingly this is quite a common problem on event-driven architectures.
Here is interesting paper - they did an automated analysis on Google MyTracks and some of the issues sound (still) very familiar. 
We have in the TrackRecordingService some workarounds for this in place (i.e., due to the timer that sends data to the UI).
 https://web.eecs.umich.edu/~nsatish/papers/PLDI-14-AndroidRace.pdf

dennisguse:
@brianjmurrell Can you give this a test drive?
I don't have any sensors at hand....

dennisguse:
Just checked the code and it wasn't working.
Change was more complicated than expected; tests work now locally again.
@brianjmurrell now you could actually test it.

rgmf:
> @rgmf Code did not change that much; only thing is that we now the whole adding to TrackPointCreator is synchronized and should therefore enforce an order (and not be interrupted/ multi-threading shouldn't not happen inside).
> 
> I am still not sure why this happens exactly: a) we have multiple threads (IMHO we don't) b) we have one thread (i.e., managed by one handler), but it might be interrupted to handle another event while it is processing an event.
> 
> PS/ Seemingly this is quite a common problem on event-driven architectures. Here is interesting paper - they did an automated analysis on Google MyTracks and some of the issues sound (still) very familiar. We have in the TrackRecordingService some workarounds for this in place (i.e., due to the timer that sends data to the UI). https://web.eecs.umich.edu/~nsatish/papers/PLDI-14-AndroidRace.pdf

Thank you for the explanation. I'll read the paper ;)

dennisguse:
@rgmf @pstorch I would like to merge this and it should work.
However, I am not 100% certain that we do not run into deadlocks.
No experience there.
So, my proposal is to merge it and create a release to test it in the wild...

src/main/java/de/dennisguse/opentracks/services/handlers/TrackPointCreator.java
```diff
@@ -114,16 +115,28 @@ public void stop() {
         this.context = null;
     }
 
+    public void onChange(@NonNull Location location) {
+        synchronized (this) {
+            TrackPoint trackPoint = new TrackPoint(location, createNow());
+            addSensorData(trackPoint);
+            onNewTrackPoint(trackPoint);
```
pstorch, line 122: First `addSensorData` then `onNewTrackPoint`. In `onNewTrackPoint` line 139 the `addSensorData` is called again. Is that correct?

src/main/java/de/dennisguse/opentracks/services/handlers/TrackPointCreator.java
```diff
@@ -114,16 +115,28 @@ public void stop() {
...
+    }
+
     /**
      * Got a new TrackPoint from Bluetooth only; contains no GPS location.
      */
     @Override
-    public void onChange(SensorDataSet sensorDataSet) {
-        onNewTrackPoint(new TrackPoint(TrackPoint.Type.SENSORPOINT, createNow()));
+    public void onChange(@NonNull SensorDataSet unused) {
+        synchronized (this) {
```
pstorch, line 131: A synchronized block for the whole method? Can this be simplified by adding the synchronized keyword to the method? I think it' equivalent.

pstorch:
> @rgmf @pstorch I would like to merge this and it should work. However, I am not 100% certain that we do not run into deadlocks. No experience there. So, my proposal is to merge it and create a release to test it in the wild...

I don't know if I fully understand the change, but a short test did no harm.

rgmf:
@dennisguse I'm ready to test it ;)

src/main/java/de/dennisguse/opentracks/services/handlers/TrackPointCreator.java
dennisguse: Nice catch; I removed the call - would just be wasting time.

src/main/java/de/dennisguse/opentracks/services/handlers/TrackPointCreator.java
dennisguse: Good question; not sure.
Just checked the documentation and making the onChange()-methods would have the same effect.
Was not aware of this - thanks for the hint :D

https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html

dennisguse: Just added a commit changing this :D

Issue #683, opened 2021-03-26
Export and import GPS accuracy data


dennisguse:
We store the GPS accuracy data, but we don't export/import it.

sebastianha:
Funny, I yesterday thought the same, that it would be interesting to graph this data. I like the idea!

brianjmurrell:
I do have a patch for this.  I'll push a PR soon.

dennisguse:
@brianjmurrell for testing this, accuracy verification needs to be enabled here: https://github.com/OpenTracksApp/OpenTracks/blob/28944eff07282453fa631a5301fc9246ff33f921/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java#L210

Technically: `TrackPointAssert.noAccuracy()` can then be deleted.

PR #1182, opened 2022-04-13
Include GPS accuracy in GPX export

brianjmurrell:
**Describe the pull request**
Add a new "accuracy" tag to the opentracks extension as there was no
other extension providing this tag that could be found.

**Link to the the issue**
Fixes #949 #683

**License agreement**

None

### DIFF
--- a/doc/opentracks-schema-1.0.xsd
+++ b/doc/opentracks-schema-1.0.xsd
@@ -29,4 +29,13 @@
             </xsd:documentation>
         </xsd:annotation>
     </xsd:element>
+    <xsd:element name="accuracy"
+        type="xsd:float">
+        <xsd:annotation>
+            <xsd:documentation xml:lang="en">
+                Accuracy in meters of the current TrackPoint.
+                Only used in GPX.
+            </xsd:documentation>
+        </xsd:annotation>
+    </xsd:element>
 </xsd:schema>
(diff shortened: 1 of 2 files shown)


doc/opentracks-schema-1.0.xsd
```diff
@@ -29,4 +29,13 @@
             </xsd:documentation>
         </xsd:annotation>
     </xsd:element>
+    <xsd:element name="accuracy"
+        type="xsd:float">
+        <xsd:annotation>
+            <xsd:documentation xml:lang="en">
+                Accuracy in meters of the current TrackPoint.
```
dennisguse, line 36: Can we add a link the link and the description (i.e., 68 percentile)?

https://developer.android.com/reference/android/location/Location#getAccuracy()

dennisguse (changes requested):
Can we get the import as well :)
... and we need this for GPX/ KML - just for one would be unfair ;)
... and then we need test coverage (i.e., TrackPointAssert.noAccuracy() should be removed).

dennisguse:
And I would name the element either `horizontal_accuracy` or `accuracy_horizontal` as there is also a vertical accuracy (at some point we might want to export this too).

dennisguse:
@brianjmurrell Code is close by: https://github.com/OpenTracksApp/OpenTracks/tree/main/src/main/java/de/dennisguse/opentracks/io/file

PS/ I can really recommend using AndroidStudio if you did not yet do this.
Setup is really easy and code navigation + search is awesome :)

dennisguse:
... in KML the data is stored not with the point, but as a separate array.

And if you need help: ping me.

dennisguse:
Superseeded by #1188.

Issue #1183, opened 2022-04-14
Low Max Speed values are displayed


SanskritFritz:
The MAX SPEED value is wrong when a track is displayed in OpenTracks (it was recorded by OpenTracks, but imported gpx tracks are affected too). When I open the track in say OsmAnd, there is a correct value displayed. Example: OpenTracks displays MAX SPEED: 22.3 km/h while OsmAnd displays 35 km/h. All other values (like Distance, Average Moving Speed, or Moving Time) are correct). I can provide a gpx export if needed.
Here is a short table of the Max Speed values (in km/h) for some randomly chosen tracks:
|OpenTracks|OsmAnd|
|---------------|------------|
|24.6|28|
|22.8|34|
|21|38|
|9.4|18.8|
|25.2|33|

**Technical information**
 - Device: Samsung Galaxy S7
 - OS: LineageOS 16
 - OpenTracks version: 3.26.1 from Fdroid

dennisguse:
That is not unexpected as there are different options to compute the max speed.
At least:
* GPS estimated speed (OpenTracks still does this; there is some simple filtering)
* the traveled distance using the provided GPS coordinates

And GPS is not that precise, so is a multitude of options to filter the data before the actual computation.
Don't know how OsmAnd does the computation and the question is what the "correct" value actually is.
If you use a BLE sensor to measure speed/distance, it is way more precise.

Anyhow, there is the general plan to move to distance-based avg/max speed estimation for OpenTracks.

All computations are in `TrackStatisticsUpdater` if you want to take a look, but don't expect rocket science there :rocket:

rgmf:
More information: https://github.com/OpenTracksApp/OpenTracks/issues/1003

dennisguse:
@rgmf Somehow I had a feeling that there should be a ticket for this.
Thanks!

SanskritFritz:
OK, fair enough. I don't know much about the calculation, but I can assure you that those values displayed for max speed by OpenTracks are **clearly** wrong. I know my speed on my bike and just look at the difference in the table. OsmAnd gets it right, but I can't tell you why or how. Anyway, it is what it is, if that is the final verdict, so be it. I still choose OpenTracks for tracking my rides. Thank you for developing this very useful app.

dennisguse:
@SanskritFritz the computation is just speed = distance / time. Nothing fancy.
I can try to create a quick (untested) hack in the following days if you want to try it out

[...]

PR #1185, opened 2022-04-15
TrackStatistics: compute max speed by traveled distance

### LINKED ISSUE #1003: Proposal to obtain better GPS speed values
Using OpenTracks I have noticed some problems with the GPS speed (for example, maximum speed lower than the current one). Other people have reported related problems (e.g. #978, #990).

Modify the way OpenTracks processes GPS speed. The proposed changes are:

On TrackStatisticsUpdater:

1) Discard the use of speedBuffer_mps and use the GPS speed to determine the maximum speed (this buffer causes the reduction of the maximum speed).

2) Use a lower value for MAX_ACCELERATION (at least for positive acceleration). Tentatively it could be around 0.7 - 1g.

3) Calculate the acceleration with respect to the previous received location (preferably to the last one with a valid acceleration) and not to the last one recorded.

4) Move the acceleration filter to TrackRecordingManager.onNewTrackPoint() so that all the locations filtering is done in one place and the statistics correspond to what is actually recorded.

dennisguse:
GPS estimated speed values are very noisy.
To derive the max speed the values were put in a ring buffer and an average computed.
This in itself has several limitations incl. that the time between TrackPoints is not taken into account.
Also only those TrackPoints are considered that were stored - ignored data is not used.

This approach uses the stored TrackPoints, but computing the speed by traveled distance over time.
This applies an inherent smoothing of the data (was also present before) as some TrackPoints might be ignored due to accuracy or not enough traveled distance.

For cycling sensors, the impact is that the ring buffer based smoothing is removed, but as we compute the speed anyhow in the sensor driver from distance/time, it should have no effect.
For running sensors, their might be an impact as those sensor report their own speed estimate.

Fixes #1003.

Todo:
* [ ] actual testing in the real world
* [x] adjusting affected tests
* [ ] check if there is an impact of segments (implement a test case if there is none)


dennisguse:
@rgmf I did almost exactly the same as you did: https://github.com/OpenTracksApp/OpenTracks/tree/maxSpeed%231003

The only obvious difference is that you kept the max acceleration check in place.
And I guess this is quite a good idea!

dennisguse:
Regarding: https://github.com/OpenTracksApp/OpenTracks/issues/1183#issuecomment-1100755229

@SanskritFritz Thanks for the evaluation - and welcome to the world of magic data processing.
The two high outliers (>50kmh) seem to be a problem and we should check why this is happening.
One potential reason, I removed the filter for unexpected high speed (acceleration > 2G); probably a good idea to re-implement it.

dennisguse:
Not valid anymore; needs to be done again.

Issue #999, opened 2021-10-29
Export track to CSV


dennisguse:
Mainly for analytics purpose and debugging: it would be great to export the trackpoints into a really simple CSV format.

Import is optional.

pstorch:
Just curios: is it for your own development purposes? You know you can directly access the database via the AppInspector from AndroidStudio, while you debug the app, query the trackpoints table of with a trackid and export as CSV.

dennisguse:
Yes it is - sometimes AppInspector is just not enough:
* I have some devices with development mode disabled (for reasons)
* I don't always carry my development machine with me

For weird cases, I often use a spreadsheet program and then CSV is choice to go.
That would even work on the phone itself :sunglasses:

pstorch:
Ah yes, I already guessed that you have a special usecase for this. Just wanted to make sure you know that AppInspector DB access. Because I found out rather late 😉

dennisguse:
@pstorch +1
actually, I did not use the AppInspector that often so far :)

MrS1R1US:
It's very good idea!! It would be nice to choose the export method to csv, broken down by activity and time week / month / year

dennisguse:
@MrS1R1US That was not the intention, but we could also include something like this.

brianjmurrell:
How do you include all of the various extension items that may or may not exist from one track to another?

dennisguse:
@brianjmurrell Give me 5min and then you will see :)

Issue #639, opened 2021-03-08
KML with empty gx:coords.


dennisguse:
Follow-up to #629.

As of v3.15.0, we generate KML TrackPoints without coordinates (empty gx:coords).
We use this to store timestamps of events without a GPX fix (like recording started), but it will be extended for cases like #500 (indoor training without GPS).

Although it is complaint with the OGC KML 2.2/2.3 standard allows this, many open-source tools do not and use 0.0/0.0 as coordinates if none are present:
* https://github.com/OpenTracksApp/OpenTracks/issues/629#issuecomment-791555698
* https://github.com/OpenTracksApp/OpenTracks/issues/629#issuecomment-791944551

We need to talk to them, how to handle this.

Relevant part from OGC KML 2.3: http://docs.opengeospatial.org/is/12-007r2/12-007r2.html#625
```
Missing data values for positions on the track, can be accommodated by empty kml:coord(i.e. <kml:coord></kml:coord>) or kml:angles values. Any element except kml:when can be empty and may be used to represent missing data to balance the size of the kml:when, kml:coord, and kml:anglesarrays, since the data array sizes must be equal. Missing numerical data, for example represented by an empty kml:coord or kml:angles, can be estimated by linear interpolation between two well-specified data points nearby. Such interpolation can also be applied to missing numerical data arrays in kml:ExtendedData for a track.
```

voegelas:
gpsbabel and OsmAnd have already accepted pull requests. In OpenLayer's [KML.js](https://github.com/openlayers/openlayers/blob/main/src/ol/format/KML.js) the functions gxCoordParser and readGxTrack need to be adapted. If time permits, I will create a pull request this week.

dennisguse:
@voegelas awesome!

dennisguse:
Empty coordinates are now also support by OpenLayers 6.6.0: https://github.com/openlayers/openlayers/releases/tag/v6.6.0
Thanks @voegelas!

Great job!

PR #1188, opened 2022-04-17
Import/Export GPS accuracy ot GPX/KML

dennisguse:
Follow-up to #1182.

@brianjmurrell Does this work for you?

### DIFF
--- a/doc/opentracks-schema-1.0.xsd
+++ b/doc/opentracks-schema-1.0.xsd
@@ -29,4 +29,13 @@
             </xsd:documentation>
         </xsd:annotation>
     </xsd:element>
+    <xsd:element name="accuracy"
+        type="xsd:float">
+        <xsd:annotation>
+            <xsd:documentation xml:lang="en">
+                Accuracy in meters of the current TrackPoint.
+                Only used in GPX.
+            </xsd:documentation>
+        </xsd:annotation>
+    </xsd:element>
 </xsd:schema>
--- a/doc/opentracks-schema-1.0.xsd
+++ b/doc/opentracks-schema-1.0.xsd
@@ -29,11 +29,12 @@
             </xsd:documentation>
         </xsd:annotation>
     </xsd:element>
-    <xsd:element name="accuracy"
+    <xsd:element name="accuracy_horizontal"
         type="xsd:float">
         <xsd:annotation>
             <xsd:documentation xml:lang="en">
-                Accuracy in meters of the current TrackPoint.
+                Accuracy in meters of the current TrackPoint; 68% chance that the actual location is
+                within this radius around the measured location.
                 Only used in GPX.
             </xsd:documentation>
         </xsd:annotation>
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
@@ -206,8 +206,7 @@ public class ExportImportTest {
         assertEquals(track.getIcon(), importedTrack.getIcon());
 
         // 2. trackpoints
-        TrackPointAssert a = new TrackPointAssert()
-                .noAccuracy();
+        TrackPointAssert a = new TrackPointAssert();
         List<TrackPoint> actual = TestDataUtil.getTrackPoints(contentProviderUtils, importTrackId);
         a.assertEquals(trackPoints, actual);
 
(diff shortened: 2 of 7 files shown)


brianjmurrell:
@dennisguse Awesome!  You beat me to it.  I was interior house painting this weekend so no time for hacking.

I have merged this into my branch and will give it a go on my next outdoor walk.

Am I correct to understand from the patch that if GPS is disabled for a track, there will be no `accuracy_horizontal` with the trackpoints in the GPX file?  If so, I can also confirm that on my next treadmill activity.

dennisguse:
@brianjmurrell No worries.
If  no GPS is used, there will be no TrackPoints exported to GPX.
Here, we only include TrackPoints that were provided by GPS (i.e., have latitude/longitude) as this is mandatory for GPX (to the best of my knowledge) - so this cannot happen.
For sensor data: we aggregate it (for distance summing it up) and add it to the next TrackPoint was provided by GPS.

For KML, coordinates are optional, but most tools interpret missing coordinates as lat=0,lng=0.
Here, we deployed some fixes for open-source apps - some details can be found in #639.

If you want to export non-GPS data, you need to use KML or we need to figure out how to do this with GPX or we implement #999.
That's why I implement it, so you can could use KML :D

brianjmurrell:
But I get GPX files with plenty of points in them even when I have the GPS icon on the top bar toggled off.

Is https://github.com/pytrainer/pytrainer/blob/1ebddea436117fec35abff6d0cf060bc290e5a04/schemas/Cluetrust_gpxdata10.xsd#L46-L48 for "trackpoints" with no coordinates?

dennisguse:
@brianjmurrell Can you check what TrackPoints are exported and if they confirm if they contain coordinates or not?

And regarding Cluetrust... jupp - but I can really recommend KML (because that is already implemented) :D

brianjmurrell:
> @brianjmurrell Can you check what TrackPoints are exported and if they confirm if they contain coordinates or not?

Yes they do.  The same (for all intents and purposes) `lat` and `lon` on every one of them.

> And regarding Cluetrust... jupp - but I can really recommend KML (because that is already implemented) :D

Implemented on OpenTracks' end.  My app was written to import GPX.  :frowning:

dennisguse:
Are those coordinates valid?
I mean that should not be the case if the GPS is disabled or the phone was used indoor.
Interesting but weird - I will check this tomorrow with my phone. Hopefully, there is no ugly bug.

About GPX: which parser did you use? I guess you can also just use the KMLTrackImporter from OpenTracks (it's Apache 2.0) and replace your current implementation. No need to reimplements a parser and it should not be too complicated.

Extending GPX might not be straight forward and makes parsing it even more complex.

brianjmurrell:
> Are those coordinates valid?

They are, indeed!  Opening that GPX file with gnome-maps takes me right to where that activity was going on.

> I mean that should not be the case if the GPS is disabled or the phone was used indoor. 

Indeed.

> About GPX: which parser did you use?

I am using Python and pygpx.  I suppose there is probably a KML reader as well that I (hope) I can substitute in easily enough.

dennisguse:
@brianjmurrell before you do this: let me implement #999.
And reading CSV should be piece cake - no point in creating a parser to KML and all it's "features" (aka the custom data arrays).
The CSV export would only include the TrackPoints - no statistics; and very look like our TrackPoint table.

brianjmurrell:
> creating a parser to KML

Oh, I am fairly sure I don't have to do that.  I would be very surprised if there were not already a python library for doing it.  So hopefully it operates similarly to the python GPX parser.

brianjmurrell:
> > Are those coordinates valid?
> 
> They are, indeed! Opening that GPX file with gnome-maps takes me right to where that activity was going on.

Any thoughts on this location privacy leak?

dennisguse:
@brianjmurrell depends.
Was the location permission granted?
Was the GPS disabled on the device while recording?

OpenTracks has no functionality to disable/ignore GPS for a specific recording (yet).
And if the OS is providing this data, it will just store it.

brianjmurrell:
> Was the location permission granted?

For the app or the track?  For the app, of course.  I use it both indoors and out.

> Was the GPS disabled on the device while recording?

Yes.
 
> OpenTracks has no functionality to disable/ignore GPS for a specific recording (yet). And if the OS is providing this data, it will just store it.

Oh.  What does the location toggle at the top of the screen where you press start to start a track do then?

I guess I am back to my original suggestion of a data item in a track whether that toggle was on or off for the track if it's not actually disabling the tracking of GPS.

Ultimately reading a track I want to know if the GPS data is meaningful (i.e. I was outdoors walking) or I should disregard it as it was all in the same location as I was on a treadmill.

I suppose I could derive the latter simply by measuring GPS-calculated distance of the track (compared to sensor distance which is used in preference to GPS distance in the GPX entries as I understand it) and if it's below some threshold assume the user was on a treadmill (or jogging in place, etc. :smile:).

dennisguse:
The location toggle on TrackListActivity can be used to pre-start the sensors (e.g., that the GPS already can acquire fix) before starting the actual recording. We might need to change this, but that was the original purpose ;)
So, if you don't want GPS, you need to disable via the Android system.

There is quite some potential to overhaul part of the UI, but that is never as easy as it sounds...

And yes: if you want to figure out if something was indoors that might be an option.
You can also check how many TrackPoints actually have a location.

PS/ are you using KML or CSV now?
PPS/ CSV cannot be imported, so not useful for backups.

PR #1193, opened 2022-04-19
Add CSV as export format (TrackPoints only).

### LINKED ISSUE #999: Export track to CSV

dennisguse:
Fixes #999.

Really not fancy.

@brianjmurrell does this work for you?


dennisguse:
Added check for null data - columns will still be generated.

PR #1202, opened 2022-04-24
Intervals dropdown shows wrong unit

### LINKED ISSUE #1201: Intervals dropdown text unit is wrong
**Describe the bug**
When user selects metric units and got to Intervals tab to see intervals, then the dropdown text shows "mi" unit instead of "km" unit. It's only wrong the first item.

The following screenshot shows the issue:

imagen

**To Reproduce**
1. Select metric units on Settings.
2. Open a recorded activity.
3. Go to intervals tab.
4. Look at the value on the dropdown when you can select the split interval.

**Technical information**
 - Device: All
 - OS: All
 - OpenTracks version: current main branch

rgmf:
Fixes #1201.


dennisguse (approved):
Looks good!
PS I really dislike the spinner implementation required for material design.
It feels so dirty...

rgmf:
> Looks good! PS I really dislike the spinner implementation required for material design. It feels so dirty...

It really sucks

PR #1207, opened 2022-04-30
Configurable filename for export.

### LINKED ISSUE #512: Automatic export with default track name

dennisguse:
Used for all automatic exports after recording, exporting all, and sharing tracks.
Ready for testing.
Also changes default fileformat to: `time_name.suffix`
Configuration option is in the settings under Import/Export.

Fixes #512.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,7 +60,6 @@ captures/
 .idea/encodings.xml
 .idea/jsLibraryMappings.xml
 .idea/libraries/
-.idea/misc.xml
 .idea/modules.xml
 .idea/scopes/scope_settings.xml
 .idea/shelf/
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/TrackFilenameGeneratorTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/TrackFilenameGeneratorTest.java
@@ -2,6 +2,8 @@ package de.dennisguse.opentracks.io.file;
 
 import static org.junit.Assert.assertEquals;
 
+import androidx.test.core.app.ApplicationProvider;
+
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -60,7 +60,6 @@ captures/
 .idea/encodings.xml
 .idea/jsLibraryMappings.xml
 .idea/libraries/
-.idea/misc.xml
```
rgmf, line 63: Is it needed? Can't we ignore it?

rgmf:
@dennisguse it seems not too easy for the user to know what to do here, right?

imagen

dennisguse:
@rgmf Good point - I was assuming it would show the default value.

.gitignore
dennisguse: It was ignored twice and once is enough ;)

pstorch:
It works. 
I'm just thinking about to make it more user friendly. At the moment there is no feedback if the pattern has a mistake, it just reverts to the previous value. 
Idea: also add buttons (or make the patterns clickable) to insert them into the textfield. Typing curly braces on an android keyboard is a bit difficult.

dennisguse:
Valid argument.
Another problem is that the keys are not localized and also no explanation is given - so you need to know (and be able to type ASCII).

Beside using a textfield, I was thinking about a list preference and predefine all option. Then we get labels, but people are not able to define the separators (if they want custom ones).

 It sure what the best solution is....

In any case, we can pre-render the current filename in the dialog.

pstorch:
> In any case, we can pre-render the current filename in the dialog.

Good idea, that will help.

dennisguse:
@pstorch sadly it is not that easy :)
We need a custom dialog and then some translations etc.
I am somehow inclined to ignore this for now and put it on the "to be done later if somebody complains"-list - any hard objections?

rgmf:
@dennisguse IMHO go ahead, for now it's a good solution. We always will been able to improve ir but for now it works and fix an issue.

PR #1211, opened 2022-05-03
Vertical accuracy: process and import/export.

### LINKED ISSUE #78: TrackRecordingSevice only filters location by horizontal accuracy
For some reason, the `TrackRecordingService` only filters location by horizontal accuracy while ignoring vertical accuracy provided by `location.getVerticalAccuracyMeters()`.

It might be a good idea to account for both.
In addition, it needs to be investigated if location might not have accuracy (i.e., `location.hasAccuracy()` => false).
In this case, all locations are ignored.

NOTE: vertical accuracy refers to elevation.

dennisguse:
Fixes #78.

So far, we ignored the vertical accuracy.
And now we store + import/export it.

In #78 it was proposed to filter TrackPoints based upon vertical accuracy (like we do it with horizontal accuracy).
However, I don't see the point anymore.
If somebody wants to filter afterwards, the data is now there.

Requires API26 as vertical accuracy is not exposed before.


dennisguse:
Holy smokes: we have some kind of race condition in the ExportImport test.
The recording does not always return the same result and therefore the hardcoded CSV expectation is not working always.

Issue #1165, opened 2022-04-02
Security Exception when trying to store Blutooth sensor data


dennisguse:
```
03-31 16:39:07.591 20611 27265 D BluetoothConnectionManager: Received data from E5:F7:CA:AB:CD:EF
03-31 16:39:07.591 20611 28467 D BluetoothConnectionManager: Received data from E5:F7:CA:AB:CD:EF
03-31 16:39:07.591 20611 26888 D BluetoothConnectionManager: Received data from E5:F7:CA:AB:CD:EF
03-31 16:39:07.592 20611 26886 D BluetoothConnectionManager: Received data from E5:F7:CA:AB:CD:EF
03-31 16:39:07.593 20611 28467 I opentracks: cadence read: 124.0
03-31 16:39:07.593 20611 26888 D BluetoothConnectionManager: Decoded data from E5:F7:CA:AB:CD:EF: sensorAddress='E5:F7:CA:AB:CD:EF heart=HeartRate{value=92.0 bpm}
03-31 16:39:07.593 20611 28467 D BluetoothConnectionManager: Decoded data from E5:F7:CA:AB:CD:EF: sensorAddress='E5:F7:CA:AB:CD:EF
03-31 16:39:07.593 20611 28467 D BluetoothRemoteSensorManager: Previous: sensorAddress='E5:F7:CA:AB:CD:EF; CurrentsensorAddress='E5:F7:CA:AB:CD:EF
03-31 16:39:07.593 20611 26888 D TrackRecordingManager: Not recording TrackPoint
03-31 16:39:07.594 20611 27265 D BluetoothConnectionManager: Decoded data from E5:F7:CA:AB:CD:EF: sensorAddress='E5:F7:CA:AB:CD:EF heart=HeartRate{value=92.0 bpm}
03-31 16:39:07.595 20611 26886 I opentracks: cadence read: 124.0
03-31 16:39:07.595 20611 26886 D BluetoothConnectionManager: Decoded data from E5:F7:CA:AB:CD:EF: sensorAddress='E5:F7:CA:AB:CD:EF
03-31 16:39:07.599 20611 26886 D BluetoothRemoteSensorManager: Previous: sensorAddress='E5:F7:CA:AB:CD:EF; CurrentsensorAddress='E5:F7:CA:AB:CD:EF
03-31 16:39:07.599 20611 28467 D TrackRecordingManager: Not recording TrackPoint
[...]

PR #1214, opened 2022-05-05
Bluetooth: use TrackRecording handler for callback.

dennisguse:
Otherwise the thread is not defined and a SecurityException may be triggered when inserting data into the ContentProvider.

Part of #1165.

BUT we can only deploy this actual fix for API>26.
All others may always/often/sometimes/never be able to store BLE sensor provided data (without GPS).

```
/home/runner/work/OpenTracks/OpenTracks/src/main/java/de/dennisguse/opentracks/sensors/BluetoothConnectionManager.java:132: Error: Call requires API level 26 (current min is 21): android.bluetooth.BluetoothDevice#connectGatt [NewApi]
        bluetoothGatt = device.connectGatt(context, true, connectCallback, BluetoothDevice.TRANSPORT_AUTO, 0, this.observer.getHandler());
```

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
@@ -158,7 +158,7 @@ public class ExportImportTest {
         service.insertMarker("Marker 2", "Marker 2 category", "Marker 2 desc", null);
 
         trackPointCreator.setClock("2020-02-02T02:02:06Z");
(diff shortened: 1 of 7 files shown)


dennisguse:
@brianjmurrell I cannot reproduce the SecurityException (just happened in one recording).
So, it is hard to verify that this is not happening anymore.
Do you know how to reproduce it?

brianjmurrell:
I don't know of any way specifically.  I just came across it one day while I was looking for other things in the logcat.

dennisguse:
@brianjmurrell :)
Thanks that you spotted it - it would have been a really tricky debugging as the exceptions is not killing the app.
I will do some test drives and then merge - we will have to test this with a release.
At least it will crash if it is not working...

PR #1216, opened 2022-05-08
Implement Public API stop

dennisguse:
The stop intent is now handled properly.
This is achieved by extracting starting the recording from `TrackRecordingActivity` into the calling code.
`TrackRecordingActivity` finishes automatically if the recording is stopped. 

Beforehand `TrackRecordingActivity` was starting the recording if the `TrackRecordingService` was not recording.
However, this restarted recording if stopped via Intent if `TrackRecordingActivity` was brought into foreground.

Build on top of #1155.

TODO:
- [x] wait for merging #1155.
- [x] rebase onto main
- [x] test again ;)


dennisguse:
Works for me :)

PR #1217, opened 2022-05-08
Upgrade to KML2.3

dennisguse:
Export format is now KML2.3.
The importer supports KML2.2 and KML2.3.

Tested with:
- [x] ~OpenLayers~ (not supported as of 7.1.0): https://github.com/openlayers/openlayers/issues/14104
- [x] ~GoldenChetah~ (cannot import KML2.2 or KML2.3) https://github.com/GoldenCheetah/GoldenCheetah/issues/4271
- [x] ~OsmAnd~ (not supported; supports KML2.2); actually does not have proper KML support; they use XML transformation to GPX https://github.com/osmandapp/OsmAnd/issues/15271
- [x] ~OrganicMaps~ (not supported: supports KML 2.2)
- [x] GpxPod (only supports 2.3)
- [x] ~Google Maps~ (not supported; supports KML2.2)
- [x] https://www.gpsvisualizer.com (supported) ~(not supported; supports KML2.2): informed by email~

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,19 +64,19 @@ _OpenTracks_ is a sport tracking application that completely respects your priva
 * __Voice announcements__
 * __Photos and Markers:__ mark interesting locations while tracking
 * __Export:__
-  * export tracks either as KMZ (incl. photos), KML, or GPX
-  * export automatically after each recording (e.g., to sync via [Nextcloud](https://nextcloud.com/))
-  * avoid duplication: each exported file contain a random unique identifier (i.e., `opentracks:trackid`)
+    * export tracks either as [KMZ 2.3](https://docs.opengeospatial.org/is/12-007r2/12-007r2.html) (incl. photos), [KML 2.3](https://docs.opengeospatial.org/is/12-007r2/12-007r2.html), or [GPX 1.1](https://www.topografix.com/GPX/1/1/)
+    * export automatically after each recording (e.g., to sync via [Nextcloud](https://nextcloud.com/))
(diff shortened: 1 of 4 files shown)


pstorch:
I tried it with two web kml viewer (one of them was google maps). And they cannot read the files.
I could open a few files with a Map Viewer plugin in Visual Studio Code.
Didn't see any error logs or so. I try to analyze further.

dennisguse:
I just tested with OpenLayers and there it works...

pstorch:
Hmm, looks like the support for KML 2.3 is not that widespread.

dennisguse:
:/

Thanks for testing.
I had hoped that support would be more widespread.
I guess, we then have to do some more communication - at least to the open-source projects out there.

dennisguse:
GPXPod does not support KML2.2, but only KML2.3 https://github.com/eneiluj/gpxpod#known-issues

dennisguse:
@pstorch last weekend, I had some time check the KML2.3 support out there.
And to be honest - it was worse than expected.
I now see, why people are sticking with GPX ;)

Anyhow....
My proposal: switch to KML2.3 as export format while allowing to import KML2.2.
In addition, we keep pushing open-source projects to slowly migrate as well.
Finally, we add a "compatibility matrix" to our README showing which format to use when.

Does this sounds reasonable?

pstorch:
I don't know how many users rely on KML 2.2 as a transfer format into another app.
I use KML export only as backup for OT. So for me it's fine :wink: 

And there is also GPX which is widely supported by other apps.

PR #1223, opened 2022-05-10
Move to androidx.constraintlayout.

dennisguse:


dennisguse:
Tests work locally (emulator and device); seems to be we are back to flacky UI tests...

PR #1226, opened 2022-05-11
Public API: respect post-workout export.

### LINKED ISSUE #1225: AFTER v3.28.0: Public API stop does not auto export
@dennisguse
Looks like auto export doesn't happen on StopRecording. New ticket or do you reopen this here?

_Originally posted by @sgofferj in https://github.com/OpenTracksApp/OpenTracks/issues/1215#issuecomment-1124007123_

dennisguse:
Fixes #1225.
Don't know if this is the best implementation.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/publicapi/StopRecording.java
+++ b/src/main/java/de/dennisguse/opentracks/publicapi/StopRecording.java
@@ -1,11 +1,24 @@
 package de.dennisguse.opentracks.publicapi;
 
+import de.dennisguse.opentracks.data.models.Track;
+import de.dennisguse.opentracks.services.RecordingData;
 import de.dennisguse.opentracks.services.TrackRecordingService;
+import de.dennisguse.opentracks.util.ExportUtils;
 
 public class StopRecording extends AbstractAPIActivity {
 
     protected void execute(TrackRecordingService service) {
+        RecordingData recordingData = service.getRecordingDataObservable().getValue();
+        Track.Id trackId = null;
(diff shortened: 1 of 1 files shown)


dennisguse:
@rgmf I am not really happy with this solution as it duplicates the places that trigger an export.
What about the following:
* While being on TrackStoppedActivity, we just pause the recording (also keeps sensors hot)
* When the selects the "done" option, we trigger `service.endCurrentRecording()` 
* in `service.endCurrentRecording()`, we trigger the export

Then StopRecording only needs to call endCurrentRecording (which it is already doing) and we are done.

Does this makes sense?

rgmf:
> @rgmf I am not really happy with this solution as it duplicates the places that trigger an export. What about the following:
> 
> * While being on TrackStoppedActivity, we just pause the recording (also keeps sensors hot)
> * When the selects the "done" option, we trigger `service.endCurrentRecording()`
> * in `service.endCurrentRecording()`, we trigger the export
> 
> Then StopRecording only needs to call endCurrentRecording (which it is already doing) and we are done.
> 
> Does this makes sense?

That sounds like a good option. But, in that case we can't remove pause logic from service.

dennisguse:
Really important argument!
So, implementing this requires that keep the pause functionality, which is kind of a complex thing in TrackRecordingService.
And it is probably less complex to just duplicate the export code into TrackStoppedActivity and StopRecording.
Then we have no state handling and keep our core component "cleaner".

Then I am favor of my PR...

PS/ just did some fine-tuning: aka first end recording and then export.

rgmf (approved):

PR #1229, opened 2022-05-12
Changed second t.getLatitude() to t.getLongitude()

Pixtxa:
I thought it was EXCEL or the csv-to-xlsx-converter, until I realized, the data is also broken in newly generated csv files.

**Describe the pull request**
Should fix csv exports, so they contain latitude AND longitude instead of latitude twice.

**Link to the the issue**
I've not looked in the issues, I just put "csv latitude" into the searchbar, found two files of this repo, looked at the one which has not "Test" in it's name, searched for "latitude" there and found three hits in two lines and changed one of them into a naming, that seems legit to me. I'm not a java-Dev and also not really into this project, I don't even know how to compile/run/test it.

**License agreement**
It's ok to put my contribution under the _Apache License 2.0_ (see [LICENSE.md](LICENSE.md)).

None


dennisguse:
There is a test that will be failing.

PS/ thx :)

dennisguse:
We are using the same values for lat/lng in the tests...

Issue #240, opened 2020-05-27
2 Feature Requests: Countdown, Stop button lock


developersx3:
1) Before the tracking starts, it would be great to have a possible countdown.  This could be further enhanced by allow the countdown timer length to be chosen in settings.

2) Accidental clicking of stop. I've seen this on other running apps that require you to press an "unlock" button before stopping a recording to prevent accidental stopping of a track.  An alternative method might be requiring a long press to stop the recording.

I've switched over to OpenTracks recently and its really great.  Low footprint and very functional for my runs.

dennisguse:
Valid requests! Can you split it into two issues?
Easier to work with for us...

dennisguse:
@developersx3 Why do you want a countdown (What is your use case)?
To wait for first GPS fix?

developersx3:
On my phone which is inside my armband before a run, registering a touch is sometimes difficult.  After pressing start thru the armband, you may need to readjust the position of the band.

Screens with LCDs have bad visuals at angles. Additionally looking at a dark mode screen (black background with a red button) is harder, especially when looking at it thru a foggy armband cover, against a sun.  Countdown gives you an audio queue that you have successfully pressed the start button.

When stating a run, I prefer to build a pace when starting for closer to accurate results.

Additionally this might now only give you a better GPS fix as well as proper bearing calculation

PR #1230, opened 2022-05-12
Refactor permission requests.

dennisguse:
We now request GPS and Bluetooth on application startup aka in TrackListActivity.

Before that the GPS permission were requested in TrackRecordingActivity as this activity was actually starting the recording.
As TrackListActivity now starts the recording, the permissions would be then requested too late.
Now we request both at startup.
... but we request permissions that we do not yet need.

In addition, the Bluetooth permissions (Android 12+) were requested in the settings only (now also on startup).

To be considered: Public API does not request permissions.
Do we need this?

Implementing #240 would enable us to request permissions before actually starting the recording?

Permissions can only be requested in activities.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/TrackRecordingActivity.java
@@ -1,6 +1,5 @@
 package de.dennisguse.opentracks;
 
-import android.Manifest;
 import android.app.AlertDialog;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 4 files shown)


dennisguse:
@rgmf @pstorch does this makes sense?

@brianjmurrell can you test this?
I only have an emulator for Android 12...

rgmf:
> @rgmf @pstorch does this makes sense?

I think it's a good idea ask for all permissions on startup and on the same place whenever possible.

Issue #1232, opened 2022-05-13
v3.28.0: NPE when generating filenames (export/ share)


dennisguse:
Template for TrackFilenameGenerator might be null.

```
java.lang.RuntimeException: 
  at android.os.AsyncTask$4.done (AsyncTask.java:415)
  at java.util.concurrent.FutureTask.finishCompletion (FutureTask.java:383)
  at java.util.concurrent.FutureTask.setException (FutureTask.java:252)
  at java.util.concurrent.FutureTask.run (FutureTask.java:271)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:923)
Caused by: java.lang.NullPointerException: 
  at java.lang.StringBuilder.<init> (StringBuilder.java:113)
  at de.dennisguse.opentracks.io.file.TrackFilenameGenerator.format (TrackFilenameGenerator.java:49)
  at de.dennisguse.opentracks.io.file.TrackFilenameGenerator.format (TrackFilenameGenerator.java:45)
  at de.dennisguse.opentracks.io.file.TrackFilenameGenerator.getExample (TrackFilenameGenerator.java:97)
  at de.dennisguse.opentracks.io.file.TrackFilenameGenerator.isValid (TrackFilenameGenerator.java:84)
  at de.dennisguse.opentracks.settings.PreferencesUtils.getTrackFileformatGenerator (PreferencesUtils.java:580)
  at de.dennisguse.opentracks.util.ExportUtils.getExportDocumentFileUri (ExportUtils.java:111)
  at de.dennisguse.opentracks.util.ExportUtils.exportTrack (ExportUtils.java:63)
  at de.dennisguse.opentracks.io.file.exporter.ExportService.onHandleWork (ExportService.java:62)
  at androidx.core.app.JobIntentService$CommandProcessor.doInBackground (JobIntentService.java:396)
  at androidx.core.app.JobIntentService$CommandProcessor.doInBackground (JobIntentService.java:387)
  at android.os.AsyncTask$3.call (AsyncTask.java:394)
  at java.util.concurrent.FutureTask.run (FutureTask.java:266)
```

dennisguse:
@rgmf @pstorch Sadly, that is an ugly bug.
[...]

PR #1237, opened 2022-05-15
Add ExceptionHandler and ShowErrorActivity

pstorch:
As discussed here https://github.com/OpenTracksApp/OpenTracks/issues/1232#issuecomment-1126897115, I've added an ExceptionHandler and a ShowErrorActivity with the possibility to send us crash reports.

There is no test menu item, I've also removed it from OSMDashboard. But I guess it's easy to add an Exception for testing somewhere :wink:


src/main/java/de/dennisguse/opentracks/ShowErrorActivity.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private static final String TAG = ShowErrorActivity.class.getSimpleName();
+
+    private ActivityShowErrorBinding binding;
+
+    @Override
+    public void onCreate(final Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        binding = ActivityShowErrorBinding.inflate(getLayoutInflater());
```
dennisguse, line 30: Why not inflate in `getRootView()`?
PS/ don't what the difference actually is...

src/main/java/de/dennisguse/opentracks/ShowErrorActivity.java
```diff
@@ -0,0 +1,86 @@
...
+        Uri uriUrl;
+        try {
+            uriUrl = Uri.parse(
+                    String.format(
+                            getString(R.string.report_issue_link),
+                            URLEncoder.encode(binding.textViewError.getText().toString(), StandardCharsets.UTF_8.toString())
+                    )
+            );
+        } catch (final UnsupportedEncodingException e) {
+            Log.e(TAG, "Can't encode error report", e);
```
dennisguse, line 53: Do we need user feedback for this case?
Or is this implicit?

src/main/java/de/dennisguse/opentracks/Startup.java
```diff
@@ -37,6 +43,40 @@ public void onCreate() {
...
+
+        if (!isCrashReportingProcess) {
+            Thread.UncaughtExceptionHandler defaultPlatformHandler = Thread.getDefaultUncaughtExceptionHandler();
+            ExceptionHandler crashReporter = new ExceptionHandler(this,
+                    defaultPlatformHandler);
+            Thread.setDefaultUncaughtExceptionHandler(crashReporter);
+        }
+    }
+
+    private String getAppProcessName() {
```
dennisguse, line 62: I would avoid this; there is a good reason why it was removed as of Android P.

src/main/java/de/dennisguse/opentracks/util/ExceptionHandler.java
```diff
@@ -0,0 +1,90 @@
...
+import androidx.annotation.NonNull;
+
+import java.util.Arrays;
+
+import de.dennisguse.opentracks.BuildConfig;
+import de.dennisguse.opentracks.ShowErrorActivity;
+
+public class ExceptionHandler implements Thread.UncaughtExceptionHandler {
+
+    private static final int EXCEPTION_FORMAT_MAX_RECURSIVITY = 10;
```
dennisguse, line 16: Why 10? And not say: 3?

src/main/java/de/dennisguse/opentracks/util/ExceptionHandler.java
```diff
@@ -0,0 +1,90 @@
...
+    private String formatExceptionRecursive(final Thread thread, final Throwable exception, final int count) {
+        if (count > EXCEPTION_FORMAT_MAX_RECURSIVITY) {
+            return "Max number of recursive exception causes exceeded!";
+        }
+        // print exception
+        StringBuilder stringBuilder = new StringBuilder();
+        StackTraceElement[] stackTrace = exception.getStackTrace();
+        stringBuilder.append(String.format("Exception in thread \"%s\" %s\n", thread.getName(), exception));
+        // print available stacktrace
+        Arrays.stream(stackTrace).forEach(element -> stringBuilder.append("    at ").append(element).append("\n"));
```
dennisguse, line 57: Is there no cooler way than manually formatting it (again)?
Like
```
Writer writer = new StringWriter();
exception.printStackTrace(new PrintWriter(writer));
String s = writer.toString();
```

src/main/java/de/dennisguse/opentracks/util/ExceptionHandler.java
```diff
@@ -0,0 +1,90 @@
...
+        return stringBuilder.toString();
+    }
+
+    private String generateErrorReport(final String stackTrace) {
+        return "### Cause of error\n" +
+            "```java\n" +
+            stackTrace + "\n" +
+            "```\n" +
+            "\n" +
+            "### App information\n" +
```
dennisguse, line 73: I would show the device information first and then the stacktrace.
That allows the user to see the (potential) private information that is going to be shared.
At the end it will be easily overlooked.

dennisguse (approved):
Tiny nitpicks only.

src/main/java/de/dennisguse/opentracks/ShowErrorActivity.java
pstorch: Ah, right. OT has it's own AbstractActivity with `getRootView()`, I'll change it.

src/main/java/de/dennisguse/opentracks/ShowErrorActivity.java
pstorch: I think this can't happen as UTF-8 is available everywhere, but the Exception has to be caught. I can remove the `Log.e()` and rename the variable to `ignored`, if you like.

src/main/java/de/dennisguse/opentracks/Startup.java
pstorch: Hmm, good point. I found another workaround for older devices here: https://stackoverflow.com/a/55842542/4952072
Better or should we omit this check completely for older devices? I don't know what can happen, if an Exception is happening in the ErrorHandler itself.

src/main/java/de/dennisguse/opentracks/util/ExceptionHandler.java
pstorch: That was the value defined in the Nextcloud code. I didn't question it. Do you think 3 is enough?

src/main/java/de/dennisguse/opentracks/util/ExceptionHandler.java
pstorch: Thanks, that simplifies a lot and also takes care of the `Cause`, so no recursion necessary to handle on our own.

src/main/java/de/dennisguse/opentracks/util/ExceptionHandler.java
pstorch: Gone

pstorch:
> Tiny nitpicks only.

As always valuable feedback.

src/main/java/de/dennisguse/opentracks/Startup.java
dennisguse: Actually, what is the benefit of exposing this?
Will this give us some meaningful intel?

pstorch: What do you mean by "of exposing this"?
As I understand the ShowErrorActivity is launched with its own process name `:crash` as defined in the manifest: https://github.com/OpenTracksApp/OpenTracks/pull/1237/files#diff-6305bbf50570f0946de077f04e8ea56c53d8f6e1b53ec403c8c2ddb10d531db1R373
This is used to distinguish between the normal app run and the crash report. If the crash report is launched then the `DefaultUncaughtExceptionHandler` is not changed to prevent a loop in case there is an Exception in the `ExceptionHandler` or `ShowErrorActivity`. 
Do you understand it differently? Do I miss something?

dennisguse: Now I got ;)
Then my suggestion is to refactor getAppPackageName() to isCrashReportingProcess() and do the whole thing in there.

dennisguse (approved):
:D

src/main/java/de/dennisguse/opentracks/Startup.java
pstorch: Ah ok, the method name is not clear. Yes, that can be improved. I've moved the `processName.endsWith(":crash")` logic into the `getAppPackageName()` method and renamed it to `isCrashReportingProcess()`, because it is only used for that. Looks better now, what do you think?

dennisguse: Awesome!

dennisguse:
Now I am waiting for the nightly, so I can submit crash reports myself (and don't have to copy the stacktraces from the playstore) :partying_face:

Issue #645, opened 2021-03-13
AnnouncementPeriodicTask always loads all TrackPoints for each announcement.


dennisguse:
AnnouncementPeriodicTask always loads all TrackPoints for each announcement.
The longer a track gets, the more time, we spent on loading the data.

We should somehow store the state and just load the TrackPoints that were not yet there when the last announcement was made.

Relevant code: 
```
        List<TrackPoint> trackPoints = contentProviderUtils.getTrackPoints(track.getId());
        IntervalStatistics intervalStatistics = new IntervalStatistics(trackPoints, (float) (PreferencesUtils.isMetricUnits(context) ? 1d * UnitConversions.KM_TO_M : 1d * UnitConversions.MI_TO_M));
        IntervalStatistics.Interval lastInterval = intervalStatistics.getLastInterval();
```

rgmf:
Maybe it could be improved with #703... I take this so I'll try to fix it together #703

dennisguse:
This also includes SENSOR_STATS_QUERY as it queries all the sensor data (in avg, max) for the announced track and this will scan all data as well.

PR #1244, opened 2022-05-23
Announcement computes intervals from last track point.

rgmf:
**Describe the pull request**
VoiceAnnouncement computed all intervals again and again.

In this PR VoiceAnnouncement computes intervals from last TrackPoint.

Is it enough? I'm not sure because `ContentProviderUtils.getLastValidTrackPoint` method is deprecated. Why is that method deprecated?

**Link to the the issue**
#645 

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncement.java
+++ b/src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncement.java
@@ -34,8 +34,8 @@ import de.dennisguse.opentracks.data.ContentProviderUtils;
 import de.dennisguse.opentracks.data.TrackPointIterator;
 import de.dennisguse.opentracks.data.models.Distance;
 import de.dennisguse.opentracks.data.models.Track;
+import de.dennisguse.opentracks.data.models.TrackPoint;
 import de.dennisguse.opentracks.settings.PreferencesUtils;
-import de.dennisguse.opentracks.settings.UnitSystem;
 import de.dennisguse.opentracks.stats.SensorStatistics;
 import de.dennisguse.opentracks.stats.TrackStatistics;
 import de.dennisguse.opentracks.ui.intervals.IntervalStatistics;
@@ -56,6 +56,10 @@ public class VoiceAnnouncement {
     private final AudioManager audioManager;
 
     private final ContentProviderUtils contentProviderUtils;
+    private TrackPoint.Id startTrackPointId = null;
+
+    private IntervalStatistics intervalStatistics;
+    private Distance intervalDistance;
 
     private final AudioManager.OnAudioFocusChangeListener audioFocusChangeListener = new AudioManager.OnAudioFocusChangeListener() {
         @Override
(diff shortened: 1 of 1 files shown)


dennisguse:
getLastValidTrackPoint is deprecated as it uses a db to fetch something that is already available in memory. Day should be in TrackPointCreator ;)

dennisguse:
getLastValidTrackPoint is deprecated as it uses a db to fetch something that is already available in memory. Day should be in TrackPointCreator ;)

src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncement.java
```diff
@@ -183,6 +193,8 @@ public void announce(@NonNull Track track) {
 
         // We don't care about the utterance id. It is supplied here to force onUtteranceCompleted to be called.
         tts.speak(announcement, TextToSpeech.QUEUE_FLUSH, null, "not used");
+
+        startTrackPoint = contentProviderUtils.getLastValidTrackPoint(track.getId());
```
dennisguse, line 197: startTrackPoint is the TrackPoint from the previous announcement?

What about passing it into `update()` and then pass it to `announce()`?
TrackRecordingService should have it available.

More problems with getLastValidTrackPoint: 
* what is a valid trackpoint? (it was a location previously, right?)
* it does not take care of segments
  (might be a problem if one resumes a recording, has a time-based announcement, and no new trackpoints on first announcement)

dennisguse (changes requested):
Can we omit using getLastValidTrackPoint?

src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncement.java
rgmf: > What about passing it into `update()` and then pass it to `announce()`? TrackRecordingService should have it available.


That was my first thought, but I decided not to use it for looking for the easier approach. I back to my first though (I thought it was a worse approach :D).


> 
>     * it does not take care of segments
>       (might be a problem if one resumes a recording, has a time-based announcement, and no new trackpoints on first announcement)


IMHO interval is a segment between two points: A and B, and every interval shows the time (and stats)   takes the user to travel from A to B. So, in that approach pauses, stops... don't take account.

The more you pause/stop, the more it takes you to go from A to B.

I was thinking on that when I implemented intervals at the begining.

dennisguse:
One more thing: accessing only stored data might have some pitfalls.
For example, somebody using voice announcement 100m with min recording distance of 1000m.
Then the voice announcements would rarely be correct.

rgmf:
> One more thing: accessing only stored data might have some pitfalls. For example, somebody using voice announcement 100m with min recording distance of 1000m. Then the voice announcements would rarely be correct.

You're right. That's a good point I overlooked, thank you very much ;)

dennisguse:
Is this ready?

rgmf:
@dennisguse It's ready... By the way, I fixed a just discovered bug.

dennisguse:
@rgmf can you rebase onto main?

PR #1248, opened 2022-05-25
Allow settings reset and import/export while recording

dennisguse:
Nowadays, we should be able to reset the preferences while recording.
Also the import/export functionality should be working.

Only nitpick: the export all feature.
A currently recording track would only include the data that was already recorded.


rgmf (approved):
It seems work like a charm. Great for me.

PR #1252, opened 2022-05-26
Fix EspressoDeleteTrackTest.

### LINKED ISSUE #817: Failing test: espressoDeleteTrackTest
i get:
```
androidx.test.espresso.IdlingResourceTimeoutException: Wait for [de.dennisguse.opentracks.EspressoDeleteTrackTest$MyIdlingResource] to become idle timed out
```

image

PS: Please ignore EspressoCategoryUnitTest as this is a problem with localization setting.

dennisguse:
Fixes #817.


dennisguse:
@rgmf Any idea why this fails on API24?
Somehow, the idling resource seems to be blocking and then nothing happens until timeout (no recording started).

dennisguse:
Got it: we can just check if the list is empty (in the test we always get an empty database) and we need to wait briefly before resuming on API24.

Issue #660, opened 2021-03-17
Small screen layout


sebastianha:
I sometimes use split screen to show OsmAnd and OpenTracks both on my screen. Unfortunately the layout is not optimal for this, see screenshots

![Screenshot_20210317-073755_OpenTracks](https://user-images.githubusercontent.com/420807/111428320-a0568180-86f7-11eb-81e4-197248a7c0bb.png) ![Screenshot_20210317-073811_OpenTracks](https://user-images.githubusercontent.com/420807/111428370-b106f780-86f7-11eb-86a8-e40fdc83beec.png) ![Screenshot_20210317-073819_OpenTracks](https://user-images.githubusercontent.com/420807/111428372-b2382480-86f7-11eb-9169-96e9626d1405.png)

Perhaps it would be possible to detect the very small app layout and hide some elements of make them smaller. This would be really handy.

dennisguse:
@sebastianha Does the 4.0.0 release (i.e., the UI update) solve this problem for you?

sebastianha:
As soon, as it hits f-droid, I will check.

dennisguse:
In addition, we might be able to hide some UI elements when in multi-window mode.
This method allows to get events: https://developer.android.com/reference/android/app/Activity#onMultiWindowModeChanged(boolean,%20android.content.res.Configuration)

sebastianha:
> @sebastianha Does the 4.0.0 release (i.e., the UI update) solve this problem for you?

Hi,

no, unfortunately not completely. When not running it looks much better now:

![signal-2022-05-30-121952](https://user-images.githubusercontent.com/420807/170972283-36063d06-6187-4566-b40b-aa995244c4ec.png)

But as soon as you are on the recording screen not even one row of data is visible.

![signal-2022-05-30-121949](https://user-images.githubusercontent.com/420807/170972480-1046b044-ed87-431d-b2f8-596a8d8e9fba.png)

It would be nice to have one row visible, this should be do-able with hiding the top bar when the screen is below a certain height.

dennisguse:
Can you check how it looks like in Fullscreen mode? (It is in the Open tracks preferences)

sebastianha:
It looks the same. I think the is no "fullscreen" when in "split screen" mode.

rgmf:
> @rgmf In one of the screenshots, the fabButton overlaps the actual content. Any idea why this is happening?

[...]

PR #1265, opened 2022-05-30
Fullscreen: now more fullscreen.

dennisguse:

We could also go for the display cutout, but my phone doesn't support it.
https://developer.android.com/reference/android/view/WindowManager.LayoutParams#LAYOUT_IN_DISPLAY_CUTOUT_MODE_ALWAYS


dennisguse:
Hiding the system bar is sadly not allowed in multi-window mode: https://developer.android.com/guide/topics/large-screens/multi-window-support
Might have been nice for #660.

rgmf (approved):
It's awesome!!! And really full screen!!!

Great investigation ;)

PD/ It's working very well on my phone. In split window it doesn't work.

Issue #1285, opened 2022-06-16
Voice announcements incorrectly round the fractional part


WPettersson:
I'm using v4.0.1 (code 4926) from f-droid, and occasionally I'll notice that when I'm running around a 10km/h pace, the voice announcements will say 9.0km/h. This happens most noticeably with the lap announcements, for some reason. I know my pace well enough to know that it must mean 10.0km/h, based on overall track statistics. I did some digging and found https://github.com/OpenTracksApp/OpenTracks/blob/1c3a2bd2e419dc9f5cf3446d2e8316e0b90a8996/src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncementUtils.java#L110

I see what the intention with this line is, but it seems Java does by default using rounding when using String.format. This can be seen with the following Java test:

```
public class Main
{
    public static void main(String[] args) {
        double speed  = 9.9999;
        long integral = (long)speed;
        String frac = String.format("%.1f", speed - integral);
        System.out.println("speed: " + speed + " integral: "  + integral + " frac: " + frac);
    }
}
```
The output is `speed: 9.9999 integral: 9 frac: 1.0` which would explain the odd announcements. This can be tested online at https://www.online-java.com/6XT9vwJcZi as well.

I think I see a few possible solutions. One is to disable rounding in String.format somehow (I had a search, but couldn't work out how to do this properly). Another might be to print the value to a string, and then extract integral and decimal portions from the string itself. Or even use BigDecimal which might be an easier way.

dennisguse:
Actually yes :)
The use of `String.format()` didn't feel very nice.

https://www.baeldung.com/java-separate-double-into-integer-decimal-parts

@WPettersson do you have time to fix this?

WPettersson:
I've thrown a bug fix to https://github.com/OpenTracksApp/OpenTracks/pull/1286 which "should" be correct. However, I don't have any android dev environment setup so I haven't actually run the tests or checked the formatting.

WPettersson:
Okay, seems test 30 worked, test 24 failed (looks unrelated to my changes) and linting passed :)

dennisguse:
@WPettersson welcome to the world of Android in CI; tests are flacky as hell - especially the UI tests :)

PR #1286, opened 2022-06-16
Fix voice announcement rounding problem

WPettersson:
We extracted integral parts before rounding, but String.format does
rounding which caused occasional issues. This fixes that by rounding all
values as we fetch them, so the String.format rounding won't break
anything.

Fixes: #1285

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncementUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncementUtils.java
@@ -65,12 +65,9 @@ class VoiceAnnouncementUtils {
 
         if (shouldVoiceAnnounceTotalDistance()) {
             builder.append(context.getString(R.string.total_distance));
-            long distanceIntegerPart = (long) distanceInUnit;
-            // Extract the decimal part
-            String distanceFractionalPart = String.format("%.2f", (distanceInUnit - distanceIntegerPart)).substring(2);
             // Units should always be english singular for TTS.
             // See https://developer.android.com/reference/android/text/style/TtsSpan?hl=en#TYPE_MEASURE
-            appendDecimalUnit(builder, context.getResources().getQuantityString(distanceId, getQuantityCount(distanceInUnit), distanceInUnit), distanceIntegerPart, distanceFractionalPart, unitDistanceTTS);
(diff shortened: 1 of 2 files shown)


src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncementUtils.java
```diff
@@ -61,7 +61,9 @@ static Spannable getAnnouncement(Context context, TrackStatistics trackStatistic
                 throw new RuntimeException("Not implemented");
         }
 
-        double distanceInUnit = distance.toKM_Miles(unitSystem);
+        // Total distance is announced to two decimal places, so round early to
+        // avoid rounding issues (see #1285)
+        double distanceInUnit = Math.round(distance.toKM_Miles(unitSystem) * 100) / 100.0;
```
dennisguse, line 66: extract into a method?

dennisguse (commented):
Adding a test to VoiceAnnouncementUtilsTest would be great.

And why not use BigDecimal?

src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncementUtils.java
WPettersson: I was thinking about adjusting `appendDecimalUnit` to instead take a double, and a precision, which seemed to make sense.

WPettersson:
> Adding a test to VoiceAnnouncementUtilsTest would be great.

Think I can manage that, probably not today though.

> And why not use BigDecimal?

Even with BigDecimal, we need to do rounding before splitting into integer and fractional parts. I had a BigDecimal version worked out, but realised that using BigDecimal is unnecessary as I doubt we'll overflow a double.

WPettersson:
Pushed a new commit which puts the rounding code inside `appendDecimalUnit` and also has some tests that pass and should be checking for this issue in the future.

dennisguse:
Looks good.
Same question: why not use big decimal?

WPettersson:
Even with BigDecimal, we need to do rounding before splitting into integer and fractional parts. I had a BigDecimal version worked out, but realised that using BigDecimal is unnecessary as I doubt we'll overflow a double.

PR #1293, opened 2022-06-21
Statistics: remove speed filtering.

dennisguse:
Part of #1003.

I am kind of annoyed by all the max speed issues we have - let's start by removing stuff, we don't understand...


dennisguse:
@vlmendz @ae5960e8-a6fc-491f-b252-898ecf59af95 @vsevjednom-cz   do you have some time to test this?
If doesn't completely misbehave, I would merge it.
And then let's see how we continue from there.

vsevjednom-cz:
> do you have some time to test this?

Of course but please, write me how to do it automatically in Android studio. I've done it before, but I don't remember. I don't do it often.

src/main/java/de/dennisguse/opentracks/stats/TrackStatisticsUpdater.java
```diff
@@ -218,29 +197,12 @@ private void updateSpeed(@NonNull TrackPoint trackPoint, @NonNull TrackPoint las
 
     private boolean isValidSpeed(@NonNull TrackPoint trackPoint, @NonNull TrackPoint lastTrackPoint) {
         // There are a lot of noisy speed readings. Do the cheapest checks first, most expensive last.
-        if (trackPoint.getSpeed().isZero()) {
-            return false;
-        }
-
         Duration timeDifference = Duration.between(lastTrackPoint.getTime(), trackPoint.getTime());
         Speed maxAcceleration = Speed.of(MAX_ACCELERATION * timeDifference.toMillis());
```
rgmf, line 201: I think, it should be `MAX_ACCELERATION * timeDifference.getSeconds()`.

rgmf (commented):
Only a tiny thing.

vsevjednom-cz:
> @vsevjednom-cz Just look for the `command line instructions` from GitHub 

I'm afraid of deleting my edits and fragments 😱. Sorry.
This is sounds better: full code with code changes - do you have it somewhere?

src/main/java/de/dennisguse/opentracks/stats/TrackStatisticsUpdater.java
```diff
@@ -41,23 +40,15 @@
...
-    private static final int SPEED_SMOOTHING_FACTOR = 25;
-
     private static final String TAG = TrackStatisticsUpdater.class.getSimpleName();
     /**
      * Ignore any acceleration faster than this.
      * Will ignore any speeds that imply acceleration greater than 2g's
-     * 2g = 19.6 m/s^2 = 0.0002 m/ms^2 = 0.02 m/(m*ms)
      */
-    private static final double MAX_ACCELERATION = 0.02;
+    private static final double SPEED_MAX_ACCELERATION = 2 * 9.81;
```
rgmf, line 48: Where 9.81 value comes from? Why not to use 19.6 value directly?

dennisguse: 9.81 is one G (gravity earth).
It is just documenting that it is used.

ae5960e8-a6fc-491f-b252-898ecf59af95:
I see this hit F-Droid today. Does this change only affect new tracks? My max speeds for past tracks recorded with a Bluetooth speed sensor still show up as being almost half of what they should be.

dennisguse:
@ae5960e8-a6fc-491f-b252-898ecf59af95 no, you would have to export, delete, and import the tracks. Max speed is computed while storing a tracks.

PR #1307, opened 2022-07-01
Gain loss delay

dennisguse:
We don't need to process the pressure sensor as fast we were doing it.
5s should even be way to fast for most cases like hiking and cycling.

@rgmf could you test in some mountainish areas?

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/sensors/AltitudeSumManager.java
+++ b/src/main/java/de/dennisguse/opentracks/sensors/AltitudeSumManager.java
@@ -5,6 +5,7 @@ import android.hardware.Sensor;
 import android.hardware.SensorEvent;
 import android.hardware.SensorEventListener;
 import android.hardware.SensorManager;
+import android.os.Handler;
 import android.util.Log;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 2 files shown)


crankc:
I tested this briefly by just walking and swinging the phone around. The app no longer registers the gain/loss changes as quickly as before, and the total accumulated values are now smaller, but it is still too sensitive on my Pixel 5. It registered 30 meters gain and loss in about a minute of just walking around my house. It's less than before, but still. Just for the record, i tested it on another Pixel 5 in case my phone was defective, but the same issue occurs on the other phone too.

A thought; is it possible to add an optional setting to disable gain/loss recording (= disable barometer) in the app? Then the data wouldn't be recorded or shown.

In any case, this change seems to make a difference but i fear it will still make gain/loss data so inaccurate it is useless when cycling, at least if the phone is in my pocket. I can test this again when keeping the Pixel 5 attached on the handlebar instead of my pocket (i usually use an older phone without a barometer for that, because i fear the shaking will break the camera OIS on the Pixel) but it will take a week or so before i get the chance to go cycling again.

rgmf:
> @rgmf could you test in some mountainish areas?

I'm going to install this branch for tomorrow riding ;) I'll tell you tomorrow.

rgmf:
@dennisguse it's too pesimistic.

In today's route I've ridden 108km with a gain of 600m but OpenTracks got 480m.

I've noticed the more we climb the bigger the difference between OpenTracks and real gain.

Here are the stats with differents bike computers (from some of my friends):

| | OpenTracks|Garmin Edge 530|Bryton Rider 450 |
|-|-----------|------------------|-------------|
|Gain|480m|603m|594m|

What I forgot was to record the route with OpenTracks F-Droid version to compare to this branch. Monday I can test OpenTracks This Branch vs OpenTracks F-Droid version.

rgmf (approved):
After several more tests it seems to work like a charm, at least as current F-Droid version does.

PR #1310, opened 2022-07-01
[Wait for AGP 7.3.0] Refactoring things regarding Duration, Speed, and Distance

dennisguse:
@pstorch uses desugar 1.20 using AGP 7.3.0-beta04.
It gives us the full implementation of Duration.
I was waiting for this since ages :sunglasses:


rgmf:
That sounds like a good news :clap: :clap: :clap: 

I've not been able to test it because Android Studio complains about Gradle 7.3.0-beta04 version. I'll try to re-install Android Studio.

dennisguse:
@rgmf it requires AndroidStudio Dolphin aka the current beta.

pstorch:
Sounds good, but I hope it doesn't take long to be released. I'm hesitating to use the beta version.

dennisguse:
@pstorch I share the sentiment.
And release will take something between 2 weeks to 6 month :)
If we don't change too many of these things that are related, we can wait for some time.
If we have to, then I would merge this beforehand.

Until then, I will just rebase it and check if it still works (aka no merge conflicts).
If we get some, I would merge it before the final release to save some time on our side.

PR #1318, opened 2022-07-05
Rough draft for export all tracks into one file

dennisguse:
I created a rough draft to export all tracks into one file.

TODO:
- [x] GPX/KML/KMZ can export multiple tracks into one file (was already done)
- [x] add setting to export into one file
- [x] send export task to ExportService
- [x] ExportService actually exports

@rgmf I was looking through the ExportService (and the communcation with it) and I am not sure how to do this.
Any suggestions?

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
@@ -208,7 +208,7 @@ public class ExportImportTest {
 
         // when
         // 1. export
-        trackExporter.writeTrack(track, context.getContentResolver().openOutputStream(tmpFileUri));
+        trackExporter.writeTrack(List.of(track), context.getContentResolver().openOutputStream(tmpFileUri));
         contentProviderUtils.deleteTrack(context, trackId);
 
         // 2. import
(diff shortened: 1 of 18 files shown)


dennisguse:
@rgmf if you have some, it would be great if you could take a look at this.
Took me quite a while to figure out how to do this - and it is not yet the best solution.
... but it works :D

pstorch:
I would like to test it, but can currently not switch to AS beta.

dennisguse:
@pstorch the build should also work with AGP 8.0.X - the only thing that will be missing is Android 13's app-specific language selector.

pstorch:
I exported > 100 tracks. File created without error. 
Looking for a tool which is capable to import and show the tracks. :wink:

A small issue: it reports that only 1 track was exported.

Screenshot_20230528_182643_OpenTracks

pstorch:
Maybe it's not a good idea to visualize it in another tool. As this is meant for backup. So I deleted all tracks and tried to import again. But got this error:
Screenshot_20230528_190629_OpenTracks

dennisguse:
@pstorch Amazing!
First: You found a bug in the KML exporter not creating valid XML for multiple tracks (this must  be ancient).
Second: for some mime types Android adds file suffixes (KML is one, but not GPX) :sunglasses: 
Translations for english updated.

The UI still lacks quite a lot, but it works - so our users will have to live with it aka MVP-style.

pstorch:
Now it works. I see minimal differences in the overall statistic after reimport the big backup file. With this amount of files it might be negligible. 

As you said, the UI can be improved afterwards. It would help for example if the import counts the imported tracks to see some progress.

dennisguse:
@pstorch thanks!!!

PR #1325, opened 2022-07-10
Store sensor-only data (without distance)

### LINKED ISSUE #1088: Record heartrate history even if position is unavailable or does not change
Currently OpenTracks only saves history along with the GPS track, including the heart rate.

It would be useful if one could use the app to record their heartrate when they, for example, do a workout in a gym, where GPS position is be unknown, or known, but constant.

One way to represent it would be to make a KML track with dummy / constant trackpoints that still go every second even when position does not change, with `<when>` element only, and fill the heartrate section in `ExtendedData` with real data.  (There could be other ways too.)

dennisguse:
Stores sensor-only data like heartrate without requiring GPS/distance data.
Allows to train indoor and still collect heartrate or powemeter over time.

Fixes #1088.

Note: not yet tested.

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
@@ -155,26 +155,31 @@ public class ExportImportTest {
         mockAltitudeChange(trackPointCreator, 1);
         mockBLESensorData(trackPointCreator, 15f, sensorDistance, 66f, 3f, 50f);
 
-        trackPointCreator.setClock("2020-02-02T02:02:05Z");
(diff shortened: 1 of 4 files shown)


dennisguse:
Just tested it - and it works like charm :sunglasses: 
Yesterday, I was on a train and had no sensor to test...

rgmf (approved):
It works like a charm :+1:

dennisguse:
@rgmf Awesome! 
I was expecting some more NPEs, but I won't complain :nerd_face:

rgmf:
> @rgmf Awesome! I was expecting some more NPEs, but I won't complain 🤓

So, great job 👏 

I have used it on both an indoor activity with HR sensor and a riding activity with HR, Cadence and Speed sensor, and no NPE.

dennisguse:
@rgmf it was just luck :)

Issue #1329, opened 2022-07-16
PlayStore release of 4.0.4 rejected


dennisguse:
The 4.0.4 was rejected by the PlayStore due to an impersonation violation - they complain about the icon and the title.
I submit an "Advance Notice Form" - let's see how long it takes.

Until then 4.0.4 will be _not found_ in the PlayStore.
And all future releases may be rejected as well.

PS/ interestingly 4.0.3 is still available, so okay.

![image](https://user-images.githubusercontent.com/2820825/179349032-3cadfa28-317e-4596-ba0b-ce0c46030a50.png)

pstorch:
This gets more and more rediculous.

rgmf:
🤬🤬

pstorch:
And they don't tell which app it is accused of impersonating?

pstorch:
DirtTracks maybe?
https://play.google.com/store/apps/details?id=com.zihua.android.dirttracks

Or MyTracks from the same dev?

dennisguse:
Actually no.
The screenshot contains all the info I got.
I send them feedback using three different channels.

Anyhow, if they are going after copy cats it is not all bad.
Let's see.

pstorch:
> Anyhow, if they are going after copy cats it is not all bad.

Yes, that's right. It's just that they don't give enough information and often don't respond to questions.

dennisguse:
Jason takes care :)

```
Hi Developer at Dennis Guse,

Thanks for contacting the Google Play team. 

I've received your appeal and I appreciate your patience while I look into it.

I'll let you know as soon as I have any additional information to share. 

Please let me know if you have any questions in the meantime.

Regards,
Jason
The Google Play Team
```

dennisguse:
Another name friend: https://play.google.com/store/apps/details?id=app.opentracks&hl=de&gl=US

dennisguse:
And we are impersonating MyTracks, but I am not exactly sure which one...
* https://play.google.com/store/apps/details?id=com.zihua.android.mytracks
* https://play.google.com/store/apps/details?id=com.google.android.maps.mytracks
  (de-published Google MyTracks; only visible if you installed it and are logged in into the PlayStore)

![image](https://user-images.githubusercontent.com/2820825/179449253-d1589f48-062b-46f2-9acf-296d6b54ade3.png)

dennisguse:
My response (2022-07-18)
```
Hi,

[...]

PR #1336, opened 2022-07-24
PlayStore: use an inverted icon (app launcher only).

dennisguse:
Part of #1329.

If I didn't do anything completely wrong, buildTypes only allow us to modify _some_ resources (like app_icon).
So, to replace the internally used app_icon, we would need to use buildFlavors as well.

https://developer.android.com/studio/build/build-variants#groovy

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -100,6 +100,7 @@ android {
         releasePlayStore {
             applicationIdSuffix ".playstore"
             versionNameSuffix "-PlayStore"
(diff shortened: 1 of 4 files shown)


pstorch:
If it is sufficient for Google Play, then I'm fine with it

pstorch (approved):

dennisguse:
No clue. I would just try it out -
SPAMing them with releases ;)

PR #1343, opened 2022-08-02
EGM2008 offset lookup  for -longitude was wrong.

dennisguse:
Some days ago I got an email stating that our EGM2008 implementation is misbehaving for negative longitudes.

Provided test data:
```
Example 1 (Lat, Lon in degs): 33.628149, -117.725816 (Irvine, CA, US)
Geoid height (aka "h" in your code) should be ~ -34.5m
Opentracks as-is: -35.4m
Opentracks with proposed fix: -34.2m

Example 2 (Lat, Lon in degs): 37, -42 (middle of Atlantic ocean!)
Geoid height (aka "h" in your code) should be ~ +22.8m
Opentracks as-is: -35.637m
Opentracks with proposed fix: +22.8m

Example 3 (Lat, Lon in degs): -15.78303, -47.93330 (Brasilia, Brazil)
Geoid height (aka "h" in your code) should be ~ -12.66m
Opentracks as-is: -10.04m
Opentracks with proposed fix: -12.68

Example 4 (Lat, Lon in degs): 47.63153, -122.30938 (Seattle, WA, US)
Geoid height (aka "h" in your code) should be ~ -22.8m
Opentracks as-is: -4.29m
Opentracks with proposed fix: -22.99m
```

Proposed solution: 
Replace `longitude = 180 + Math.abs(location.getLongitude());`
With `longitude = 360 + location.getLongitude();`

Seemingly the data layout of the EGM2008 data is different than I figured out back then.

Documentation: https://geographiclib.sourceforge.io/C++/doc/geoid.html#geoidformat

GeopraphicLib code: https://github.com/ObjSal/GeographicLib/blob/bcc64710cabae6f9313e1ac109a1f9eb5e67fd21/src/Geoid.cpp#L451


src/main/java/de/dennisguse/opentracks/util/EGM2008Utils.java
```diff
@@ -65,7 +65,7 @@ static Indices getIndices(Location location) {
         if (location.getLongitude() >= 0) {
             longitude = location.getLongitude();
         } else {
-            longitude = 180 + Math.abs(location.getLongitude());
+            longitude = 360 + location.getLongitude();
```
gdt, line 68: So here you are converting -180 to 0 into 180 to 360.   This is basically normalizing the value into [0,360) east longtitude.   I missed earlier that it's happening only for negative longitude.

src/main/java/de/dennisguse/opentracks/util/EGM2008Utils.java
```diff
@@ -65,7 +65,7 @@ static Indices getIndices(Location location) {
...
         }
         int longitudeIndex = (int) (longitude * RESOLUTION_IN_MINUTES);
```
gdt, line 70: Now I follow and this seems ok.  (comment in PR was earlier).  I do think this is needlessly hard to follow because of the if structure, rather than
  get longtitude
  if longitude is negative, += 360 with comment that we are doing range normalization from (-180,180] to [0,360)

Issue #1340, opened 2022-07-28
Deleting a recorded track takes too long


Saboe:
**Describe the bug**
When I delete a track it takes up to several minutes until the track is deleted. Few days ago while I was deleting several tracks I got the following message from Android:

![image](https://user-images.githubusercontent.com/22373148/181400755-6d9cf424-3c53-4cbc-8028-38cbf99dc673.png)

This issue I've already experienced several month ago. Yesterday, when I've deleted the last recorded track, it took around 50 seconds of time. Deleting several tracks while deleting of one track is in progress it could lead to phone freezes and the shown screenshot above.

**Technical information**
 - Device: Fairphone 3
 - OS: Android 11, build 8901.4.A.0010 release keys
 - OpenTracks version: 4.0.3

dennisguse:
One reason is that after each track deletion we are vacuuming the db. An optimization would be to do this only once at the end.

Adding more tracks to be deleted should not lead to an ANR. We have to check why this is happening.

pstorch:
I remember I once implemented a mass delete function for tracks to have the VACUUM triggered only one time after all tracks are deleted and not after each one.
This got changed to individual deletion with the track deletion service which had a notification with progress bar.
In PR https://github.com/OpenTracksApp/OpenTracks/pull/1358 I changed this to one SQL call and an indeterminate progress. So now the long running vacuum happens only once.

But I think we should find another way to do the VACUUM. Not automatically after each delete. What about an extra DB maintenance task via the settings? So the user can trigger it manually. When can also remember the user from time to time or after such a mass delete.

dennisguse:
We discussed this about 2 years ago and I was too cautious.
Vacuum every now and then is sufficient. Who needs space right? ;)

dennisguse:
One solution: https://stackoverflow.com/questions/36126842/count-deleted-records-after-a-delete

pstorch:
The deletion count could be a solution.
Regarding the space: according to SQLite the free space in the db file is not lost. SQLite will reuse it to store new data. It will just not shrink the file without vacuuming. 

Another observation because @Saboe is reporting that it takes minutes. I realized it takes far longer in debug mode than in normal release. You don't run an OT debug build, do you?

pstorch:
How do we want to implement the deletion count?

- remove the auto-vacuum on delete
- count all deletes
- check on startup of OT if "threshold" is reached and ask the user to perform a vacuum and reset count

[...]

PR #1358, opened 2022-08-19
Improve #1340 mass track deletion time

pstorch:
Delete all tracks with one DB call instead of one call for each track. This reduces the DB vacuum to happen only once.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/data/ContentProviderUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/data/ContentProviderUtils.java
@@ -187,7 +187,6 @@ public class ContentProviderUtils {
             FileUtils.deleteDirectoryRecurse(FileUtils.getPhotoDir(context, trackId));
         }
 
-        // Delete track last since it triggers a database vacuum call
         String whereClause = String.format(TracksColumns._ID + " IN (%s)", TextUtils.join(",", Collections.nCopies(trackIds.size(), "?")));
         contentResolver.delete(TracksColumns.CONTENT_URI, whereClause, trackIds.stream().map(id -> Long.toString(id.getId())).toArray(String[]::new));
     }
@@ -195,8 +194,6 @@ public class ContentProviderUtils {
     public void deleteTrack(Context context, @NonNull Track.Id trackId) {
         // Delete track folder resources.
         FileUtils.deleteDirectoryRecurse(FileUtils.getPhotoDir(context, trackId));
-
-        // Delete track last since it triggers a database vacuum call
         contentResolver.delete(TracksColumns.CONTENT_URI, TracksColumns._ID + "=?", new String[]{Long.toString(trackId.getId())});
     }
 
--- a/src/main/java/de/dennisguse/opentracks/data/CustomContentProvider.java
+++ b/src/main/java/de/dennisguse/opentracks/data/CustomContentProvider.java
@@ -27,6 +27,7 @@ import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteException;
 import android.database.sqlite.SQLiteQueryBuilder;
 import android.net.Uri;
+import android.os.Bundle;
 import android.text.TextUtils;
 import android.util.Log;
 
@@ -39,6 +40,7 @@ import de.dennisguse.opentracks.data.models.TrackPoint;
 import de.dennisguse.opentracks.data.tables.MarkerColumns;
 import de.dennisguse.opentracks.data.tables.TrackPointsColumns;
 import de.dennisguse.opentracks.data.tables.TracksColumns;
+import de.dennisguse.opentracks.settings.PreferencesUtils;
 
 /**
  * A {@link ContentProvider} that handles access to track points, tracks, and markers tables.
@@ -52,6 +54,8 @@ public class CustomContentProvider extends ContentProvider {
     private static final String TAG = CustomContentProvider.class.getSimpleName();
 
     private static final String SQL_LIST_DELIMITER = ",";
+
+    private static final int TOTAL_DELETED_ROWS_VACUUM_THRESHOLD = 10000;
 
     private final UriMatcher uriMatcher;
 
@@ -131,14 +135,12 @@ public class CustomContentProvider extends ContentProvider {
     @Override
     public int delete(@NonNull Uri url, String where, String[] selectionArgs) {
         String table;
-        boolean shouldVacuum = false;
         switch (getUrlType(url)) {
             case TRACKPOINTS:
                 table = TrackPointsColumns.TABLE_NAME;
                 break;
             case TRACKS:
                 table = TracksColumns.TABLE_NAME;
-                shouldVacuum = true;
                 break;
             case MARKERS:
                 table = MarkerColumns.TABLE_NAME;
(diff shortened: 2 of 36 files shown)


dennisguse:
Looks okay too me.
One more thing: we could use a shared preference to, say, count deleted tracks and vacuum when a threshold is reached.

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
```diff
@@ -137,35 +122,28 @@ public TrackDeleteService getService() {
...
+            return this.trackIds != null && this.trackIds.contains(trackId);
         }
 
         @NonNull
         @Override
         public String toString() {
             return "DeleteStatus{" +
-                    "number of deletes=" + progress +
-                    ", total=" + max +
+                    "trackIds=" + trackIds +
```
dennisguse, line 146: This should not work as toString is not nicely implemented on ArrayList.

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
```diff
@@ -137,35 +122,28 @@ public TrackDeleteService getService() {
...
-        private DeleteStatus(@Nullable Track.Id trackId, int progress, int max) {
-            this.trackId = trackId;
-            this.progress = progress;
-            this.max = max;
+        private DeleteStatus(@Nullable List<Track.Id> trackIds) {
+            this.trackIds = trackIds;
         }
 
         public boolean isFinished() {
-            return progress == max;
```
dennisguse, line 156: This is not needed anymore?

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
```diff
@@ -137,35 +122,28 @@ public TrackDeleteService getService() {
...
+            return true;
         }
 
         public boolean isDeleted(Track.Id trackId) {
-            return this.trackId != null && this.trackId.equals(trackId);
+            return this.trackIds != null && this.trackIds.contains(trackId);
```
dennisguse, line 139: Can also be removed, right?

src/main/res/values/strings.xml
```diff
@@ -494,7 +494,7 @@ limitations under the License.
     <string name="track_delete_one_confirm_title">Delete track?</string>
     <string name="track_delete_multiple_confirm_message">The selected tracks and their markers will be permanently deleted from the device.</string>
     <string name="track_delete_progress_message">Deleting&#8230;</string>
-    <string name="track_delete_progress">%1$d/%2$d</string>
```
dennisguse, line 497: need to be removed in the other translations as well.

dennisguse (commented):
Some things

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
pstorch: This is properly implemented in a super class of ArrayList: https://cs.android.com/android/platform/superproject/+/master:libcore/ojluni/src/main/java/java/util/AbstractCollection.java;drc=12fae6f00fd514c510fa58f64c8870ab062733fd;l=464

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
pstorch: Actually the whole `isFinished()` could be removed, because the TrackDeletionService only emmits the `DeleteStatus` once at the end. So we can remove it completely.

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
pstorch: The `TrackRecordedActivity` is checking it. It could be the case that the currently recorded track was in the list of to be deleted tracks, but was filtered out, because it is not allowed to delete it.

src/main/res/values/strings.xml
pstorch: Ok, I thought this will be handled by weblate later on. No problem, I'll remove them.

dennisguse: It triggers linter issues; a while ago this could be done in AndroidStudio (with this translation UI), but at least that stopped working a while ago...

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
pstorch: I've renamed the `DeleteStatus` class to `DeletionFinishedStatus` to make this clear.

src/main/java/de/dennisguse/opentracks/data/ContentProviderUtils.java
```diff
@@ -181,23 +181,24 @@ public void deleteAllTracks(Context context) {
         FileUtils.deleteDirectoryRecurse(dir);
     }
 
-    public void deleteTracks(Context context, @NonNull List<Track.Id> trackIds) {
+    public int deleteTracks(Context context, @NonNull List<Track.Id> trackIds) {
```
dennisguse, line 184: Can we get JavaDoc for the return (it is cout), but not that is not that obvious.

src/main/java/de/dennisguse/opentracks/data/CustomContentProvider.java
```diff
@@ -148,22 +147,38 @@ public int delete(@NonNull Uri url, String where, String[] selectionArgs) {
...
+        try (Cursor cursor = db.rawQuery("SELECT total_changes()", null)) {
+            cursor.moveToNext();
+            totalCount = cursor.getInt(0);
+        }
+        return totalCount;
+    }
+
+    @Override
+    public Bundle call(String method, String arg, Bundle extras) {
+        if (method.equals("vacuum")) {
```
dennisguse, line 177: At least the string should be a constant.

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
```diff
@@ -59,22 +62,25 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
         serviceExecutor.execute(() -> {
             ContentProviderUtils contentProviderUtils = new ContentProviderUtils(this);
-            for (int i = 0; i < trackIds.size(); i++) {
-                updateNotification(i + 1, trackIds.size());
-                contentProviderUtils.deleteTrack(this, trackIds.get(i));
-                sendResult(trackIds.get(i), i + 1, trackIds.size());
+            int rowsDeleted = contentProviderUtils.deleteTracks(this, trackIds);
+            PreferencesUtils.addTotalRowsDeleted(rowsDeleted);
+            int totalRowsDeleted = PreferencesUtils.getTotalRowsDeleted();
+            if (totalRowsDeleted > TOTAL_DELETED_ROWS_VACUUM_THRESHOLD) {
```
dennisguse, line 71: We could move this decision as a static method into the CustomContentProvider (incl. the TOTAL_DELETED_ROWS_VACUUM_THRESHOLD); then everything DB related is one place and TrackDeleteService would just call it.

dennisguse (approved):
Looks still good.
Adds quite some complexity, but I guess it is worth it.

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
pstorch: Yes, that would reduce some complexity.
I did it this way, so the TrackDeleteService can update the notification that the vacuuming is in progress.
But if this info is not so important, we can simplify the code.

src/main/java/de/dennisguse/opentracks/data/CustomContentProvider.java
pstorch: Sure

src/main/java/de/dennisguse/opentracks/data/ContentProviderUtils.java
pstorch: Will add it

dennisguse:
@pstorch merge whenever you feel it is ready :)
I was planning for a new release anyhow in the next days...

src/main/java/de/dennisguse/opentracks/data/ContentProviderUtils.java
pstorch: Return value not needed anymore

src/main/java/de/dennisguse/opentracks/services/TrackDeleteService.java
pstorch: Nearly 20 lines less code (and complexity) if we move this logic as proposed into the CustomContentProvider. That's not worth to keep for the notification which is hard to find anyway. I've simplified it. Thanks for the suggestion.

PR #1362, opened 2022-08-22
Remove icon from bottombars.

dennisguse:
Saves screen space and (hopefully) less complaints from PlayStore.

Part of #1329.

TL;DR

I am still "working" with the Google people to get accept a new version of OpenTracks.

### DIFF
--- a/src/main/res/layout/about.xml
+++ b/src/main/res/layout/about.xml
@@ -20,28 +20,13 @@
                 android:orientation="vertical"
                 android:padding="20dp">
 
-                <LinearLayout
+                <TextView
+                    android:id="@+id/about_text_description"
                     android:layout_width="match_parent"
-                    android:layout_height="wrap_content"
-                    android:orientation="horizontal"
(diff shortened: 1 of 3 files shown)


pstorch:
The version with the new gradle build tools doesn't work on my system. But I could build and run it when reverting it.
So yes, icons are gone from the bottom bar and the about dialog.

dennisguse:
I am confused  about this whole desugar thing.

I am running AndroidStudios Beta in parallel as I always hope they fixed some bugs...


PS does the whole icon removal make sense?

pstorch:
> PS does the whole icon removal make sense?

It is not as pretty as with icon, but if the icon is the problem, then ...

rgmf (approved):
It's strange for me because It got used to seeing it, but It's good for me.

dennisguse:
Felt the same for me.
For now v4.0.6 got accepted; let's build a solution if need it next time ;)

Issue #1118, opened 2022-02-04
Heart rate zones and graph in the real of time


MrS1R1US:
**Describe the solution you'd like**
There is a very lack of heart rate zones, as in Polar Beat. When the heart rate zones are set and there is a real-time color graph, it is easier to follow the correct heart rate. This is a very important schedule during a serious workout. 

**Additional context**
[This is what the final chart looks like. Also, the training graph with a line is built heart rate, which allows you to see the line that you cannot cross. ](https://i.imgur.com/hsvnR75.jpeg)

P.S. You took into account my previous wish regarding automatic backups to the phone and via Syncthing to the PC, for which I am very grateful to you!

MrS1R1US:
[This is how the graph looks like during a workout, the line moves depending on the heart rate zone. Over time, when the workout is long, the first heart rate point goes beyond the graph in order to make the graph smoother. Please note that the heart rate numbers themselves change color depending on the heart rate zone, this is very helpful for real-time analysis to slow down or increase it. ](https://i.imgur.com/tOMxAsX.png)

This is how I would like to see it in OpenTracks:
[https://i.imgur.com/MPQ7DYg.png](https://i.imgur.com/MPQ7DYg.png)

This is a very important function, to see in which pulse zone you are. This feature is available in all sports devices to which a heart rate monitor is connected. In the settings you set the maximum heart rate, and then it is calculated down from 100%. It is also worth noting that it is important to save the graph at the end of the workout, this will allow you to compare with past changes.
Zone 5 - 90-100%
Zone 4 - 80-90%
Zone 3 - 70-80%
Zone 2 - 60-70%
Zone 1 - 50-60%

MrS1R1US:
Please pay due attention to this. )

dennisguse:
@MrS1R1US somebody might if somebody has time ;)
PS/ it would be helpful if you upload the images to Github instead of imgur.

Is this issue different than #185?

MrS1R1US:
> PS/ it would be helpful if you upload the images to Github instead of imgur.
Ok

> Is this issue different than #185?

This topic complements and correctly reveals the question) 

P.S. After 2 years, no one wants to do this? This is an important functionality for a sports application.

dennisguse:
From my side, it is lack of time and I rather did other things in OpenTracks (mainly focusing on sensors and fixing internals).
So yeah - it was not a high priority and seemingly everybody else felt the same or there are not enough people out there that could contribute.
[...]

Issue #24, opened 2019-08-11
Heartrate bluetooth sensor: show battery info


dennisguse:

gordonel:
**Is your feature request related to a problem? Please describe.**
I'd like to be notified if it's time to charge my devices or replace the batteries in advance.

**Describe the solution you'd like**
When starting an activity, OpenTracks should read the battery level if available and notify the user if it falls below a user-specified level. Also, the app can provide convenient readouts in the notifications drawer as extra info

**Describe alternatives you've considered**
reading battery levels through nRF connect before a workout

dennisguse:
Parsing is already there; we "just" miss the UI.
Probably a notification would be sufficient in case of low battery (=<15%)?

edern-appere:
Yes, a notification would be great !

Would it be possible to show the notification "at any time" ? For example, when starting the application, if the battery of sensor is already low (=< 15%) displaying the notification would inform that the sensor won't last very long before begining a road trip...

PR #1386, opened 2022-09-29
Refactor#1118

dennisguse:
Some refactoring to enable #1118.

- [x] ~move toCSV and fromCSV for DataField into one place~
- [x] ~Remove Parceable from Layout and DataField~: used by:         `intent.putExtra(SettingsCustomLayoutEditActivity.EXTRA_LAYOUT, layout);`

- [x] move data extraction from StatisticsDataBuilder into StatisticData (use subclasses)
- [x] make to be inflated layout depending on DataField
- [x] ~use reflection to create new ViewHolder (`HashMap<DataFieldKey, class>`?)~ use a Map with Lambda for construction
- [x] Settings: figure out how to get the title (again)
- [x] rename Layout to `OpenTracksLayout/StatisticsLayout` RecordingLayout (or whatever)
- [x] ~StatisticsRecordingFragment: re-add divider (HorizontalLineStyle)~ not yet done; kind of complicated
- [x] ~StatisticsRecordedFragment: re-add sensor data~ (moved to #1426)


dennisguse:
@rgmf quick question: why did you chose to use a RecyclerView for the StatisticsUI?
Somehow it feels like adding a lot of complexity just for rendering up to 10 views. 

And I am currently thinking is towards introducing separate layouts: one generic, one for clock, one for coordinates, one for BLE sensors (add battery level).
For data presentation, I was thinking about subclasses and thus separating the large if-block.

rgmf:
> @rgmf quick question: why did you chose to use a RecyclerView for the StatisticsUI? Somehow it feels like adding a lot of complexity just for rendering up to 10 views.
> 

Well, I usually use RecyclerView for lists. It seems easy to me though there are a lot of complexity like you say.

> And I am currently thinking is towards introducing separate layouts: one generic, one for clock, one for coordinates, one for BLE  sensors (add battery level).
> For data presentation, I was thinking about subclasses and thus separating the large if-block.

That sounds like a good idea.

dennisguse:
@rgmf thanks :)

dennisguse:
@rgmf Another question: why did you compute the statistics for the UI in  a separate thread (in StatisticsDataModel)?

rgmf:
> @rgmf Another question: why did you compute the statistics for the UI in a separate thread (in StatisticsDataModel)?

Because I've always made database operations in a separate thread. I guess there are db operations because I'm in the phone now. Tomorry I'll check the code out.

dennisguse:
Took forever, but it is not even feature identical.

@rgmf @pstorch I really love working with all this frontend things, so this tooook way too long and was not that much fun.
Anyhow: we have now the ability to use a separate layout per DataField, so we can implement things like #1118 (e.g., presenting the current heart rate zone using some fancy UI) or #24.
For the UI I dropped the RecyclerView as it was adding too much complexity (we might re-add it later).

Anyhow: feedback appreciated.

PR #1402, opened 2022-10-19
Fixed bad string for an answer in the FAQ

Lorin-lab:
The question "Does OpenTracks require/use Google Play Service?" used "@android:string/no" as answer. But actually the value of this string is "Cancel". So i add new string.


dennisguse:
Thanks;

Can you remove the merge commit?

Lorin-lab:
Done

Issue #1328, opened 2022-07-15
Crash when stopping track (race condition)


dennisguse:
### App information 
* ID: de.dennisguse.opentracks 
* Version: 4874 v4.0.0 

### Device information 
* Brand: Xiaomi 
* Device: beryllium 
* Model: POCO F1 
* Id: QKQ1.190828.002 
* Product: beryllium 

### Firmware 
* SDK: 29 
* Release: 10 
* Incremental: V12.0.3.0.QEJMIXM 

### Cause of error 
```java 
Exception in thread "main": java.lang.NullPointerException: Attempt to invoke virtual method 'void de.dennisguse.opentracks.services.TrackRecordingServiceNotificationManager.updateContent(java.lang.String)' on a null object reference 
  at de.dennisguse.opentracks.services.TrackRecordingService.newGpsStatus(TrackRecordingService.java:304) 
  at de.dennisguse.opentracks.services.handlers.TrackPointCreator.sendGpsStatus(TrackPointCreator.java:217) 
  at de.dennisguse.opentracks.services.handlers.GPSHandler.onGpsStatusChanged(GPSHandler.java:170) 
  at de.dennisguse.opentracks.services.handlers.GpsStatus.sendStatus(GpsStatus.java:210) 
  at de.dennisguse.opentracks.services.handlers.GpsStatus.checkStatusFromLastValidLocation(GpsStatus.java:170) 
  at de.dennisguse.opentracks.services.handlers.GpsStatus.onLocationChanged(GpsStatus.java:121) 
  at de.dennisguse.opentracks.services.handlers.GpsStatus$GpsStatusRunner.run(GpsStatus.java:52) 
  at android.os.Handler.handleCallback(Handler.java:883) 
  at android.os.Handler.dispatchMessage(Handler.java:100) 
  at android.os.Looper.loop(Looper.java:224) 
  at android.app.ActivityThread.main(ActivityThread.java:7562) 
  at java.lang.reflect.Method.invoke(Native Method) 
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:539) 
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:950) 
```

PR #1406, opened 2022-10-24
Handle LocationManager callbacks with TrackRecordingService's handler.

dennisguse:
Fixes #1328.

This should fix the race condition while shutting down TrackRecordingService and getting GPS data in parallel.

PR #1407, opened 2022-10-27
Bugfix: prevent crash while stopping recording (race condition).

dennisguse:
Part of #1328


dennisguse:
This is just a workaround and will be replace by #1406 (hopefully).

PR #1412, opened 2022-11-02
Fix incorrect German translation of "Open Track"

dmarku:
Change the German translation of the "Open Track" button label from "Offene Strecke" ("open" as an adjective) to "Strecke öffnen" ("open" as a verb).

Without context, the original english phrase is ambiguous and both variants would be correct translations, but as a button label in the context of importing a track it is clear that "open" is supposed to be a verb and the phrase is an action prompt, and not an adjective to describe the track.

_On a personal note: I've been happily using OpenTracks for about two years now to track shorter routes and display tracks from a fitness tracker. It's a really useful app to me. Thanks for all the effort! :)_


dennisguse:
:rocket:

PR #1427, opened 2022-11-22
Heartrate zones

### LINKED ISSUE #1118: Heart rate zones and graph in the real of time

dennisguse:
- [x] adding preference for max heartrate.
- [x] render UI

Fixes #1118.


dennisguse:
Uses the color codes provided here: https://github.com/OpenTracksApp/OpenTracks/issues/1118#issuecomment-1032290554

Issue #1424, opened 2022-11-17
BLE: get barometric pressure from Environmental Sensing service (0x181A) // Pressure characteristic (0x2A6D)


vlszabolcs:
#[continue here:](https://github.com/OpenTracksApp/OpenTracks/discussions/1423) 
I would like to use an Arduino Nano 33 BLE Sense dev board as an external Bluetooth pressure sensor (my phone hasn't built in) to get more precise altitude data.

What is currently working:
I created a code  for the nano33 sense what can you[ find here.](https://github.com/vlszabolcs/ble_sensors) I use Environmental Sensing service (0x181A) and Pressure characteristic 
(0x2A6D, datatype: uint32). (Notice: For testing  BLE characteristic i use the Elevation, Humidity,Temperature characteristic and an unknown characteristic for setup the sea level pressure to corrigation the altitude.)
I tested in nRF Connect and looks like it working there. I checked the ble sensor pressure value another device and it is correct. I think this service and characteristic setup can use with any bluetooth device for mesauring pressure . 

The problem for me i'm not really know the java. In the sensor settings I created a new generic option to the scan the ble pressure device but after i lost in the code. So i need some help to implement this if you think it s usefull.

![Inkedphoto_2022-11-17_12-37-50](https://user-images.githubusercontent.com/47059056/202436702-5f97a2d8-8390-4662-8306-f15e40904a34.jpg)

dennisguse:
Okay - that looks like something is already working.

Would you like to fetch the elevation or the pressure data?
For practical reasons it might be easier to fetch the pressure data as in OpenTracks we have the computation of the gain/loss already in place - it is "just" fetched from the internal sensors.
Steps to be taken:
- [ ] we need some test data to implement the decoding (see BluetoothUtilsTest)
- [ ] add sensor selection to be settings (`PreferencesUtils` and `BluetoothUtils`: service, characteristics, and measurement UUID)
- [ ] add data decoding to BlueoothUtils (should be similar to the heartbeat code)
- [ ] do some magic to create use `AltitudeChange` (for BLE data: we can use SensorDataSet)
- [ ] figure out what we do if an internal pressure sensor and a BLE pressure sensor are connected.

PS/ all doable no rocket science included.
@vlszabolcs I can help you down the road ;)
Can you already compile and start OpenTracks on your phone?
If not that needed before everything: just get AndroidStudio and have some fun :)

vlszabolcs:
[...]

PR #1428, opened 2022-11-22
Ble barometric pressure sensor

vlszabolcs:
[issue #1424](https://github.com/OpenTracksApp/OpenTracks/issues/1424)

Bluetooth device can connecting and send data.

Problems in the _sensorData_ and _sensors_ folders files.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -445,6 +445,7 @@ limitations under the License.
 
     <!-- Settings public api -->
     <string name="settings_api_title">Public API</string>
+    <string name="settings_api_summary">Remote control and statistics API for other apps</string>
     <string name="settings_public_api_package_title">Package name for Public API</string>
     <string name="settings_public_api_enabled_title">Public API</string>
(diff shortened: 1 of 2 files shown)


dennisguse:
@vlszabolcs quick feedback: it seems you are to solve everything at once and get something working.
I often approach problems the same way when I need to figure out how to get things done.
.... but it makes it complicated (especially) when communicating with others as there are often a lot of changes: some are working, some aren't and it is hard to determine what the problem is as well as next steps to be taken.

I would recommend splitting up your work into different commits (also use git amend and force push):
* one commit for adding the preference
* one commit adding the decoding of the data (i.e., BluetoothUtils + tests in BluetoothUtilsTest)
* one commit that adds a data class called GainLoss and SensorData<GainLoss>
  (SensorData must have some internal state as it aggregate several measurements from the sensor over time; SensorData<Distance> is doing this as well)
  For this you could reuse AltitudeChange and PressureSensorUtils (that's actually optional)

If this is working, we can figure out how to actually get the data and actually use it.

PR #1433, opened 2022-12-04
Ble pressure sensor

vlszabolcs:
New pr for the https://github.com/OpenTracksApp/OpenTracks/issues/1424

I didn't have better idea to fix the previous branch commits. (Now the commits are separated and the code is auto-formatted)

I tried to solve: 
 `Bluetooth pressure sensor settings`: The bluetooth device can connect.
`Bluetooth pressure data parse`: Bluetooth sensor data parsed to int type. (need other solution)


src/main/java/de/dennisguse/opentracks/sensors/BluetoothUtils.java
```diff
@@ -119,6 +124,17 @@ public static Integer parseHeartRate(BluetoothGattCharacteristic characteristic)
         return null;
     }
 
+    public static Integer parsePressure(BluetoothGattCharacteristic characteristic) {
```
dennisguse, line 127: Could we introduce a data class for this?
It is not that obvious that this is 0.1 pa rather than hPA.

Something like 
`AirPressure` that just has the value and provides a get-function like `int getHPA()`?

dennisguse: Again separate commit as it makes the review way easier :)

dennisguse:
@vlszabolcs I am finishing the required refactoring within the existing code, so we could continue this PR.

dennisguse:
@vlszabolcs are you still interested in this?

PR #1072, opened 2021-12-30
Public API: enable other applications to start recordings.

dennisguse:
Fixes #671.

* [x] add preference (default disabled)
* [ ] add STOP_RECORDING
* [ ] Fix TrackRecordingActivity: if it is open, it resumes the track recording after STOP_RECORDING

Issue #331, opened 2020-07-19
Implement Android app shortcuts


dennisguse:
Since Android 8.0, some launchers support app shortcuts.
We could implement to start, pause, and stop from the app icon itself.
Or also add a marker, if a recording is ongoing.

https://developer.android.com/guide/topics/ui/shortcuts

Similar to #279.

dennisguse:
Will not work nicely with something like #1171.
Also we have the Public API, so people can script whatever they want.

PR #1443, opened 2023-01-04
Feat: Add Android shortcut to start track recording

qcasey:
**Describe the pull request**
Adds Android launcher shortcut to _Start recording_. This uses the Public API intents from #1072.

Screenshot_20230104-092605~2

**Link to the the issue**
#331, I wanted to revisit this and check if shortcuts are still unwanted. Please close this PR if so

**License agreement**

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -96,6 +96,9 @@ limitations under the License.
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <meta-data
+                android:name="android.app.shortcuts"
+                android:resource="@xml/shortcuts" />
         </activity>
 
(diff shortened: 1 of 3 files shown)


dennisguse:
Nice that you implemented this!

Two things:
1. I am not sure about the use of the Public API as it is disabled by default.
So, the shortcut will be shown, but not work until the Public API gets enabled.
And this might be confusing for the user.

2. Also, the shortcut does not reflect the current recording state and does not allow to stop the recording.

In general, it feels like that for the shortcut an internal API should be created within OpenTracks.

qcasey:
Agreed, on my own device I added "Start walk" and "Start cycling" which add the relevant categories. Overall this would work better as dynamic shortcuts, with an understanding of recording state and most used activity types.

dennisguse:
@qcasey do you wanna work on that? 
If so, just reopen the closed issue

Issue #710, opened 2021-04-14
Allow App Backup


sebastianha:
I just played around with seedvault as backup and it would not make a backup of open tracks and it shows: "App doesn't allow backup".

Is this intentional? I assume that it is not properly configured, see here: https://github.com/seedvault-app/seedvault/issues/81

dennisguse:
Yes, it is disabled on purpose: https://github.com/OpenTracksApp/OpenTracks/issues/346

If you want to enable this, you can build OpenTracks yourself.

sebastianha:
Perhaps it would make sense to enable this for fdroid? Because this is also used by seedvault which is the open source backup solution built in into lineage os.

dennisguse:
Sadly that would be an anti-feature (IMHO) as this allowBackup will (also) enable backups to Google on phones with Google Play. That sucks hard and therefore it is disabled in OpenTracks.
When Google changes this, we can enable it. Some more infos are in the Android documentation: 

https://developer.android.com/guide/topics/data/autobackup

sebastianha:
Ok, thanks for the explanation!
That's really sad :(

sebastianha:
Sorry for still annoy you with this:

Wouldn't it be an option to allow backups but only when they are client side encrypted? Then also google should not have access even it gets backed up accidently.

https://github.com/seedvault-app/seedvault/wiki/FAQ#why-do-some-apps-not-allow-to-get-backed-up
https://developer.android.com/guide/topics/data/autobackup#define-device-conditions

dennisguse:
@sebastianha a PR is welcome; if you want to work on this, just re-open this issue.
Back then I just chose the safest/simplest option.
I would just like to be sure that enabling backups is not limiting privacy.

sebastianha:
I am sorry but my last app development was long time ago. Perhaps at some time in the future I will again get into it. I would be more than happy to create a PR.

dennisguse:
@sebastianha in this case it is more digging through the documentation.
The rest is just configuration.

mamoit:
@sebastianha
FYI: backups are now a thing :slightly_smiling_face:

sebastianha:
I just read the changelog, THANKS!

Issue #346, opened 2020-08-01
Allow backup


pstorch:
**Is your feature request related to a problem? Please describe.**
I cannot perform an ADB backup of OpenTracks by myself, because the Manifest doesn't allow it.

**Describe the solution you'd like**
Allow backup, set 
`android:allowBackup="true"` in the manifest.

dennisguse:
This flag disables backup to Google Cloud (on Google-enabled phones).
I am not eager to send backups there.

https://developer.android.com/guide/topics/data/autobackup

dennisguse:
And we should add tools:replace="android:allowBackup" to enforce this for potential libraries that provide a AndroidManifest.xml

pstorch:
Ok, I understand. 
Can the cloud backup and the ADB backup be separated somehow? Because this flag also hinders me as a user to do a local backup on my own.

dennisguse:
I guess not - either you allow backups (incl. adb and Google Cloud) or not.
adb and Google Cloud Backup seem to use the same stuff on the device.

@pstorch Why would you like to backup the data via adb?
Is something missing in the KMZ export/import?
Or do we miss something else?

pstorch:
I tried https://codeberg.org/izzy/Adebar to do a complete Backup of my phone. But it gave me a list of apps which doesn't allow backups. So I came to open this ticket.
Yes, export via kmz works, but would be an extra manual step.

There are settings in the manifest to require client encryption of the backup or only allow device to device backup https://developer.android.com/guide/topics/data/autobackup#define-device-conditions . Would that be an option?

dennisguse:
It is API29 - but we can give it a try.
Should we create an issue for this?

pstorch:
> It is API29 - but we can give it a try.

API 28, but still very recent. When I read it correctly that means that there won't be any backup for older devices because they don't know about the `clientSideEncryption` requirement. I hope older devices really disallow backups if they can't fulfill this requirment.

> Should we create an issue for this?

If you don't want to tacke it here, I'm fine with a separate issue.

dennisguse:
We can also tackle it here ;)

dennisguse:
@pstorch Did you try `clientSideEncryption`?

pstorch:
No, haven't looked into it yet.

pstorch:
I learned in the past days how important trust is and now I think we should close the android backup way and concentrate on https://github.com/OpenTracksApp/OpenTracks/pull/383

[...]

PR #1495, opened 2023-03-01
Enable backups if client side encryption is enabled

mamoit:
TL;DR: This PR enables the native android backup feature **only** if the backup is encrypted on the client side.

I was setting up [seedvault](https://github.com/seedvault-app/seedvault) on LineageOS when I realized that backups were not a thing on my device for OpenTracks only.
From my brief issue searching I found 2 issues, #346 and subsequently #710 I understand that this is a touchy subject.

Even if I don't completely align with the rational behind the decision to not implement this straight out of the box, I respect it.

So as @dennisguse [replied in the latter issue](https://github.com/OpenTracksApp/OpenTracks/issues/710#issuecomment-822656039), here goes a PR to try to reach a solution to which everyone can be content with.

Followed the [android documentation](https://developer.android.com/guide/topics/data/autobackup#define-device-conditions).

It has been over 10 years since I last did any Android development and I don't have an environment set up, this should be a pretty simple change (famous last words) so I'll test the build output from the CI :disappointed:
In the meantime I'll keep this marked as draft.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -62,7 +62,9 @@ limitations under the License.
 
     <application
         android:name=".Startup"
-        android:allowBackup="false"
+        android:allowBackup="true"
+        android:fullBackupContent="@xml/backup_rules_30_lower"
+        android:dataExtractionRules="@xml/backup_rules_31_higher"
         android:hardwareAccelerated="true"
         android:hasFragileUserData="false"
         android:icon="@drawable/ic_launcher"
--- a/src/main/res/xml/backup_rules_30_lower.xml
+++ b/src/main/res/xml/backup_rules_30_lower.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <include domain="sharedpref" path="." requireFlags="clientSideEncryption"/>
+    <include domain="database" path="." requireFlags="clientSideEncryption"/>
+    <include domain="external" path="." requireFlags="clientSideEncryption"/>
+</full-backup-content>
--- a/src/main/res/xml/backup_rules_31_higher.xml
+++ b/src/main/res/xml/backup_rules_31_higher.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<data-extraction-rules>
+    <cloud-backup disableIfNoEncryptionCapabilities="true">
+        <include domain="sharedpref" path="."/>
+        <include domain="database" path="."/>
+        <include domain="external" path="."/>
+    </cloud-backup>
+    <device-transfer>
+        <include domain="sharedpref" path="."/>
+        <include domain="database" path="."/>
+        <include domain="external" path="."/>
+    </device-transfer>
+</data-extraction-rules>


mamoit:
Ok, so the CI only runs the `apk-build ` step on `main` :facepalm: 
Well.. time to setup and relearn how to use the android studio.

mamoit:
Config backup and restore seem to work, but not the tracks themselves.
They're probably stored on a different path.

dennisguse:
Tracks are stored in a Sqlite database. 
Path should be defined in DatabaseHelper IIRC.

mamoit:
Thanks @dennisguse, that was all the help I needed :slightly_smiling_face:
I only had included `sharedpref` (SharedPreferences), and skipped the `database`.
Now a backup, uninstall (with data deletion), and subsequent install also restores the tracks.

Documentation is [here](https://developer.android.com/guide/topics/data/autobackup#XMLSyntax).
The full list of options is `file`, `database`, `sharedpref`, `external` and `root`.
I don't know if something else should be included besides `database` and `sharedpref`.

The `sharedpref` should not need to have the `requireFlags="clientSideEncryption"` since they don't really expose the tracks, but I'll keep it unless you want to relax this option.

There seems to be some indication in the doc where this should only be used [if targetting android 11 or lower](https://developer.android.com/guide/topics/data/autobackup#include-exclude-android-11), and an extra xml should be used [if targetting android 12 or higher](https://developer.android.com/guide/topics/data/autobackup#include-exclude-android-12).

I'm guessing that the configuration for android >=12 should also be included but I'd rather get a more insightful opinion before changing what seems  to work on my android 12.

dennisguse:
@mamoit looks great (still had no time to re-read the docs).
One thing: can you also test this with a track that contains a marker with a picture?
The pictures are stored as files in the private storage, so those might not yet be included.

PS/ we can enable this functionality as there seems to be some need.
All the arguments are still valid, but I have relaxed my view ;)

mamoit:
Indeed the pictures on markers are not kept in the backup.
I'll test the inclusion of the `domain="file"` after work (GMT).

mamoit:
Tested the `file` domain, but after looking into the `file_paths.xml` I figured the pictures actually are stored as external files.

dennisguse:
Cool!

Then the only thing missing is the Android 12+ configs, right?

@rgmf @pstorch I more or less changed my mind; having automated backups would be great. And if people use Google's cloud backup feature, it is their decision.
Does this make sense?

pstorch:
@dennisguse if the backup is encrypted, I'm fine with it.

dennisguse:
@mamoit ping :)

mamoit:
I'm sorry, got caught up on life.
I will most likely not be able to update the PR today, will do it tomorrow the latest.

I won't have a way to test it as thouroughly though since the backup already seems to work as is 😐

mamoit:
Ok, I was able to do the changes still today.
Tested it again on my Android 12 and backup/restore still works (including marker pictures).

Don't really know how to perform any further tests, this is the implementation to the best of my abilities and documentation interpretation skills.
Feel free to suggest better XML file naming :persevere:

dennisguse:
@mamoit we could use Androids API level for naming :)

dennisguse:
Looks good!
Will be in tomorrow's nightly.

I will create a new release in the upcoming days.

mamoit:
@dennisguse is it normal for the fdroid release to take considerably longer than the Google play store to be shipped?
I can still only find the previous version there.

dennisguse:
Yes and varies a lot... longest was 10 days - there is a crawler checking repos, then compilation and finally signing (needs a human) on an airgapped computer. 

That's why I use the nightly myself ;)

PR #1514, opened 2023-04-04
Bugfix: GPS disabled warning was showing (repeatedly but incorrectly).

dennisguse:
Fixes #142924.

Hopefully this fixing the incorrectly popping up of the "GPS disabled"-snackbar.


rgmf (approved):
It seems grea and it works perfectly. Also the code is better to understand;)

MarkWieczorek:
I am having the same problem with version 4.3.4.

"GPS disabled: OpenTracks is recording without GPS data" keeps appearing on the screen, even though the GPS is correctly enabled and logging.

Android 10
/e/OS 1.10 custom rom

Issue #1527, opened 2023-04-14
UI shows incorrect language


nthtimes:
UI shows incorrect language in some cases.

Language settings:
- zh-TW: It shows the en language
- zh-HK: It shows the zh language

**Technical information**
 - Device: ASUS Zenfone 6 (2019)
 - OS: OmniROM 13
 - OpenTracks version: v4.3.0, v4.3.1

dennisguse:
IIRC that is the default behavior for missing translations in the regional specific locales. 
First check zh-tw, then zh and then fallback to default  (on Android this is en-us).

nthtimes:
I note an untranslatable string named settings_layout_reset_key was translated in values-zh-rTW/strings.xml.
Is this the reason?

dennisguse:
Actually, those should not be translated as they are purely technical. I will remove these translations.

dennisguse:
Since you are on Android 13, how do you set the language?

PS/ `settings_layout_reset_key` is now in the correct file and cannot be translated anymore.

nthtimes:
To set the language as zh-TW:
1. Going into "App info\Language" on OpenTracks.
2. Choose "中文 (台灣)", that is "Chinese (Taiwan)"

nthtimes:
This issue started in v4.3.0.
In the old version 4.2.2, I got the correct language that same the system setting.

dennisguse:
@elastic-rock this seems to be an effect of #1468.
Any ideas?

dennisguse:
@elastic-rock ping

elastic-rock:
I confirm that I am alive and know about it, but I unfortunately had not the time to do it yet. I shall do it in a week max.

elastic-rock:
https://github.com/OpenTracksApp/OpenTracks/pull/1540#issuecomment-1528784034

dennisguse:
@nthtimes could you check that this fixing your issue?
Nightly is already available and F-Droid might take some days as usual.

nthtimes:
It is correctly shows my language now.
But I found that all the regional option are getting the same translation, the Traditional Chinese (Hong Kong) seems to show zh-TW translation.

dennisguse:
So, now it gets dirty.

I did some digging into when things started going wrong.

I compiled the APK for different versions and checked the included locales `resources.asc/strings`
* v4.3.4 does include 44 locales (no country-specific locales)
* v4.3.0 does include 44 locales (no country-specific locales)
[...]

PR #1540, opened 2023-04-18
Build: remove resConfigs limiting included languages.

dennisguse:
@elastic-rock was `resConfigs` required for Android 13's per app-language setting?

In 7dcc0ec8, I changed the behavior to use AGP  to generate the configuration.
I presume that thus resConfigs should not be necessary anymore.

Would you have time to test this?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,8 +74,6 @@ android {
 
         testInstrumentationRunner "de.dennisguse.opentracks.TestRunner"
         testInstrumentationRunnerArguments clearPackageData: 'true'
-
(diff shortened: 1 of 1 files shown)


elastic-rock:
I will have time to test it during the weekend.

elastic-rock:
So,according to my understanding of https://developer.android.com/guide/topics/resources/app-languages#gradle-config, it seems to me that resConfig is recommended, as it is preventing translated strings from being included from other libraries that might support languages that the app does not support, but is not required. My conclusion is that I was wrong in my testing when I created the original pull request, where I did not include the regional options in resConfig because I couldn't get them to work as said [here](https://github.com/OpenTracksApp/OpenTracks/pull/1468#issuecomment-1416863163). This means that the different regional options were blocked by resConfig resulting in #1527. I have tested it without resConfig and it works fine, fixing the issue encountered (please check it after me, because I don't want to cause another issue).

dennisguse:
@elastic-rock thanks for testing it ;)
And don't worry about making a mistake (or just having a side effect) - it is kind of normal.
If we would know what we are doing, software would always work.
I broke OpenTracks (incl. releases) personally multiple times: sometimes due to lack of testing, sometimes due to lack of knowledge, or even because I was doing something at the wrong time of the day.
Things like this happen, because it is an open-source project where people contribute and not a professional product, right?
And as long as we learn something all is good ;)

Issue #1557, opened 2023-05-10
Occasional huge speed tracking bicycle with bluetooth speedometer


rjo12:
On occasion, when stopping riding for a short period (dismounted to safely cross a road with heavy traffic) the recorded speed in that time period is huge.

In one short bicycle ride yesterday my recorded statistics show I reached a maximum speed of 2394.0 km/hr

Uncertain how to reliably reproduce during a ride, and I suspect relatively uncommon edge case(s).

However, I believe I've managed to track down a contributor, so suggest a FIX.

Relevant recorded speeds from KMZ file (two points before and after)

<value>2.9</value>
<value>0.1</value>
<value>665</value>
<value>665</value>
<value>2.5</value>
<value>2.5</value>

The 665 value is consistent with incorrect handling of a 16-bit unsigned underflow (and then dividing that incorrectly underflowed result by about 1000).   That gave me enough information to track down (what I believe is) a contributor in OpenTracks/src/main/java/de/dennisguse/opentracks/sensors/UintUtils.java (which is used in computations in sensors/SensorDataCyclingSpeed.java and for other sensors)

In the function diff() the last line is

      return (UINT_MAX - b) + a;

and should be

      return (UINT_MAX - b + 1) + a;

Rationale:   The function diff() is emulating an unsigned subtraction of a - b, which works modulo UINT_MAX + 1, not UINT_MAX  (so, for example, subtracting 1 from 0 should produce UINT_MAX, not UINT_MAX-1).

**Technical information**
 - Device: [Motorola G9 Play]
 - OS: [Stock.   Build RPXS31.Q2-58-17-7-3   (latest update currently available)
 - OpenTracks version: v4.3.4  (latest currently available through F-Droid)

dennisguse:
Thanks!

Can you create a PR for this incl. a test?

PR #1559, opened 2023-05-11
Handling of unsigned values

rjo12:
**Describe the pull request**

(1) Correction of bug related to calculating difference of "unsigned" values.   Essentially an "off by one" error.     If an unsigned variable has range 0 to UINT_MAX,  numeric operations on those variables are modulo UINT_MAX + 1 (e.g. subtracting 1 from 0 produces UINT_MAX).    The bug was doing operations modulo UINT_MAX, not modulo UINT_MAX + 1 when subtracting a large value from a small one.

(2)   Correction of test case that allowed preceding bug (1) to pass!

(3)    When calculating UintUtils.diff(a, b, UINT_MAX), no longer throw exceptions if a or b are out of range [0, UINT_MAX].   Instead, correct the values a and b using modulo arithmetic, and then calculate their difference.   This is "normal" handling of numeric operations unsigned integral types, both in hardware and in other programming languages. 

(4)    Add test cases to verify behaviour of code added in (3)

(5)   Existing test cases were only for "16 bit" unsigned subtraction.   "32 bit" tests added in interests of completeness (verify rather than assume that passing 16-bit tests mean that 32-bit tests will pass).

Edits in this pull request MAY address some contributors to this and other bugs related to spikes in sensor data (e.g. my experience of a recorded speed of 2394.0 km/hr on a bicycle, despite belief I did not achieve such a speed :-) ).    The nature of the bug is that it can map a small value to a large value, and vice versa.

**Link to the issue**

Bug (1 above) with handling calculation of difference of unsigned values is mentioned in https://github.com/OpenTracksApp/OpenTracks/issues/1557

**License agreement**

Acknowledged.

Refrained!

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/sensors/UintUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/sensors/UintUtils.java
@@ -15,6 +15,13 @@ public class UintUtils {
      */
     public static long diff(long a, long b, final long UINT_MAX) {
         
+        if (a < 0 || b < 0) {
+            throw new RuntimeException("a or b cannot be less than zero.");
+        }
+        if (a > UINT_MAX || b > UINT_MAX) {
+            throw new RuntimeException("a or b are outside of the allowed range.");
+        }
+        
         a %= (UINT_MAX + 1);
         if (a < 0) {
             a += (UINT_MAX + 1);
--- a/src/main/java/de/dennisguse/opentracks/sensors/UintUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/sensors/UintUtils.java
@@ -14,17 +14,20 @@ public class UintUtils {
      * @return diff
      */
     public static long diff(long a, long b, final long UINT_MAX) {
-        if (a < 0 || b < 0) {
-            throw new RuntimeException("a or b cannot be less than zero.");
+        
+        a %= (UINT_MAX + 1);
+        if (a < 0) {
+            a += (UINT_MAX + 1);
         }
-        if (a > UINT_MAX || b > UINT_MAX) {
-            throw new RuntimeException("a or b are outside of the allowed range.");
+        b %= (UINT_MAX + 1);
+        if (b < 0){
+            b += (UINT_MAX + 1);
         }
 
         if (a >= b) {
             return a - b;
         }
 
-        return (UINT_MAX - b) + a;
+        return (UINT_MAX + 1 - b) + a;
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/de/dennisguse/opentracks/sensors/UintUtils.java
```diff
@@ -14,17 +14,20 @@ private UintUtils() {
      * @return diff
      */
     public static long diff(long a, long b, final long UINT_MAX) {
-        if (a < 0 || b < 0) {
-            throw new RuntimeException("a or b cannot be less than zero.");
+        
+        a %= (UINT_MAX + 1);
```
dennisguse, line 18: Crashing on overflow is intended.
This will show if something else in the whole Bluetooth parsing is broken: crash early.
Otherwise a problem might be hidden for ages.

rjo12:
OK.   I've added the "crash on overflow" back in, and modified test cases accordingly (wrapping tests of values out of range in an "if (false)" block, since I suggest they are appropriate in future).

I haven't removed the corrections to values of a and b, since they have no effect if a and b are in range (effectively a no-op). 

I've done it that way because, in the long run, I suggest it would be better to catch parsing errors by constructing test cases for the parsing code, and not crashing out of sensor-related code.  [Although I accept that would be a challenging job,  I also suggest it's not really the role of sensors to detect parsing errors].

Crashing out here also means crashing out if the parser is working correctly (i.e. extracting a value that the sensor sent) but has been sent data out of range by a buggy sensor.    In that case, the appropriate action is to correct a and b (which is what the sensor should have done) and not crash out.   Handling a buggy sensor is a job appropriate for sensor-related code rather than the parser, and would also avoid inflicting crashes "in the wild"  on users unlucky enough to have bought such a sensor when we can easily deal with the problem.

dennisguse:
One final thing: can you squash the commits and set a useful commit message.
Something like: 
`Bugfix: UINT parsing overflow of by one.
Fixes #1557`

And can you remove the code handling the overflow?
Keeping the test cases is fine!

PS/ I would love to push a new release tomorrow and this fix must be definitely in it!

rjo12:
OK;  done!

I've started looking at the code that is extracting the values from bluetooth for each sensor.   Although not for today, I am concerned that doing the overflow check in the diff function and crashing out there may be too late in some cases  (essentially, an overflowed value is only picked up if calculating its difference from another value).

dennisguse:
One final thing: can you squash the commits into one (and also drop the merge commit)?

rjo12:
I'm at a loss.

I've been doing all this through a browser on github.com.     I've searched, and can't find any mechanism to squash the commits, or drop the merge commit (which I hadn't realised I had done).

Instructions I find to squash commits on docs.github.com describe a process using Github Desktop, but I haven't managed to find anything even remotely comparable via the github.com site.

I'm probably missing something blindingly obvious, but I've just lost a couple of hours going round in circles trying to work out how.

dennisguse:
Then leave it as it is.

For squashing you have to use a local checkout if the repo: https://www.baeldung.com/ops/git-squash-commits

We can also squash while merging via GitHub (just not that cool).

dennisguse:
Manually merged: 7208d2a346604a96f5ac68f33674517f16960b2b

Issue #1563, opened 2023-05-13
App Shortcuts for OpenTracks


christophrrb:
**Is your feature request related to a problem? Please describe.**
No problem. Just that OpenTracks has no app shortcuts.

**Describe the solution you'd like**
When the app icon is pressed and held, app shortcuts to start recording and to view statistics would show.

**Describe alternatives you've considered**
Considered using Tasker but thought that an app shortcut made more sense.

**Additional context**
I've created the code for it [in a fork here](https://github.com/christophrrb/OpenTracks) (and it least is works in  debug; haven't tried it for the other release versions), but between 8 and 9 of the tests fail when I run them. Some only fail when all the tests are run (deleteTrack and editTrackRecording for example) while exportImport tests have failed every time.

dennisguse:
Can you open a PR for this?

And which tests are failing?

PS/ AndroidStudio beta is required due to AGP 8.1.0

dennisguse:
And do you also want to implement pause/stop?

christophrrb:
Just created a pull request here https://github.com/OpenTracksApp/OpenTracks/pull/1564, and sure, I'll try to implement pause/stop in about a few hours.

I'm downloading Android Studio Beta now to see if that helps me.

christophrrb:
Updated the pull request. I think I have some files that need to be deleted in it (like the build.gradle file).

I was able to get an emulator up in AS Beta, and through the emulator all the tests passed except `testBuildUniqueFileName_exist()`.

I added the stop functionality, but since I didn't see a "start/stop" string, it still just says "Record". If there is no recording, it starts one, and if one is ongoing, it ends it.

dennisguse:
Neat! If you need stop as a translation just add the key :)

christophrrb:
Thanks!

And sorry, what do you mean by "add the key"? I wanted to just combine the words "Record" and "Stop", but in XML, I don't know how to concatenate string resources (if that can even be done). :/

dennisguse:
Now I understood it. 
The shortcuts you added are static.
This will work nicely for the agregated statistics. 

For the record start/stop, you are using the Public API to trigger the action. 
This has two downsides:
* The Public  API is disabled by default: this might be misleading to a new user (and it is a security issue; not a big one)
* The shortcuts do not show the actual status (aka recording vs not recording) and this is likely misleading - especially if no notification is required anymore (Android 13)

[...]

PR #1564, opened 2023-05-14
App Shortcut for Aggregated Statistics

christophrrb:
**Describe the pull request**
Added app shortcuts for ~starting an activity and~ opening the aggregated statistics activity. When the app icon is pressed and held, the shortcut is displayed.

**Link to the the issue**
https://github.com/OpenTracksApp/OpenTracks/issues/1563

tempFileForShare_20230513-233335


dennisguse:
Cool!

One thing: do we really need the configuration for each build flavor?

christophrrb:
Yeah, it's a bit clunky, but without a thrid party library (or something like that) there's not an easier way to do it. :(

That's because Android's app shortcuts work through an XML file, and they need the applicationId. In OpenTrack's case, the applicationId is specified in build.gradle and gets a suffix depending on the build version. Without the correct suffix for the build version, it doesn't work. :/

dennisguse:
Now I understood how you did it.
You are using static shortcuts (hence configured via xml).

For the aggregated statistics, that works pretty nicely!
However, for the record start/stop this will likely not work that well:
* it is using the Public API (disabled by default) and thus might be misleading (enabling it is a slight security risk)
* the shortcut doesn't show the actual recording status - so the user has to guess which one is the correct one.

I guess, dynamic shortcuts are the way to go: https://developer.android.com/develop/ui/views/launch/shortcuts/creating-shortcuts#dynamic
... but sadly they need some kind of implementation.

What about the following: we modify this PR to only add the aggregated statistics shortcut and move the other shortcuts into a separate PR?

PS/ the preferences also use Intents using the applicationId without specifying it for each build flavor.
Would this be working?
`
        <Preference
            android:key="@string/settings_import"
            android:title="@string/menu_import_all">
            <intent
                android:targetClass="de.dennisguse.opentracks.io.file.importer.DirectoryChooserActivity$ImportDirectoryChooserActivity"
                android:targetPackage="@string/applicationId" />
        </Preference>
`

christophrrb:
Good point. I forgot to mention that it was through the public API because I didn't know how to just start a recording with the app's functions, but I can look at some of the existing code and I should be able to figure it out. I'm going to try to see if the applicationId works with the way the preferences do it too.

christophrrb:
And I just remembered why the string resources method for the applicationId won't work. In the [app shortcuts documentation](https://developer.android.com/develop/ui/views/launch/shortcuts/creating-shortcuts#inner-elements), it says "Note: This intent element cannot include string resources.", and the ID is specified in that intent tag. :(

image

dennisguse:
I just tested it `@string/applicationId` doesn't work in shortcut.xml - so as documented and there seems to be no gradle task who generates the files during build :/

Anyhow for dynamic shortcuts: you can take a look at `StartRecording` and `StopRecording`.
There is a component called `TrackRecordingService` (a foreground services) that handles all the recording related things (e.g., sensors etc).

One can start/connect to it via `TrackRecordingServiceConnection` and it provides `TrackRecordingService.recordingStatus`.
The `TrackRecordingService` will only be running when a sensors are in use (might be before recording already: warm-up of GPS).

christophrrb:
I'm cleaning up this commit, but it's going to take me a day or two. The app shortcut functionality still works, but I'm just realizing how many Gradle files I've accidentally steamrolled in the process, so I'm going back and making those to match the ones in the repository.

dennisguse:
@christophrrb there is no need to keep the original commits in place; you can just squash them together. Usually it is then easier to remove unwanted changes.

christophrrb:
That's a good idea. I think it's done now. Apologies that this took so long. I got swamped in other things.

christophrrb:
Or let me know if there's anyhting else that needs to be looked addressed.

dennisguse (changes requested):
I guess, only the XML files should be in this PR. 
All other files seem to contain whitespace changes. 

What you could do:
* Remove all your commits (7 is assumed): git reset HEAD~7
* XML files only: git add ...
* git commit
* git push -f

dennisguse:
> That's a good idea. I think it's done now. Apologies that this took so long. I got swamped in other things.

PS no worries - I know the situation. 
It will take as long as it takes, right?

dennisguse:
Manually merged via https://github.com/OpenTracksApp/OpenTracks/commit/3ca83815047986ceb20fd1dd99afb0a6db8566c2

PR #1575, opened 2023-05-31
Statistics: distance, total time, moving time are counted between aut…

### LINKED ISSUE #1566: Total distance does not include segments with no GPS data
**Describe the bug**
In comparing the total distance of a track generated by OpenTracks and google maps, OpenTracks provided a number that was considerably lower. Looking at the track, it appears that the distance driving in tunnels was not counted. Many GPS tracking apps simply connect the points with a straight line when there are gaps in GPS data, but OpenTracks does not appear to do this: it instead produces a segmented track that is not continuous.

In my opinion, ignoring the segment with no GPS data is probably the worst option to use, but I understand why someone might want to do this. A solution would be to add a preference to the app that lets the user decide what to do in this circumstance.

**To Reproduce**
Drive in a tunnel with no GPS, and then look at the track.

**Technical information**
 - Device: /e/OS 1.10
 - OS: Adnroid 10
 - OpenTracks version: v4.3.4

dennisguse:
…omatic segments.

Fixes #1566.

TODO:
- [x] make sure that this is working
- [x] make sure this is a good idea
- [x] fix tests

### DIFF
--- a/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
+++ b/src/androidTest/java/de/dennisguse/opentracks/io/file/importer/ExportImportTest.java
@@ -201,7 +201,7 @@ public class ExportImportTest {
         // given
         Track track = contentProviderUtils.getTrack(trackId);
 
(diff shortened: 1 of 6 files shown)


dennisguse:
@MarkWieczorek 
does this make sense for you?

PR #1640, opened 2023-07-12
Fix aggregated stats don't show imperial units for avg pace

### LINKED ISSUE #1634: Aggregated stats shows km when user selects imperial units
Should display minutes per mile not min per kilometer 

**To Reproduce**
1. In settings set the preferred units to imperial
2. Set the preferred activity to running or street running and preferred rate to "pace (min/mi)"
3. Click on stats icon on lower left.

See screen shot.  

Oddly, if the activity type is "unknown" for all the aggregated workouts, the correct units are displayed

**Technical information**
 - Device: Google pix re on Android 13
 - OpenTracks version: 4.6.1
 
Screenshot_20230708-155351

khao-soi:
fixes #1634

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/ui/aggregatedStatistics/AggregatedStatisticsAdapter.java
+++ b/src/main/java/de/dennisguse/opentracks/ui/aggregatedStatistics/AggregatedStatisticsAdapter.java
@@ -143,9 +143,9 @@ public class AggregatedStatisticsAdapter extends BaseAdapter {
         }
 
         public void setPace(AggregatedStatistics.AggregatedStatistic aggregatedStatistic) {
-            SpeedFormatter formatter = SpeedFormatter.Builder().setUnit(unitSystem).setReportSpeedOrPace(reportSpeed).build(context);
-
             setCommonValues(aggregatedStatistic);
+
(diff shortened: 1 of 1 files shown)


dennisguse (approved):
Looks good. 

Only thing: can you add the "Fixes #1684." into the commit message (2nd line)? 
This way the commit gets linked to the issues; useful for future engineers ;)

khao-soi:
Done. I screwed up with android studio and created a second commit instead of ammending it. Hope thats ok, otherwise I will create a new PR.

dennisguse:
You can just squash the commits and then force push. No need to create a new PR ;)

khao-soi:
Ehm, it says I can't because the inital commit has 2 parents.

dennisguse:
You have to remove the merge commit and then it works.

PS on branches it is way easier to rebase + force push to get updates from the parent branch. 

Merging is usually complicating things :)

khao-soi:
Yeah, I wanted to rebase like usual via CLI but somehow I thought it'd be a good idea to deal with it via Android Studio and ended up with this mess.

Anyways, its fixed now.


Bei sowas wird git dann echt unübersichtlich...

Issue #1426, opened 2022-11-22
V4.2.+: Re-implement StatisticsRecordedFragment's sensor statistics UI


dennisguse:
**Is your feature request related to a problem? Please describe.**
While refactoring #1386, I removed the SensorStatistics (e.g., max heartrate) from the StatisticsRecordedFragment.
It was a side effect of reducing complexity to add more custom UIs to the StatisticsRecordingFragment.

**Describe the solution you'd like**
Re-implement a RecyclerView.Adapter that manages the SensorStatistics on the UI.

**Additional context**
This seems like a separate UI, so re-using existing components might not be beneficial.
Layout: `stats_generic_item.xml`

c33s:
would this also include having pace statistics (max pace, min pace, avg pace) in the recorded track detail view (currently v4.1.1 only visible while recording)?

dennisguse:
@c33s yes. 
I needed to remove this due to some refactoring - some things were tied together too much to implement the heart rate zone feature. 

And I had no time to reimplement it.
Help is appreciated.

c33s:
please provide the 4.1.2 here and on fdroid as long as this feature isn't reimplemented. currenlty it is not possible to downgrade from a trusted source :(

MrS1R1US:
[OpenTracks.v4.1.2.zip](https://github.com/OpenTracksApp/OpenTracks/files/11583712/OpenTracks.v4.1.2.zip) I use this version because I need heart rate data for comparison (

c33s:
> [OpenTracks.v4.1.2.zip](https://github.com/OpenTracksApp/OpenTracks/files/11583712/OpenTracks.v4.1.2.zip) I use this version because I need heart rate data for comparison (

thank you very much. as far as i know this is the playstore version. i try to stick by the fdroid versions but it is gone from fdroid:

https://f-droid.org/en/packages/de.dennisguse.opentracks/

![image](https://github.com/OpenTracksApp/OpenTracks/assets/649209/ed9d452c-c99b-4b62-bafc-adc9b13756b1)

i assume it is because of the fact that no direct downgrade is possible via fdroid.

for the records 
- i copied the 4.1.2 apk from a 2nd phone to the main phone already having installed a 4.3 version. 
- i exported the data from 4.3
- the installation of 4.1.2 failed
- uninstalled 4.3
- installed 4.1.2 (worked)
- imported the data 
- i got back the average and the max heart rates even from tracks i recorded with 4.3

[...]

PR #1646, opened 2023-07-27
StatisticsRecordedFragment: Reimplement Sensor Statistics

lmarz:
Closes: #1426

This adds the sensor stats back to the StatisticsRecordedFragment by adding a new class SensorStatisticsAdapter for the RecyclerView and connecting everything.

I have used the `stats_recorded_item` layout, but left the `description_secondary` field empty.

The basic thought here is, that we look through each type in SensorStatistics and when it's available, we fill a struct-like class called Stats with the field values, so in onBindViewHolder(), we only have to set these values in the views.

I have put SensorStatisticsAdapter.java into the same directory as StatisticsRecordedFragment.java, but I have no problem moving it to a better spot, if you have any idea ;)

I only tested this change with Heart Rate, since it's the only external device I have.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/fragments/SensorStatisticsAdapter.java
+++ b/src/main/java/de/dennisguse/opentracks/fragments/SensorStatisticsAdapter.java
@@ -1,0 +1,114 @@
+package de.dennisguse.opentracks.fragments;
+
+import android.content.Context;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.TextView;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
+
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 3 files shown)


dennisguse:
Back then when I had to remove this functionality as it was assuming that all statistics are the same (was used for recording as well as for recorded).
All the statistics views are now instantiated directly.
This keeps the code very clean and easy to adjust (although there might be some performance limitations).
I am not sure that re-introducing the RecyclerView approach for the three sensors-based statistics alone is the best idea.
Just instantiating the views directly would be easier, right?

lmarz:
I mostly followed the suggestion you mentioned in the issue:
> Describe the solution you'd like
> Re-implement a RecyclerView.Adapter that manages the SensorStatistics on the UI.

But your right, adding the stats directly would be more consistent to the rest.

I made another pull request #1649 with the different approach. You can decide, which one to use

dennisguse:
#1649 works nicely (and is easier to maintain).

PR #1649, opened 2023-07-29
Reimplement sensor statistics in UI

lmarz:
Closes: #1426

This is an alternative version of #1646, which doesn't use the RecyclerView.


dennisguse:
Looks good and works :)
Thanks for taking the time.

Issue #1613, opened 2023-06-19
Convert ListViews to RecyclerViews.


dennisguse:
... might make some things nicer to change later one.

* [x] replace all ListViews
* [ ] remove activated_background.xml

dennisguse:
EspressoUI tests are failing due to this refactoring.
`selectAndDeleteTrack()` and `espressoDeleteTrackTest()`.
The problem seems to be that the interaction in the TrackList is not working anymore.

dennisguse:
The TrackList has a bug. 
While recording, navigating back to the list shows all tracks as recording.  
In addition, the track name is sometimes not shown fully (aka dots at the end). 
Seemingly the length of another view is used. 
Both things seem to be related.

PR #1650, opened 2023-07-29
Fix alignment of Statistics

lmarz:
This bug accidentally made it into db1a37f. Android Studio, for some reason, has made this change automatically and I didn't notice it. Sorry about that.

Without this fix, the alignment of the two columns isn't correct in landscape orientation or different screen sizes.


dennisguse:
@lmarz if you want to work on some RecyclerViews in OpenTracks - there is quite some technical debt to replace: https://github.com/OpenTracksApp/OpenTracks/issues/1613

PR #1652, opened 2023-07-31
Idle detection#1187

dennisguse:
So far, idle was considered if the speed dropped below a certain criteria.
However, speed from GPS is noisy and especially for slow moving activities (like hiking), this resulted in a lot of idle detection.

This PR changes the idle detection to "time without movement" (distance covered at least minRecordingThreshold).

TODO:
- [x] import/export KML (export TrackPoint.Type)
- [x] ~import/export GPX (speed=0.0)~ not supported (for now) due to lack of GPS coordinates
- [x] extend ExportImport test
- [x] ... check TrackStatisticsUpdater is completely implemented

For legacy formats (aka prior to this change), we may create new idle TrackPoints.

NOTE: this is a breaking change as it changes the database.
Statistics for existing tracks will not be updated (only on import).


dennisguse:
@pstorch @rgmf Any opinions about this?

pstorch:
@dennisguse let's see if I can bend my head around this issue.

dennisguse:
It's about not using the speed to determine if the user became idle, but a timeout (aka no distance traveled since...).

My biggest issue was the KML/GPX spec (again).

src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
```diff
@@ -154,16 +161,28 @@ public Marker.Id insertMarker(String name, String category, String description,
...
         if (trackPoint.hasSpeed()) {
             lastTrackPointUIWithSpeed = trackPoint;
         }
         if (trackPoint.hasAltitude()) {
             lastTrackPointUIWithAltitude = trackPoint;
         }
+
+        if (trackPoint.getType() == TrackPoint.Type.IDLE) {
+            insertTrackPoint(trackPoint, true);
+            handler.removeCallbacks(this::onIdle);
```
pstorch, line 183: Interesting concept with the handler and the `postDelay()`.
Should `this::onIdle` be `ON_IDLE`?

src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
```diff
@@ -200,18 +219,17 @@ boolean onNewTrackPoint(@NonNull TrackPoint trackPoint) {
...
+        if (distanceToLastStoredTrackPoint.greaterOrEqualThan(recordingDistanceInterval)) {
             insertTrackPoint(trackPoint, false);
-            return true;
-        }
 
-        if (trackPoint.isMoving() != lastStoredTrackPoint.isMoving()) {
-            // Moving from non-moving to moving or vice versa; required to compute moving time correctly.
-            insertTrackPoint(trackPoint, true);
+            handler.removeCallbacks(ON_IDLE);
+            handler.postDelayed(this::onIdle, idleDuration.toMillis());
```
pstorch, line 232: Same here: should `this::onIdle` be `ON_IDLE`?

dennisguse: yes

src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
dennisguse: I wasn't sure how the handler keeps track of the callbacks, so I opted for a final variable to be on the safe side.
Will push -f it now.

dennisguse: ... done

pstorch:
So far it makes sense, but I don't have a full picture (for example on possible side effects). Maybe an extended test period with the nightlies?

dennisguse:
@pstorch I also expect some fancy behavior.
Merging it into main for the nightly would basically block us from creating new releases without this functionality.
That feels kind of ugly - especially as the change is relatively large.

I am not really sure how to continue here.
Would moving the nightly build to this branch for some days/weeks be an option?
Is it possible to then switch to main again?

pstorch:
@dennisguse have switched the OpenTracks nightly branch to `idleDetection#1187`, let's see. Haven't done it before for any other app.

pstorch:
@dennisguse have you got any feedback from this PR and nightly? For me it's working. Recorded a few tracks and some pause got recognized. I can see it now easily with the OSMDashboard trackpoints debug option. 😉
I'm thinking about placing "pause" icons on the Map.

dennisguse:
@pstorch Sounds great! 
I haven't used it that much; was knocked out for some days and now I have no time :D
What I want to build before merging this is the voice announcement, which is played when idle starts.
This way it is easier to figure out what is happening.
And I was thinking about showing idle in the UI somehow.
I would love to get this done until the end of September.

pstorch:
@dennisguse I'm experimenting with showing the idle/pause trackpoints in the OSMDashboard nightly. Using the previous trackpoints location. So far it looks reasonable.

pstorch:
@dennisguse build fails after rebase

dennisguse:
androidx.mediarouter upgrade introduced dependency conflicts and the tests are broken due to recent pull request.
The first thing is fixed by downgrading.

dennisguse:
We now have voice announcement when the user becomes idle (disabled by default).
From my perspective, this is ready to go into main.

@pstorch merge or do you see any showstoppers?

pstorch:
It looks good to me. I recorded a few tracks. One time I got a wrong idle detection where I was a bit slow, but I guess I just have to increase the threshold a bit.

dennisguse:
I was just doing a walk and 10s doesn't work too well :D
That's something, the users will have to figure out...

pstorch:
I really like the idle announcement. 🙂

dennisguse:
I also enjoy it a lot. 
Gives me something to look forward to at every red traffic light ;)

PS also the pause visualization shows quite nicely all the traffic lights :)

PR #1654, opened 2023-08-05
AggregatedStatisticsActivity: Convert ListView to RecyclerView

lmarz:
Part of: #1613

Arguably, since there aren't too many `ActivityType`s a person usually tracks, this change doesn't improve performance much.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/ui/aggregatedStatistics/AggregatedStatisticsAdapter.java
+++ b/src/main/java/de/dennisguse/opentracks/ui/aggregatedStatistics/AggregatedStatisticsAdapter.java
@@ -1,14 +1,15 @@
 package de.dennisguse.opentracks.ui.aggregatedStatistics;
 
 import android.content.Context;
-import android.content.res.Resources;
 import android.util.Pair;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.BaseAdapter;
 import android.widget.ImageView;
 import android.widget.TextView;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/de/dennisguse/opentracks/ui/aggregatedStatistics/AggregatedStatisticsAdapter.java
```diff
@@ -21,64 +22,43 @@
...
+        if (ActivityType.findByLocalizedString(context, type).isShowSpeedPreferred()) {
             viewHolder.setSpeed(aggregatedStatistic);
         } else {
             viewHolder.setPace(aggregatedStatistic);
         }
-
-        return convertView;
     }
 
-    public AggregatedStatistics swapData(AggregatedStatistics aggregatedStatistics) {
```
dennisguse, line 78: Nitpicking: diff is easier to read when methods are not switched.
Switching getCategories() and swapData() shows that swapData() actually didn't change.

PS/ Could it be that your AndroidStudio as a cleanup?
Also happened for instance variables (context and aggregatedStatistics).

dennisguse (approved):
Besides the re-ordering: looks good.

dennisguse:
And you can add `Part of #1613` to the commit message (2nd line).

And that's all the nitpicking :)

src/main/java/de/dennisguse/opentracks/ui/aggregatedStatistics/AggregatedStatisticsAdapter.java
```diff
@@ -183,9 +169,7 @@ private void setCommonValues(AggregatedStatistics.AggregatedStatistic aggregated
 
         private int getIcon(AggregatedStatistics.AggregatedStatistic aggregatedStatistic) {
             String localizedActivityType = aggregatedStatistic.getActivityTypeLocalized();
-            String iconValue = ActivityType.findByLocalizedString(context, localizedActivityType)
-                    .getIconId();
-            return ActivityType.findBy(iconValue)
+            return ActivityType.findByLocalizedString(context, localizedActivityType)
```
dennisguse, line 172: Nice find :)

lmarz:
I don't really mind the nitpicking. It often improves the code :)

btw. marker list and track list might take a while longer, since there is more stuff happening with them (e.g. selection and cursors).

dennisguse:
Regarding TrackList and MarkerList: this code is ancient (aka still from MyTracks, so likely 10 years old) and was rarely touched. 
IIRC correctly it has the assumption that both share some code - probably not really helpful to keep it this way.

Issue #1653, opened 2023-08-04
GPS won't activate, when tracking starts while screen is off


v-schaefer:
I use OpenTracks in combination with Gadgedbridge and a Fossil HR watch.
Whenever I start a training on my watch (for example cycling or running) and my phone is in standby, OpenTracks doesn't activate GPS.
The message title in the statusbar stays stuck at "starting GPS" and my training will get recorded without GPS. The phone also doesn't show the symbol that an app is accessing GPS.
However if my phone is on (for example at the homescreen) and I start a new training from my watch, then GPS gets started immediately, the phone shows that GPS is getting accessed by an app and everything works fine.

All battery optimization options on Android are disabled for OpenTracks.

...

**Checklist**

- [x] Consulted [https://dontkillmyapp.com/]
- [x] Power management: battery optimization is disabled for OpenTracks

**Technical information**

- Device: OnePlus 7T Pro
- OS: Stock Android 12 ROM
- OpenTracks version: v4.6.2 (from fdroid)
- OpenTracks commit id (only for nightly builds): -

v-schaefer:
I've recorded a logcat and from what it looks like, it is no problem to create a foreground service while in standby, but the access to GPS is denied then:
[logcat.txt](https://github.com/OpenTracksApp/OpenTracks/files/12270121/logcat.txt)

To better understand what I'm doing, I've also recorded it with a second phone.
It's a bit blurry, but you can still spot what's relevant:

First Situation: Phone is on the homescreen - when the training starts, you can see the GPS-Icon on the right side of the status bar and GPS gets locked quickly:

https://github.com/OpenTracksApp/OpenTracks/assets/26538718/b80d3353-7ba2-4461-a0ff-f0c03e4a10c6


Second: The phone is in standby when the training starts. After powering on, you can see that there's no GPS-Icon in the status bar and OpenTracks is stuck on "starting GPS":

https://github.com/OpenTracksApp/OpenTracks/assets/26538718/07f76ec5-2d43-4ae3-8dc8-b08ef97974b5

dennisguse:
Location access denied.

Are you able to add (and request) the location background permission?


```
2023-08-06 19:13:10.905 10566-10566 GPSManager              de.dennisguse.opentracks             E  Not started.
2023-08-06 19:13:10.907  2595-4182  OplusLbsRepairer        system_server                        V  location access denied
```

v-schaefer:
[...]

PR #1658, opened 2023-08-06
Fix: GPS Access denied when device in standby

v-schaefer:
As discussed in https://github.com/OpenTracksApp/OpenTracks/issues/1653 , I've added the ACCESS_BACKGROUND_LOCATION and tested it successfully.
Starting to track a training while the phone is in standby now works with GPS.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/util/PermissionRequester.java
+++ b/src/main/java/de/dennisguse/opentracks/util/PermissionRequester.java
@@ -57,7 +57,7 @@ public class PermissionRequester {
         locationPermissionRequest.launch(permissions.toArray(permissions.toArray(new String[0])));
     }
 
-    private static final List<String> GPS_PERMISSION = List.of(Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_COARSE_LOCATION);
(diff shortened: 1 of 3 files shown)


dennisguse:
Thx :)

Can you move the definition of this permission to PublicApi? 

The problem is, it is only needed for background starts and that only occurs due to PublicApi - otherwise we will forget why this permission is there in the first place.

src/main/java/de/dennisguse/opentracks/util/PermissionRequester.java
```diff
@@ -57,7 +57,7 @@ private void requestPermission(ActivityResultCaller context, @Nullable Runnable
         locationPermissionRequest.launch(permissions.toArray(permissions.toArray(new String[0])));
     }
 
-    private static final List<String> GPS_PERMISSION = List.of(Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_COARSE_LOCATION);
+    private static final List<String> GPS_PERMISSION = List.of(Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION, Manifest.permission.ACCESS_COARSE_LOCATION);
```
dennisguse, line 60: Can we document that ACCESS_BACKGROUND_LOCATION is used by PublicApi.StartRecording?

v-schaefer:
As discussed both my commits are squashed into one and ready now :)

Issue #1669, opened 2023-08-19
Automatic export to nextcloud


rkozlo:
I set automatic export to my nextcloud server after finished activity. Sometimes it stops working "Can't be saved"(or something similar). Manual export tracks working. Choosing the same destination directory and next activity may be send or not. Seems not to be determinstic.
**To Reproduce**
1. Set automatic export to remote nextcloud server
2. End your activity
3. Sometimes error appear "can't be saved" or something similar
4. In settings also error. Manual export works

**Technical information**
 - Device: Motorola G8 Power
 - OS: Android 11
 - OpenTracks version: 4.7.1(had this problem for whole time using this export)

Don't have any screenshots. Can gather some logs if there is a way.

pstorch:
> Sometimes error appear "can't be saved" or something similar

Can you give more details on "or something similar"? Exact error message might be helpful.

rkozlo:
I do not have English language set. So i translate it by myself, so English version it may sound different. Don't see a way to change language in application so it uses statically system language.

pstorch:
We've added some diagnostic to analyze this issue. It's already available in the nightly and soon in the next release. 
I hope we get some more insight.

WPettersson:
Just had the issue:
```
Post-workout export failed, please check the export directory.
Cannot write to export directory.: androidx.documentfile.provider.TreeDocumentFile@cde2130
```

Note that this was a first export after upgrade, I don't know whether perhaps the app needs to be re-authorised after an upgrade.

pstorch:
Looks like no write permission. Please try to authorize the the directory again. Let's see if it helps.

WPettersson:
Have done, just have to wait for the next error now

dennisguse:
No error reported, so this might be working ;)

PR #1677, opened 2023-08-26
show postworkout error messages, #1669

pstorch:
Let's show the exact error message of the automatic post workout export to the user, so we can analyze what is happening in #1669 .

For the manual export, I'm just logging the errors at the moment. But we could also add it to the `trackErrors` List.


dennisguse (approved):

dennisguse:
@pstorch create a release with this change soonish (aka this weekend) makes sense?

PR #1687, opened 2023-09-03
TrackListActivity: Switch to RecyclerView

lmarz:
Part of: #1613

I changed the padding a bit, I hope that's ok.
Also the TrackListAdapter.java is in a suboptimal place in the ui directory. Do you have an idea for a better place?


src/main/java/de/dennisguse/opentracks/data/ContentProviderUtils.java
```diff
@@ -228,6 +228,49 @@ public List<Track> getTracks(ContentProviderSelectionInterface selection) {
         return tracks;
     }
 
+    public List<Track> searchTracks(String searchQuery) {
```
dennisguse, line 231: Should be a cursor as otherwise all tracks will be loaded (even the one not visible on the screen), right?

src/main/java/de/dennisguse/opentracks/ui/TrackListAdapter.java
```diff
@@ -0,0 +1,252 @@
...
+
+    private final AppCompatActivity context;
+    private final RecyclerView recyclerView;
+    private final SparseBooleanArray selection = new SparseBooleanArray();
+    private RecordingStatus recordingStatus;
+    private UnitSystem unitSystem;
+    private List<Track> tracks;
+    private boolean selectionMode = false;
+    private ActivityUtils.ContextualActionModeCallback actionModeCallback;
+    private androidx.appcompat.view.ActionMode actionMode;
```
dennisguse, line 46: ActionMode is imported already.

src/main/java/de/dennisguse/opentracks/ui/util/ListItemUtils.java
```diff
@@ -160,6 +67,28 @@ private static String getTimeDistance(String totalTime, String totalDistance) {
...
+        if (isRecording) {
+            timeDistanceText = context.getString(R.string.generic_recording);
+        } else {
+            // Match list_item_time_distance in list_item.xml
+
+            String time = StringUtils.formatElapsedTime(totalTime);
+            // TODO unit system
+            String distance = DistanceFormatter.Builder()
+                    .setUnit(unitSystem)
+                    .build(context).formatDistance(totalDistance);
```
dennisguse, line 81: add a newline

src/main/java/de/dennisguse/opentracks/ui/util/ListItemUtils.java
```diff
@@ -160,6 +67,28 @@ private static String getTimeDistance(String totalTime, String totalDistance) {
...
+
+            timeDistanceText = getTimeDistance(time, distance);
+            if (markerCount > 0) {
+                timeDistanceText += "  ‧";
```
dennisguse, line 85: ?

dennisguse (changes requested):
I just read the PR and made some comments.

src/main/AndroidManifest.xml
```diff
@@ -333,7 +333,8 @@ limitations under the License.
 
         <activity
             android:name=".TrackListActivity"
-            android:exported="true">
+            android:exported="true"
```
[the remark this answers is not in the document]
lmarz: This is necessary to make search work. Otherwise `onNewIntent()` will never be called. This option is there, so that Android doesn't create a new instance of TrackListActivity when a search action is performed, but reuses the currently existing one.

src/main/java/de/dennisguse/opentracks/data/ContentProviderUtils.java
lmarz: We could do it multiple ways. Either we use a `CursorLoader`, `Track.Id`s or the full list of tracks.
I did a quick memory dump and found out that one instance of Track takes up ~320 bytes of memory. So that isn't the biggest concern for me. We mainly use RecyclerView to cut down the amount of Views, which take up more memory.
But I can rewrite it to use less memory, if you want me to.

src/main/java/de/dennisguse/opentracks/ui/util/ListItemUtils.java
lmarz: I don't know either, tbh. I just copy-pasted the original code there

src/main/java/de/dennisguse/opentracks/data/ContentProviderUtils.java
dennisguse: It is not an issue for newer devices, but my good old S4 is slow and the memory is limited. Especially, for people who have quite some amount of tracks (1000+), this will help quite a lot.

lmarz: I switched to `List<Track.Id>`, where each id only takes up 16 bytes. To improve this further, we would need to write quite a lot of memory management code.

dennisguse: That's not what I meant ;)
IIRC The old implementation was using the data directly from the database (via the cursor using ResourceCursorAdapter).
So, it only loaded the data from the database that was actually shown.
The change would be that `TrackListAdapter.swapData()` is changed use the cursor created by `ContentProviderUtils.searchTracks()` (instead of using a list with the resolved TrackId/data).

lmarz: Sorry for the delay and also sorry for not researching what a Cursor actually is. Feel free to tell me, if you find other things that can be improved

dennisguse: No worries :)

src/main/java/de/dennisguse/opentracks/ui/TrackListAdapter.java
```diff
@@ -0,0 +1,291 @@
...
+        public long getId() {
+            return trackId.id();
+        }
+
+
+        @Override
+        public void onClick(View v) {
+            if (selectionMode) {
+                setSelected(!view.isActivated());
+                actionMode.invalidate();
```
dennisguse, line 265: Would recommend a return here instead of a if.. else.
Easier to read

dennisguse (approved):
Works!

One final thing: the UI is slightly off in the list.
The distance between text and the separator line below is slightly too large.

image

dennisguse:
Nice work :)

PR #1704, opened 2023-09-23
Re-enable Github Action executes tests.

dennisguse:
... we have missed to adjust some test in the recent past before merging PRs.
A flacky test execution is better than no test execution.


dennisguse:
Tests are still broken - but they are now running again.

PR #1705, opened 2023-09-24
Fix broken TrackList UI tests

lmarz:
We can just find the element by matching the first child of the RecyclerView.

Tested locally, since CI has some issues right now.

`selectAndDeleteTrack` does basically a subset of `espressoDeleteTrackTest`, so it probably can be deleted without changing the test coverage.


dennisguse:
:rocket:

Issue #1711, opened 2023-09-28
App crash in track selection mode


khao-soi:
**Describe the bug**
App crashes when long pressing a second track in selection mode

**To Reproduce**
1. Long press one track to enable selection mode
2. Long press any other track
3. App crashes

### App information
* ID: de.dennisguse.opentracks
* Version: 5399 v4.8.4

### Device information
* Brand: OnePlus
* Device: OnePlus8T
* Model: KB2003
* Id: TQ3A.230805.001
* Product: OnePlus8T

### Firmware
* SDK: 33
* Release: 13
* Incremental: eng.root.20230910.055511

### Cause of error
```java
Exception in thread "main": java.lang.NullPointerException: Attempt to invoke virtual method 'long de.dennisguse.opentracks.ui.TrackListAdapter$ViewHolder.getId()' on a null object reference
	at de.dennisguse.opentracks.ui.TrackListAdapter.getCheckedIds(TrackListAdapter.java:157)
	at de.dennisguse.opentracks.ui.TrackListAdapter.onPrepareActionMode(TrackListAdapter.java:110)
	at androidx.appcompat.app.AppCompatDelegateImpl$ActionModeCallbackWrapperV9.onPrepareActionMode(AppCompatDelegateImpl.java:3017)
	at androidx.appcompat.view.StandaloneActionMode.invalidate(StandaloneActionMode.java:100)
	at de.dennisguse.opentracks.ui.TrackListAdapter$ViewHolder.onClick(TrackListAdapter.java:266)
	at android.view.View.performClick(View.java:7542)
	at android.view.View.performClickInternal(View.java:7519)
	at android.view.View.-$$Nest$mperformClickInternal(Unknown Source:0)
	at android.view.View$PerformClick.run(View.java:29476)
	at android.os.Handler.handleCallback(Handler.java:942)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7924)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
```

virginiaprivacycoalition:
[...]

PR #1723, opened 2023-10-08
TrackList: Fix crashes in selection mode

lmarz:
Fixes: #1711

The long click crash happened, because we didn't check whether we are already in selection mode.

The context menu item crash happened in `setAllSelected`, because item count != view holder count. That was a design error of me that the SparseBooleanArray uses the view holder positions, which change their contents, as the keys. So now I switched to using the trackIds, which are stored as Integers in the database, so we can easily convert them to long and back. This also simplifies the `getSelectedIds` method. The values from `getSelectedPositions` don't get used in the TrackListActivity, so we can just use null.

In regards to the [`recyclerview-selection`](https://developer.android.com/reference/androidx/recyclerview/selection/package-summary) library you mentioned, I deliberately decided against it, because
1. It's an extra library we would need to include.
2. The API, in my opinion, is awful. The method I used with the SparseBooleanArray is much more intuitive and easier to read to me.

I won't reimplement the selection mode using this library, but if someone gets along with the API and wants to implement this, feel free to :)

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/ui/TrackListAdapter.java
+++ b/src/main/java/de/dennisguse/opentracks/ui/TrackListAdapter.java
@@ -107,13 +107,14 @@ public class TrackListAdapter extends RecyclerView.Adapter<RecyclerView.ViewHold
     @Override
     public boolean onPrepareActionMode(ActionMode mode, Menu menu) {
         selectionMode = true;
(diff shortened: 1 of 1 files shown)


dennisguse:
Works!
Thx.

@lmarz is the MarkerList also affected by this?

lmarz:
Yes, thanks for reminding me

dennisguse:
Fix is released as v4.9.1

PR #1727, opened 2023-10-08
Android12+: use user-defined DynamicColors.

dennisguse:
Finally, we got support for MaterialDesign3 DynamicColors.
Android 12+ only.

:shrug:


dennisguse:
I don't know if this is a good idea.
We may also put this behind a user setting.

PR #1766, opened 2023-11-22
fix NPE at null barometer after unregister listener

pstorch:
I experimented a little bit. I'm not sure what happens here, but I guess the internal Object Instance of the SensorEventListener doesn't "see" the null observer right away. When I implement the SensorEventListener interface directly on the BarometerInternal the NPE is fixed.


dennisguse (approved):
Fixing the fix with a fix :)

pstorch:
There are other internal Listerner instances who operate on the enclosing Objects fields. Maybe we should check them, too.

PR #1771, opened 2023-11-27
fix #1769 export overwrite existing files

### LINKED ISSUE #1769: Broken .gpx export
I exported again all my activities from OpenTracks to .gpx. Some files have errors like this, see the broken element `ension>`:

```
<trkpt lat="50.xxxxxx" lon="9.yyyyyy">
<ele>427</ele>
<time>2020-06-11T18:16:43Z</time>
<extensions><gpxtpx:TrackPointExtension>
<gpxtpx:speed>3.05</gpxtpx:speed>
</gpxtpx:TrackPointExtension></extensions>
</trkpt>
<trkpt lat="50.xxxxxx" lon="9.yyyyyy">
<ele>430</ele>
<time>2020-06-11T18:16:48Z</time>
<extensions><gpxtpx:TrackPointExtension>
<gpxtpx:speed>1.75</gpxtpx:speed>
</gpxtpx:TrackPointExtension></extensions>
</trkpt>
</trkseg>
</trk>
</gpx>
ension>
<gpxtpx:speed>1.75</gpxtpx:speed>
</gpxtpx:TrackPointExtension></extensions>
</trkpt>
</trkseg>
</trk>
</gpx>

```

OpenTracks nightly v4.9.5-15-g8b7063d18

pstorch:
Found the cause of the broken .gpx exports. 
As the tracks were duplicates (from test ex- and imports). The export all function wrote two times to the same file. The second time the content was slightly less bytes, so the first file was only partly overwritten resulting in a broken xml file.

We need to make sure the target file is properly overwritten (truncated at the beginning).

Solution:
https://stackoverflow.com/questions/56902845/how-to-properly-overwrite-content-of-file-using-android-storage-access-framework

@dennisguse do we have to do anything else if two tracks result in the same filename and overwriting each other while exporting?

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/util/ExportUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/util/ExportUtils.java
@@ -68,7 +68,7 @@ public class ExportUtils {
         }
 
         TrackExporter trackExporter = exportTask.getTrackFileFormat().createTrackExporter(context, contentProviderUtils);
-        try (OutputStream outputStream = context.getContentResolver().openOutputStream(exportDocumentFileUri)) {
+        try (OutputStream outputStream = context.getContentResolver().openOutputStream(exportDocumentFileUri, "wt")) {
             if (!trackExporter.writeTrack(tracks, outputStream)) {
                 if (!DocumentFile.fromSingleUri(context, exportDocumentFileUri).delete()) {
                     throw new RuntimeException("Unable to delete exportDocumentFile");


dennisguse:
We could also use this (same link), right?
```
contentResolver.openOutputStream(uri, "wt")?.use { outputStream ->
    // Write to outputStream
}
```

dennisguse:
It is but the "wt" should work in Java as well ;)

pstorch:
Ah, I'll try

pstorch:
Yes, the simple solution works, too. Thanks for the hint.

PR #1793, opened 2023-12-25
Fix crash due to BLE sensor data reported after shutdown

dennisguse:
Problem(s):
* TrackListActivity started sensors and recording (always).
* The TrackRecordingService checked only if a recording was running and NOT if the sensors were already started.

Follow-up:
* Shutdown of sensors may not been correct for quite a while
* Sensor data shouldn't have been affected

Moreover: TrackListActivity's start and stop sensor didn't work correctly.


Mats391:
I tested the latest nightly including this fix with my heart rate sensor and it no longer crashes. I restarted recording and the app multiple times and it worked in all tests.

PR #1817, opened 2024-01-09
do not report units in the announcer

### LINKED ISSUE #1805: do not report units in the announcer
My problem is that the announcer talk can be talking for too long and "hiding" too much of a show I'm listening in the background while training.

Do not report the units. They are useless. Sometimes they are too long like e.g. "kilometres per second". They are always the same as set in the settings.

Continue to report the name of the quantity being reported, eg. "Average speed" (or only "Speed" for interval speed average) and the number. The name needs to stay so that we do not need to remember the meaning of reported numbers by their ordinal position. It also allows me to increase concentration to catch the following number.

Unit reporting can be an option in the settings but I think it is not worth it.

hercek:
Fixes #1805

**Describe the pull request**
This solves #1805. It is tested on my phone except the heart rate stuff. I do not have heart rate monitor. I removed the translation strings which were used in power announcer and which are not used any more anywhere after this PR. There are about a hundred of other unused translation strings. I did not touch those. I needed to add some translation strings (see res/values/strings.xml). Only default English version was added for those.

**Link to the the issue**
#1805

**License agreement**

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncementUtils.java
+++ b/src/main/java/de/dennisguse/opentracks/services/announcement/VoiceAnnouncementUtils.java
@@ -42,48 +42,13 @@ class VoiceAnnouncementUtils {
         SpannableStringBuilder builder = new SpannableStringBuilder();
         Distance totalDistance = trackStatistics.getTotalDistance();
         Speed averageMovingSpeed = trackStatistics.getAverageMovingSpeed();
-        Speed currentDistancePerTime = currentInterval != null ? currentInterval.getSpeed() : null;
-
-        int perUnitStringId;
-        int distanceId;
-        int speedId;
-        String unitDistanceTTS;
-        String unitSpeedTTS;
-        switch (unitSystem) {
-            case METRIC -> {
-                perUnitStringId = R.string.voice_per_kilometer;
-                distanceId = R.string.voiceDistanceKilometersPlural;
-                speedId = R.string.voiceSpeedKilometersPerHourPlural;
-                unitDistanceTTS = "kilometer";
-                unitSpeedTTS = "kilometer per hour";
-            }
-            case IMPERIAL_FEET, IMPERIAL_METER -> {
-                perUnitStringId = R.string.voice_per_mile;
-                distanceId = R.string.voiceDistanceMilesPlural;
-                speedId = R.string.voiceSpeedMilesPerHourPlural;
-                unitDistanceTTS = "mile";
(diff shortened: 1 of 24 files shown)


dennisguse:
Didn't you want to make this optional aka have setting for this?

hercek:
I mentioned it in the section "alternative options considered" and I said my opinion: "it is not worth it". I mean why would anybody want it. It is always the same. It just takes talking time. Maybe we may add a text field into announcer options which would say: "The quantities announced are always in the units set in the preferences. The units are not mentioned when announcing." But I think even this is not needed. Just makes the  settings screen more cluttered instead of making announcer messages more cluttered. Well, at least, people do not visit setting screen that often as they hear the announcer messages. So it would be better.

The current situation where units are repeated all the time in the announcer speech is outright terrible. Well, at least for me. I doubt there is a person who would be so exhausted from training that he/she would forgot what units he/she is using :)

The only place where units stay is when reporting times. The primary reason is that there is no concise way to report time with 1s precision while giving people a good intuition how long does it take. One could decide to report it all in seconds only but then hardly anybody would be able to reference it to intuitive minutes and hours when the reporting number gets into thousands. The commonly known American military time format does not report seconds. So it is not usable.

If you do not like it then do not accept it. I'm not going to bother with having announcer units option controllable but maybe somebody else will. Maybe there is some way how to get this to a vote?  ... to figure out what other people think

dennisguse:
In this situation, we are breaking an existing features (that exists since 10+ years) by replacing it with a new one.
The alternative solution is to just add the new feature (by adding a setting).

I am okay with breaking changes if they simplify the code base or add some other fancy feature.
In this case, the added complexity by having a setting is not complicated and the whole voice announcement module is also rather simple.

hercek:
Fair enough.

dennisguse:
@hercek Do you think you would change this PR to include a setting?

hercek:
No, my branch completely without units works fine for me. My branch removes quite a lot of code. Making this option controllable adds code to the original version. This has almost zero priority for me now since I think having units there is pretty much always annoying.

For anybody who would take over this:

1) Consider adding a title for "total moving time". Currently total moving time does not have any title/preamble. It is not strictly necessary since it is the only quantity without any title. So people can parse the output correctly knowing that if the reported quantity is not named then it is the total moving time. I considered myself to remove not only units but also the titles but, later, I realized that the title is good to give me time to increase concentration to catch the numbers.

2) Some strings used in the announcer are not localized. Consider make them localized.

PR #1821, opened 2024-01-10
Voice Announcements remove "point zero"

griffinonacid:
Check if total distance is a whole number to prevent TTS reading out "point zero" unnecessarily


dennisguse:
Thanks :D

PR #1856, opened 2024-02-13
UI: re-add OLED friendly (aka black) theme.

### LINKED ISSUE #1625: Pure black/OLED theme missing in v4.5.0
**Describe the bug**
The pure black/OLED theme isn't there anymore in the latest update.

**To Reproduce**
1. Go to User Interface > UI Theme
2. The black theme is missing

**Technical information**
 - OpenTracks version: v4.5.0

dennisguse:
Fixes #1625.

~TODO: needs some trying if it works ;)~

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/AbstractActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/AbstractActivity.java
@@ -22,6 +22,7 @@ import android.view.View;
 import androidx.appcompat.app.AppCompatActivity;
 
 import de.dennisguse.opentracks.services.announcement.TTSManager;
+import de.dennisguse.opentracks.settings.PreferencesUtils;
 
 /**
  * @author Jimmy Shih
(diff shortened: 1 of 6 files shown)


dennisguse:
Okay; here is a tiny problem (and I am not sure how to continue).
1. Setting a theme can only be done in `Activity.onCreate()`
2. ... if it is changed; `Activity.recreate()` must be called [works for SettingsActivity]
3. an activity in the launch stack (e.g., TrackListActivity that starts SettingsActivity) is actually only `resumed`.

Countermeasure: also after SettingsActivity is finished and control is passed it's calling activity also call `Activity.recreate()`.

BUT: settings can be opened like this:
TrackListActivity (start recording) -> TrackRecordingActivity -> SettingsActivity

So, now we also would need to recreate TrackListActivity to apply the changed theme.
How?
I mean, I don't want to recreate an activity any  time another activity it called finished - that would be waste of resources, right?

One option I checked: somehow use `getTheme` to check if the current theme needs to be updated.
I am not really sure, how to do this though...

@pstorch any other suggestions?

pstorch:
No idea. Maybe just inform the user that the changed theme needs a restart to be properly applied?

dennisguse:
@pstorch the manual approach js implemented. 
Can you give it a try and see if it makes sense?

pstorch:
Some changes are applied immediately, some (from and to the oled theme) are applied after restart. 
Works for me.  :+1:

dennisguse:
@pstorch thanks for testing.
I really don't get, why this is soo complicated.
The easiest approach would be to just trigger a configuration change (e.g. screen rotation), but Android just does not support a "non-drastic" config change.
Or something else.....

PR #1960, opened 2024-07-15
fix #1957 open MarkerEditActivity to receive external marker

### LINKED ISSUE #1957: RFE - allow adding markers after finishing track
This is a straight up request for enhancement.
I have no current troubles, but would like to have the ability to insert a marker after finishing a track.

Scenario:
- Moving through space;
- Have to reverse because of blockage (DON'T want to stop to add marker because of losing time);
- Continue to end and finish track;
- Now would be a good time to go back and add a marker and annotation at the location of the blockage;
- Oops - can't mark the track at the blockage

Ideal solution would be:
- Open finished track
- View map
- Click three dots
- Select "Edit Track"
- Tap on track at significant location
- Choose "Add Marker"
- Edit marker information
- Click "Done"

I did try "continue track" after having finishing it, but it appears markers can only be added at the current location/time.

**Additional context**

pstorch:
Opens MarkerEditActivity to be able to receive external marker for a (completed) track.


pstorch:
I pulled out the location retrieval from the `TrackRecordingManager.insertMarker()` and provide it via Intent to  the `MarkerEditActivity`.
But the `insertMarker()` is still part of the `TrackRecordingManager` because it retrieves also the `TrackStatistics` to set `length` and `duration` into the new `Marker`.
I see that both fields are used in the `ChartView`. Is that really necessary to draw the Markers on the Chart? 
And there are a lot of deprecations, I don't see what the replacements should be.

dennisguse:
Long story short: yes, duration and distance are only for the charts. Best guess, we should just remove these as it doesn't always fit. 
And the benefit is not that much there. 
It may be a leftover from MyTracks: there were statistics marker. Those are gone since ages ;)

For the API, I would just leave these values empty. 
To limit the dependency to TrackRecordingManager, I would make the call to it before the API.

If you want to have fun, you can also remove duration and distance:)

pstorch:
@dennisguse don't tell me to delete something. I love removing unused code :wink: 
`length` and `duration` are gone now from the `Marker`. And the Markers are gone from the ChartView.
We could now pull out the `insertMarker` method from TrackRecordingService/Manger into the MarkerEditActivity or the MarkerEditViewModel.

What is your opinion with `EXTRA_TRACK_ID` vs. `EXTRA_TRACK_ID_LONG`? Internally I guess you would like to stick to the `Track.Id` type. But this is not possible from the outside. That's why we have now two TRACK_ID Intent parameter.

dennisguse:
What about creating a new activity like .api.CreateMarker? Then this activity can do the parsing, security checks (if necessary), and we do not expose an internal activity;)

dennisguse:
And the DB migration is missing - deletion comes with ugly parts :)

pstorch:
> What about creating a new activity like .api.CreateMarker? Then this activity can do the parsing, security checks (if necessary), and we do not expose an internal activity;)

That's a good idea. I'll do that. So external interface is separated from the internal. Should we do it for the `MarkerDetailActivity` as well? Even if it is "just" forwarding?

pstorch:
> And the DB migration is missing - deletion comes with ugly parts :)

I see the ugly part now. Is it really necessary to drop the two columns from the DB? They are nullable anyway. 
The SQLite version in Android doesn't support `DROP COLUMN` yet, so we would need to copy and recreate the table. 

For the public API part, I've created two new Activities for it which receive the external extras from the Intent and forward to the internal Activities if appropriate. Do we want to allow that only if the public API is enabled? The ShowMarkerActivity might be ok, but the CreateMarkerActivity writes to the DB.

dennisguse:
Public API: better safe than sorry. 
Drop column: there is code for that already in place. You can just copy and adapt it. 
Been there done that countless times already :)

pstorch:
> Been there done that countless times already :)

I know, seen that :wink: 

Ok, I've added DB version 38 with upgrade and downgrade methods.

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
```diff
@@ -0,0 +1,82 @@
...
+
+        Track.Id trackId;
+        if (getIntent().hasExtra(EXTRA_TRACK_ID)) {
+            trackId = new Track.Id(getIntent().getLongExtra(EXTRA_TRACK_ID, 0L));
+        } else {
+            trackId = null;
+        }
+        TrackPoint trackPoint;
+        if (getIntent().hasExtra(EXTRA_LOCATION)) {
+            Location location = getIntent().getParcelableExtra(EXTRA_LOCATION);
```
dennisguse, line 48: Why?

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        // TODO: do we want to check if public API is enabled?
+
+        Track.Id trackId;
+        if (getIntent().hasExtra(EXTRA_TRACK_ID)) {
+            trackId = new Track.Id(getIntent().getLongExtra(EXTRA_TRACK_ID, 0L));
```
dennisguse, line 42: Rather crash if this doesn't work.

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
```diff
@@ -0,0 +1,82 @@
...
+        } else {
+            trackId = null;
```
dennisguse, line 44: Isn't this mandatory?
If yes: crash

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
```diff
@@ -0,0 +1,82 @@
...
+        TrackPoint trackPoint;
+        if (getIntent().hasExtra(EXTRA_LOCATION)) {
+            Location location = getIntent().getParcelableExtra(EXTRA_LOCATION);
+            trackPoint = new TrackPoint(location, Instant.now());
+        } else {
+            trackPoint = null;
+        }
+
+        TrackRecordingServiceConnection.execute(this, (service, self) -> {
+            Marker.Id marker = createNewMarker(trackId, trackPoint, service);
```
dennisguse, line 55: Why not pass in the location?
Or rather the value object that we now have in OpenTracks. 
The trackpoint shouldnt be required.

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
```diff
@@ -0,0 +1,82 @@
...
+                        .putExtra(MarkerEditActivity.EXTRA_MARKER_ID, marker)
+                        .putExtra(MarkerEditActivity.EXTRA_IS_NEW_MARKER, true);
+                startActivity(intent);
+            }
+            finish();
+        });
+    }
+
+    private Marker.Id createNewMarker(Track.Id trackId, TrackPoint trackPoint, TrackRecordingService trackRecordingService) {
+        try {
```
dennisguse, line 71: Why try catch?

Marking parameters as not nullable helps.

src/main/java/de/dennisguse/opentracks/publicapi/ShowMarkerActivity.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private List<Marker.Id> markerIds;
+
+    @Override
+    protected void onCreate(Bundle bundle) {
+        super.onCreate(bundle);
+
+        Marker.Id markerId = null;
+        if (getIntent().hasExtra(EXTRA_MARKER_ID)) {
+            markerId = new Marker.Id(getIntent().getLongExtra(EXTRA_MARKER_ID, 0));
```
dennisguse, line 49: No default value possible.

src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
```diff
@@ -138,7 +138,7 @@ Pair<Track, Pair<TrackPoint, SensorDataSet>> getDataForUI() {
         return new Pair<>(track, current);
     }
 
-    public Marker.Id insertMarker(String name, String category, String description, String photoUrl) {
+    public Marker.Id insertMarker(String name, String category, String description, String photoUrl, Track.Id trackId, TrackPoint trackPoint) {
```
dennisguse, line 141: Again; rather pass the location as the reference to the trackpoint will be lost anyhow.

dennisguse: And for imports: it may not even exist. 
Often markers are close to the track but not actually on it.

src/main/java/de/dennisguse/opentracks/ui/markers/MarkerEditActivity.java
```diff
@@ -59,13 +62,16 @@ public class MarkerEditActivity extends AbstractActivity {
...
+    public static final String EXTRA_LOCATION = "location";
+    public static final String EXTRA_IS_NEW_MARKER = "is_new_marker";
 
     private static final String CAMERA_PHOTO_URI_KEY = "camera_photo_uri_key";
 
     private static final String NEW_MARKER_ID = "new_marker_id";
 
     private static final String TAG = MarkerEditActivity.class.getSimpleName();
     private Track.Id trackId;
+    private Location location;
```
dennisguse, line 74: We have an internal value object for location. 
Reason was that the android location requires lat/lng which is sometimes not present (e.g. only elevation via barometer).

src/main/java/de/dennisguse/opentracks/ui/markers/MarkerEditActivity.java
```diff
@@ -176,7 +183,11 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private Marker.Id createNewMarker(TrackRecordingService trackRecordingService) {
         try {
-            Marker.Id marker = trackRecordingService.insertMarker("", "", "", null);
+            if (location == null) {
+                throw new IllegalStateException("Location is null");
+            }
+            TrackPoint trackPoint = new TrackPoint(location, Instant.now());
```
dennisguse, line 189: Trackpoints should only be created by trackpointcreator and during import. 
Rather pass a location (time is even not used for markers anymore).

dennisguse (changes requested):

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
pstorch: Unfortunately Marker need  `accuracy`, `altitude` and  `bearing` from the TrackPoint. Or should we remove them, too?

src/main/java/de/dennisguse/opentracks/services/TrackRecordingManager.java
pstorch: not possible without further stripping down of Marker, see comment above.

src/main/java/de/dennisguse/opentracks/ui/markers/MarkerEditActivity.java
pstorch: Where is that? Even TrackPoint is using the android Location.

src/main/java/de/dennisguse/opentracks/ui/markers/MarkerEditActivity.java
pstorch: again, see first comment about Location vs. TrackPoint for Marker.

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
pstorch: Why what?

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
pstorch: Ok, let's crash :wink:

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
pstorch: Ok, let's crash :wink:

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
pstorch: done

src/main/java/de/dennisguse/opentracks/publicapi/ShowMarkerActivity.java
pstorch: The `getLongExtra()` method requires a default value.

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
```diff
@@ -0,0 +1,69 @@
...
+    public static final String EXTRA_TRACK_ID = "track_id";
+    public static final String EXTRA_LOCATION = "location";
+
+    private static final String TAG = CreateMarkerActivity.class.getSimpleName();
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        if (!PreferencesUtils.isPublicAPIenabled()) {
```
pstorch, line 40: This check is rather annoying.
I'm not sure, if it really makes it more secure to check if the public API is enabled.
If the user clicks `cancel` the marker will be deleted anyhow.

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
dennisguse: https://github.com/OpenTracksApp/OpenTracks/blob/main/src%2Fmain%2Fjava%2Fde%2Fdennisguse%2Fopentracks%2Fdata%2Fmodels%2FPosition.java

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
dennisguse: Why is this optional?

pstorch: Outdated

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
pstorch: Thanks, was searching for Location

src/main/java/de/dennisguse/opentracks/publicapi/ShowMarkerActivity.java
dennisguse: Then the default value should trigger a failure.

pstorch: ok, done

src/main/java/de/dennisguse/opentracks/ui/markers/MarkerEditActivity.java
pstorch: Wasn't there a link to a Position class or record I saw in the email?

src/main/java/de/dennisguse/opentracks/io/file/importer/TrackImporter.java
```diff
@@ -220,51 +220,37 @@ private void adjustTrackPoints() {
...
-        }
+        markers.clear();
 
         if (todoMarkers.isEmpty()) {
             Log.w(TAG, "Some markers could not be attached to TrackPoints; those are not imported.");
+            return;
         }
 
-        markers.clear();
+        // TODO: why do we need to match the markers with the TrackPoints? We will loose the manual added from the map.
```
pstorch, line 232: @dennisguse what do we do here?

dennisguse: Old stuff. Back then there was code to only import markers that could be mapped to one TrackPoint. But that is not needed anymore.

pstorch: More to delete 🙌

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
dennisguse: Not sure. The problem nowadays is that the marker is saved first and deleted only if the user cancels this operation. If the app gets killed the marker will remain.

src/main/java/de/dennisguse/opentracks/io/file/importer/TrackImporter.java
pstorch: gone

src/main/java/de/dennisguse/opentracks/publicapi/CreateMarkerActivity.java
dennisguse: Sounds like a good plan :)

pstorch: Changed the behavior of the `MarkerEditActivity` to insert the Marker at the very end `onDone` in `MarkerEditViewModel`.

dennisguse:
Squashing and reordering commits and this is good to go?

pstorch:
> Squashing and reordering commits and this is good to go?

Squashing can be done here via the Web UI as well. What do you have in mind with reordering?

dennisguse:
@pstorch if you would want separate commits ;)
Squash is fine for me.

Issue #1904, opened 2024-04-26
When starting OpenTracks via GadgetBridge no data is being recorded


sebastianha:
When starting a new recording with my Mi Band 7 via Gadgetbridge a new recording is being started with opentracks but directly after the start stopped and no data is being recorded.

I know from old recordings that is worked until at least 2024-03-02, I just discovered that all other recordings are empty since then.

This is only the case when started via Gadgetbridge (via intent?), normal recording works. I checked all the settings, they seem to be correct.

...

**Checklist**

- [x] Consulted [https://dontkillmyapp.com/]
- [X] Power management: battery optimization is disabled for OpenTracks

**Technical information**

- Device: Pixel 7
- OS: GrapheneOS
- OpenTracks version: 4.12.2

dennisguse:
Might be #1653.

If the screen is off, some Android devices might enforce permission to access the location in the background.
And OpenTracks is not (yet) requesting this permission.

Can you check if it works, if the screen is on?

sebastianha:
Also does not work with screen on. I see that only Gadgetbridge is accessing Location and on Opentracks a new record is being created but directly stopped with no data.

dennisguse:
Can you check if there is a crash happening (check logcat)?

sebastianha:
I cannot see any crash, also when the app is open in foreground it stays open.
I do see a short popup at the top sometime when the app is not open and recording starts.

![Screenshot_20240429-220823](https://github.com/OpenTracksApp/OpenTracks/assets/420807/23e1c6cc-dd85-4908-b433-7b4de5477657)

dennisguse:
Can you start a recording (that has data) via the UI?
And the notification is only shown briefly or does it remain in the notification bar (assuming OpenTracks has a the notification permissoin)?

sebastianha:
Yes, with the UI it works.
The notification is only visible for a split second.

MagiX13:
I have the same issue when starting to track via intents (I have a device that I use only for sport activities and use tasker to start tracking once the connection is established). It creates a track of 0 seconds and that's all. Manually starting to track works without issues. 
I'm also running on GrapheneOS, but with a different device. The issue started occuring somewhere between 9 and 18 March. I can't say if this comes from 4.12 that I just installed or the new Graphene [version](https://grapheneos.org/releases#2024031400) that came out during that time frame.
The GrapheneOS logs for the app don't show anything suspicious in my view, but happy to share if needed.

[...]

PR #1989, opened 2024-09-25
fix(api,service): from intent api, launch record service on foreground

abdoulayeYATERA:
Fixes : https://github.com/OpenTracksApp/OpenTracks/issues/1904
As record service uses foreground notification, 
replaces the launch of the the record service from intent api.

Uses
~~~
executeForeground
~~~
Instead of 
~~~
execute
~~~

Please review and ajust (if needed) before merge, as my knowledge of the code base is close to null for now.

Thanks.

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/publicapi/AbstractAPIActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/publicapi/AbstractAPIActivity.java
@@ -37,7 +37,7 @@ public abstract class AbstractAPIActivity extends AppCompatActivity {
 
         if (PreferencesUtils.isPublicAPIenabled()) {
             Log.i(TAG, "Received and trying to execute requested action.");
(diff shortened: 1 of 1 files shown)


dennisguse:
That change will work for starting, but may crash for stopping.
`startForeground()` requires that a service is starting into foreground (if not it throws an exception).
Not sure what happens if the service is already in foreground mode.

abdoulayeYATERA:
Yes, good point, what should happen (functionally) if an api intent is received while a regular track recording is on ?

dennisguse:
For start:  nothing. 
For stop: stop the recording. 
Vice versa for stop.

arjan-s:
Hi, have you made progress on this? I'd still love to be able to use OpenTracks again!

dennisguse:
@arjan-s I didn't. 
Had no time to do anything. 
I also have no setup for this and thus testing is not that simple for me.

dennisguse:
Superseeded by: #2024

PR #2042, opened 2024-12-29
fix nightly filename

pstorch:
For the nightlies we need to `setVersionCodeOverride(getVersionCode())` before creating the `outputFileName = "${applicationId}_${variant.versionCode}.apk"`


dennisguse:
Details.....

Issue #2040, opened 2024-12-29
Possible improvements on the Statistics graphs/plots of recorded tracks


NiaJ3oE2LM:
### Is your feature request related to a problem? Please describe.

The value of pace is difficult to read precisely from the graph as the y-axis limits are way to far from the average:

+ it's not possible to zoom in the y-axis direction
+ it's not possible to hide other signals (for example: elevation)

The green/blue color legend assigned to Elevation and Speed/pace is barely distinguishable with a light theme and eye protection (night filter) on.

### Describe the solution you'd like

Plots of Elevation & pace from recorded tracks - either by time or by distance - could be improved in the following way:

+ allow scaling and limit setting of the y-axis for better visualization
+ make use of a dual y-axis for decoupling the scale of the two signals: for example you could plot the elevation on the left-hand y-axis and the pace on the right-hand y-axis, still using the same x-axis (time or distance)
+ click on the legend entry toggle hide/show of the associated signal. A user may want to hide a signal if it does not display relevant information.
+ would be *super nice* to have a light low-pass noise filter on speed/pace raw data.

### Additional context

May be related to:

+ [Charts: allow to hide elevation #1297](https://github.com/OpenTracksApp/OpenTracks/issues/1297)
+ [Something (scale?) wrong with elevation graphs #1241 ](https://github.com/OpenTracksApp/OpenTracks/issues/1241)
+ [Hairline for graphs #1596](https://github.com/OpenTracksApp/OpenTracks/issues/1596)
+ [Allow rounding elevation scale domain #795 ](https://github.com/OpenTracksApp/OpenTracks/issues/795)

App version `v4.16.1` installed from f-droid on android 10 and used for tracking **running** sessions.

dennisguse:
Any improvement of the diagrams would be nice.
Currently, no library is used (aka everything implemented in OpenTracks).

grobolom:
@dennisguse I wanted to jump into this thread, rather than have everything in the PR, so my apologies for the potential notification bombardment.

Do the maintainers of this project have a preference for a charting library that they'd be interested in using? I would not be opposed to experimenting with something like https://github.com/patrykandpatrick/vico or just directly using Jetpack Compose as a basis for generating slightly nicer charts and opening a PR, but I also don't want to step on anyone's toes if they already have something else in mind for this app.

dennisguse:
[...]

Issue #1297, opened 2022-06-25
Charts: allow to hide elevation


Mugen36:
**Is your feature request related to a problem? Please describe.**
The elevation data on my tracks is completely non-sense. I am running 5 times around the block. And the graph shows that I climbed down 20 meters. And frankly, if one use Opentracks for evening runs, tracking the altitude is not that important anyways.

**Describe the solution you'd like**
Make recording elevation data optional. Implement an option that disables the recording of altitude data.

EDIT: As not recording altitude data is not feasible,  a solution would be to give an option to not show altitude data in the graphs.

dennisguse:
The absolute elevation from GPS or gain/loss from the barometric sensor?
And not showing it or not storing it or not exporting it?

Where is it shown that you don't want it?

Mugen36:
I was referring to the absolute elevation from the gps. I actually do not see any data from the barometric sensor.
The GPS-Beidoo-Glonass location data and the running speed seems fairly accurate. Accuracy usually <5m.
Yet the elevation data is completely garbage. Not only does it show a 20m climb-down where there is none. It is also roughly 40m too high.
I would prefer an option of not storing it, since it is just plainly false.

I see the elevation data in the graphs when I look at the specific tracks.

Mugen36:
At the core, I ultimately would blame Google for not putting a better GPS in there Pixels. Yet, since elevation data is usually far less accurate than horizontal location, I can imagine many more people have this issue.

dennisguse:
Well no.

1. GPS accuracy for horizontal data is less accurate than vertical data (due to how GPS works)2
2. exported data is in WGS84 as specified by KML/GPX.
   but this is not above ground aka sea level value. That's why OpenTracks converts for the UI the elevation into EGM2008.

And implementing this switch is too much effort; especially maintaining it.
The only thing would be drop data that is not accurate enough as reported by the GPS (horizontal accuracy is provided)

Mugen36:
> And not showing it or not storing it or not exporting it?

Well, if not storing and not exporting, are not reasonable requests,
could I ask for an option for not showing elevation data in the graphs?

dennisguse:
That would work.
Interested to implement it?

Mugen36:
Sure

Mugen36:
May I ask, am I the only one who is getting non-sensical altitude values? 

Or am I just the only one who is annoyed by seeing these in the graph?

dennisguse:
[...]

PR #2054, opened 2025-01-23
WIP - add settings toggles for chart series (#1297)

grobolom:
## Description
Adds toggles to the User Interface section of the app Settings that allow the user to display or hide the three series available in the chart: Elevation, Speed/Pace, and Heart Rate. As a user of a heart rate monitor primarily during strength workouts and combat sports, the distance and elevation graphs are not helpful to me - in addition, their scale means that I am not able to easily see my exact heart rate. This should allow users in a similar situation to better determine exactly what is most helpful for them to be seen on the graph.

image
image

## Related Issues
- #1297 
- #2040

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/chart/ChartFragment.java
+++ b/src/main/java/de/dennisguse/opentracks/chart/ChartFragment.java
@@ -86,6 +86,7 @@ public class ChartFragment extends Fragment implements TrackDataHub.Listener {
                     });
                 }
             }
+
             if (PreferencesUtils.isKey(R.string.stats_rate_key, key)) {
                 boolean reportSpeed = PreferencesUtils.isReportSpeed(activityTypeLocalized);
                 if (reportSpeed != viewBinding.chartView.getReportSpeed()) {
@@ -98,6 +99,21 @@ public class ChartFragment extends Fragment implements TrackDataHub.Listener {
                         }
                     });
                 }
+            }
+
+            if (PreferencesUtils.isKey(R.string.chart_display_elevation_key, key)) {
+                viewBinding.chartView.setShowElevation(PreferencesUtils.shouldShowElevation());
+                refreshChart();
+            }
+
+            if (PreferencesUtils.isKey(R.string.chart_display_pace_or_speed_key, key)) {
+                viewBinding.chartView.setShowPaceOrSpeed(PreferencesUtils.shouldShowPaceOrSpeed());
+                refreshChart();
+            }
+
+            if (PreferencesUtils.isKey(R.string.chart_display_heart_rate_key, key)) {
+                viewBinding.chartView.setShowHeartRate(PreferencesUtils.shouldShowHeartRate());
+                refreshChart();
             }
         }
     };
@@ -272,5 +288,14 @@ public class ChartFragment extends Fragment implements TrackDataHub.Listener {
         if (fragmentActivity != null) {
             fragmentActivity.runOnUiThread(runnable);
         }
+    }
+
+    private void refreshChart() {
+        runOnUiThread(() -> {
+            if (isResumed()) {
+                viewBinding.chartView.invalidate();
+                viewBinding.chartView.requestLayout();
+            }
+        });
     }
 }
--- a/src/main/java/de/dennisguse/opentracks/chart/ChartView.java
+++ b/src/main/java/de/dennisguse/opentracks/chart/ChartView.java
@@ -86,8 +86,10 @@ public class ChartView extends View {
     }
 
     private final List<ChartValueSeries> seriesList = new LinkedList<>();
+    private final ChartValueSeries elevationSeries;
     private final ChartValueSeries speedSeries;
     private final ChartValueSeries paceSeries;
+    private final ChartValueSeries heartRateSeries;
 
     private final LinkedList<ChartPoint> chartPoints = new LinkedList<>();
     private final List<Marker> markers = new LinkedList<>();
@@ -123,6 +125,7 @@ public class ChartView extends View {
     private boolean chartByDistance = false;
     private UnitSystem unitSystem = UnitSystem.defaultUnitSystem();
     private boolean reportSpeed = true;
+    private boolean showPaceOrSpeed = true;
     private boolean showPointer = false;
 
     private final GestureDetectorCompat detectorScrollFlingTab = new GestureDetectorCompat(getContext(), new GestureDetector.SimpleOnGestureListener() {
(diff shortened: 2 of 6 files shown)


src/main/java/de/dennisguse/opentracks/chart/ChartView.java
```diff
@@ -343,9 +350,13 @@ protected boolean drawIfChartPointHasNoData() {
         setClickable(true);
         updateDimensions();
 
-        // either speedSeries or paceSeries should be enabled.
-        speedSeries.setEnabled(reportSpeed);
-        paceSeries.setEnabled(!reportSpeed);
+        // Either speedSeries or paceSeries should be enabled, if one is shown.
+        speedSeries.setEnabled(reportSpeed && showPaceOrSpeed);
```
dennisguse, line 354: I guess wrapping this in an if (showPaceOrSpeed) won'hurt.

src/main/java/de/dennisguse/opentracks/chart/ChartFragment.java
```diff
@@ -99,6 +100,50 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
...
                 }
             }
+
+            if (PreferencesUtils.isKey(R.string.chart_display_elevation_key, key)) {
+                boolean showElevation = PreferencesUtils.shouldShowElevation();
+                if (showElevation != viewBinding.chartView.getShowElevation()) {
+                    viewBinding.chartView.setShowElevation(showElevation);
+                    viewBinding.chartView.applyShowElevation();
+
+                    runOnUiThread(() -> {
```
dennisguse, line 110: This looks like a helper method (used three times).

src/main/java/de/dennisguse/opentracks/chart/ChartFragment.java
```diff
@@ -99,6 +100,50 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
                     });
                 }
             }
+
+            if (PreferencesUtils.isKey(R.string.chart_display_elevation_key, key)) {
+                boolean showElevation = PreferencesUtils.shouldShowElevation();
+                if (showElevation != viewBinding.chartView.getShowElevation()) {
+                    viewBinding.chartView.setShowElevation(showElevation);
+                    viewBinding.chartView.applyShowElevation();
```
dennisguse, line 108: Why do we need `applyShowElevation()`?
This can be done directly in `setShowElevation()`.

This affects also the other two `applyXXX`-methods you introduced.

src/main/java/de/dennisguse/opentracks/chart/ChartView.java
```diff
@@ -396,6 +411,40 @@ public boolean applyReportSpeed() {
...
+        }
+        // we want to make sure we show whatever version the user has
+        // selected when we turn this back on.
+        applyReportSpeed();
+        return true;
+    }
+    public void setShowHeartRate(boolean value) { showHeartRate = value; }
+    public boolean getShowHeartRate() { return showHeartRate; }
+    public boolean applyShowHeartRate() {
+        if (!showElevation) {
```
dennisguse, line 440: This is broken :)

dennisguse (changes requested):
Nice!

Some feedback included.
Just one more thing: can you make all methods package private that you introduce?
I guess most of them should not be used outside of `.chart`.

grobolom:
@dennisguse thank you for the feedback! I'll update, and I'm going to give adding some tests for the new functionality a shot, if I can.

I had a question - how do y'all handle translations? I see most pull requests that modify files like `settings_user_interface.xml` don't include them, but my IDE complains about it :)

src/main/java/de/dennisguse/opentracks/chart/ChartFragment.java
grobolom: @dennisguse I was thinking it would make sense that the `set`tter method only modifies the value itself, while `apply` could handle any other kind of processing that is related to the current value. But I'm not opposed to rolling it in - I was mostly following the existing convention for `applyReportSpeed`.

dennisguse:
Translations: handled via Weblate. Just add the english ones and the rest will be translated eventually if somebody feels the need for it. 
This actually works :D

Package private: yes.

Regarding the code: if you see something that can be improved, you can just do it.
Following the pattern of the existing code is fine, but there is always room for improvement (especially as the code base of OpenTracks evolved quite a lot in the 10+ years).
Just put cleanup and functional changes into separate commits :D

src/main/java/de/dennisguse/opentracks/chart/ChartView.java
```diff
@@ -396,6 +418,31 @@ public boolean applyReportSpeed() {
...
+            return;
+        }
+
+        // we want to make sure we show whatever version the user has
+        // selected when we turn this back on.
+        applyReportSpeed();
+    }
+    boolean getShowPaceOrSpeed() { return showPaceOrSpeed; }
+    void setShowHeartRate(boolean value) {
+        showHeartRate = value;
```
dennisguse, line 441: You can use isEnabled() on each Series: no need to store this value here also.

src/main/java/de/dennisguse/opentracks/chart/ChartView.java
```diff
@@ -379,6 +397,10 @@ public void setReportSpeed(boolean value) {
     }
 
     public boolean applyReportSpeed() {
+        if (!showPaceOrSpeed) {
+            return true;
```
dennisguse, line 401: Should setEnabled(false) on speedSeries, right/

dennisguse (changes requested):

src/main/java/de/dennisguse/opentracks/chart/ChartView.java
grobolom: @dennisguse while this is true for `showHeartRate` and `showElevation`, the boolean tracking is necessary for `showPaceOrSpeed` because of the interaction between the two elements on the `Chart` itself. If you prefer, I can remove it for just the elevation and heart rate, but I could imagine someone returning and needing it again, if they work on this code further.

src/main/java/de/dennisguse/opentracks/chart/ChartView.java
grobolom: @dennisguse technically this was already happening as part of `setShowPaceOrSpeed`, so it was not necessary. But I think it's probably best to consolidate that here, so I'll move it over.

grobolom:
@dennisguse ready for one more round I believe - tests are Red but they seem to be the same ones as the last release. Manual testing on API version 35 as well as my Pixel 3a shows no issues that I can find. Also, please let me know if you'd like me to squash my commits, I am happy to if that leaves a cleaner commit path.

dennisguse (approved):
Looks good. 

And yes, there are two tests prone to failure. 
No easy way to fix it :/

dennisguse:
PS squashing commits would be great.

grobolom:
@dennisguse squashed and pushed :)

Issue #2073, opened 2025-03-13
BottomAppBar partly hidden: Pixel 8 Pro / Android 15 [v4.18.0]


dennisguse:
With the release of v4.18.0, the BottomAppBar was manually set in height.
However, on a Pixel 8 Pro in vertical mode (with high resolution setting), the BottomAppBar partly hidden.
Landscape works.
This effect only occurs if a FAB Button (start/ stop recording) is shown.

The problem cannot be reproduced on:
* Pixel 8 with Android 15
* Pixel 8 Pro Emulator

![Image](https://github.com/user-attachments/assets/8a07bd53-dfbb-49a4-b0e8-328fabc2d361)
![Image](https://github.com/user-attachments/assets/a6e8f7a5-1988-4852-9b72-57b3bc4ebc65)
![Image](https://github.com/user-attachments/assets/615349a3-f6c0-491c-b2a4-1050ca7cae40)
![Image](https://github.com/user-attachments/assets/aa5ad0b5-0b9e-4f72-9e36-0c568033f260)
![Image](https://github.com/user-attachments/assets/ffa5cbd9-5ee0-4e12-8439-c2026435744a)

https://github.com/OpenTracksApp/OpenTracks/commit/5ae7ef350c6693080865fcb64bd4eb202701f5fe

dennisguse:
The problem seems to be unrelated to the full resolution mode of the Pixel 8 Pro.

Buntbart:
same Problem here with v.4.19.0 from F-Droid on a Oneplus 6t under LineageOS 22.1 (Android 15).

![Image](https://github.com/user-attachments/assets/96732bd6-d388-44ad-8600-2effdcc56956)

![Image](https://github.com/user-attachments/assets/01be92b0-b500-41cf-abdd-a2e43f8b6403)

Megadash452:
Same issue on a galaxy S24 with oneui7/android 15. It also happens on screens without the FAB, as shown in the picture.

![Image](https://github.com/user-attachments/assets/e357ee36-7c23-4b7d-8c7f-7db3e8ea4edb)

Megadash452:
[...]

Issue #2103, opened 2025-05-18
Icons (OSM dashboard icon, e.g.) don't appear at bottom of screen


dhlocker:
**Describe the bug**
The map icon for OSM dashboard launch is not appearing at the bottom of the main screen. (I recall other icons were there in past, but I never used them, so don't remember what they were.)

I'm sorry that I don't remember when this started; today is the first I noticed it and I don't remember the last time I looked at a track through the OSM dashboard. I do know that my phone was updated recently (06May2025) and there are many things that look "different".

**To Reproduce**
1 - Record a track
2 - Open the track
3 - Observe no map icon (see screenshot)
4 - screen doesn't scroll up/down, but left/right swipes select the other views (interval, by time, by distance)

![Image](https://github.com/user-attachments/assets/a1fb8da2-afd9-4f75-809f-4420d40699ae)

**Technical information**
 - Device: Samsung Galaxy S22; SM-S901U; Rev 1.0
 - OS: Android v.15; Kernel version 5.10.226-android12-9-30958166-abS901USQU8FYDB
 - OpenTracks version: Version name: v4.19.2irreproducible/v4.19.2; version code 6124
 - OSM dashboard version: release - 4.3.0 (44)

PR #2111, opened 2025-06-06
Enabled edge-to-edge rendering and applied navbar insets

Megadash452:
Starting on Android 15, the system forces edge-to-edge rendering on all apps.
This app did account for that, so the BottomAppBar was rendering under the navbar, making it unusable.

Solution: Enable edge-to-edge rendering and apply insets to BottomAppBar and the other content.

Some tests fail, but its always inconsistent on which ones fail.

This PR resolves issues #2073 and #2103.
Solution found in [this article](https://developer.android.com/develop/ui/views/layout/edge-to-edge#java).

### DIFF
--- a/src/main/java/de/dennisguse/opentracks/AbstractActivity.java
+++ b/src/main/java/de/dennisguse/opentracks/AbstractActivity.java
@@ -50,8 +50,12 @@ public abstract class AbstractActivity extends AppCompatActivity {
         View rootView = getRootView();
         setContentView(rootView);
 
+        apply_insets(rootView);
+    }
+
+    private void apply_insets(View rootView) {
         View bottom_app_bar = findViewById(R.id.bottom_app_bar);
-        if (bottom_app_bar != null) {
+        if (rootView != null) {
             // Apply navbar insets to the whole content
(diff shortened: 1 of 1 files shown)


dennisguse:
Looks good. 
And I have seen this documentation ages, but didn't understand that it was required. 

TBH I am really annoyed by the necessary effort to get this "improvement" working. I was assuming that the Material components would and should take care...

Anyhow: can you move this functionality into a helper method? Something like "configureEdge2EdgeForBottomAppBar"?

Megadash452:
> TBH I am really annoyed by the necessary effort to get this "improvement" working.

Yeah it's so annoying how unnecessarilly complex android dev is.

dennisguse:
@Megadash452 thanks for your time on this topic - it is not a fun one...

Megadash452:
> @Megadash452 thanks for your time on this topic - it is not a fun one...

Yeah man 😂. I just want to be able to use this app normally again.

dennisguse:
@Megadash452 would mind opening a PR: then I can take look and also start tinkering around.
I don't see why should have all the fun alone...

Megadash452:
oh sure

PR #2114, opened 2025-06-12
Enabled edge-to-edge rendering and applied navbar insets (Continued)

Megadash452:
Continuation of #2111. App content needs to be inset


dennisguse:
One thing we can do to simplify the situation: remove the AppBarLayout.
It gives the feature that if the user scrolls down, the AppBar is moved out of the screen.
This is totally non-essential :D

Megadash452:
Maybe instead we can give up on applying insets on individual components and just apply it on the root view?

PR #2131, opened 2025-07-03
Update Turkish

mikropsoft:
I wanted to retranslate the file from scratch for a better translation. Whether to merge it or not is up to your preference. I know that Weblate exists, but I prefer Crowdin instead, and the last time I used Weblate, I couldn't see an upload button for uploading files to any project. I would appreciate it if you could at least upload this file for me. Thank you for your understanding (the PR might look complicated because Weblate mixes up the files).


dennisguse:
I will check if it can be merged. 
However, this will be an exception as having different workflows for translations is huge effort for me.

